PR #281, opened 2014-11-26
bootstrap table method #filterBy

integer-underflow:
Public method that use to filter data on table (It can use only in client-side now.)

eg. $('#bs-tbl').bootstrapTable('filterBy', { deleted: true });
it will filter only deleted === true

I will update document later.


dabros:
Just in case anyone looking to do something with filterBy but cant find examples, im currently looking at http://bootsnipp.com/snippets/featured/panel-table-with-filters-per-column and maybe adapting code from there

djhvscf:
You should use the filterControl column option to do that

PR #306, opened 2014-12-04
On grid column filtering

ctengiz:
Column based filtering. Currently only supports text boxes for criteria entry. Select and datetime picker is going to be implemented. 

Supports both client side and server side processing.


djhvscf:
@wenzhixin you can close this PR

PR #350, opened 2014-12-16
Support for data-field with dot notation

wagnerjs:
Useful when you need fields that are inside an object.

First commit had a bug, for non-string fields and it was fixed.


wenzhixin:
Close this pull request, thanks anyway.

Issue #353, opened 2014-12-17
How do I use nested JSON data to populate a table?


ethompsy:
Hello!

I have a question:

I have a rather complicated (and useful) JSON object that I want to use with your bootstrap-table. I have looked at the javascript example, but the data2.json object is a bit more simple than my json endpoint. Mine has nested data. I want to use the nested data in the table. Here is an example of my json object:

```
{
    "messages": [
        {
            "id": 652, 
            "session_id": 8965861649940144331, 
            "date": "2014-12-17T01:01:09Z", 
            "m_type": "Error", 
            "message": "could not find a matching MDF file for legacy SQL selection $SQL:\\(local)\\Adams", 
            "file_name": "$SQL:\\(local)\\Adams", 
            "msdn": 33
        }, 
        {
            "id": 653, 
            "session_id": 8965861649940144331, 
            "date": "2014-12-17T01:01:10Z", 
            "m_type": "Error", 
            "message": "could not find a matching MDF file for legacy SQL selection $SQL:\\(local)\\TRAINING", 
            "file_name": "$SQL:\\(local)\\TRAINING", 
            "msdn": 33
        }
    ], 
    "session_id": 8965861649940144331, 
    "date": "2014-12-17T01:55:14Z", 
    "client_date": "2014-12-16T19:55:14Z", 
    "actual_cpu": "", 
    "licensed_cpu": "", 
    "license": 1, 
    "brand": 2, 
    "account": 215, 
    "computer": 289, 
    "regcode": "1:2:215:289"
}
```

How would I display a table of just the "messages" data?

Vijaeendra:
Hi I am trying to use the load a nested json object using the bootstrap-table. How do I make it work? I am not sure how and where to include the responseHandler . Can someone please explain. This library works like charm for all my other requirements, except the nested table one. Thanks in advance.

djhvscf:
Please check the flat-json extension

Vijaeendra:
I mean displaying the tree structure. My Json Object is something like this. 

{
  "children": [
    {
      "children": [
        {
          "children": [], 
          "expanded": "true", 
          "folder": "true", 
          "last_modified": "Thu Jul  9 17:13:48 2015", 
          "size": 4096, 
          "title": "child21"
        }
      ], 
      "expanded": "true", 
      "folder": "true", 
      "last_modified": "Thu Jul  9 17:13:48 2015", 
      "size": 4096, 
      "title": "Child2"
    }, 
    {
[...]

PR #355, opened 2014-12-18
Remote data container and server side pagination tuning

### LINKED ISSUE #352: Response data as a response object property
In my project I receive data to use in table grid as object property of a response:

    {
        status: 1,
        data: [...], // data array
        pagination: {...} // pagination object
    }
To interact with such response I had to made some changes to plugin. I add definition of remote data container name (null by defaults) and some workaround with server side pagination.

Would this changes be interesting to users? I can make a PR if it would be.

smgladkovskiy:
Fixes #352, #353


wenzhixin:
Let the plug-in is more simple, we don't consider to add this feature. Thanks for your advice and time anyway.

PR #384, opened 2014-12-31
Update bootstrap-table.js

arifemre:
the sorting process has been localized


wenzhixin:
Thanks! but I don't think the field `param` is necessary, because `cellStyle` is a column option, so we know which filed. I have removed it.

PR #414, opened 2015-01-09
Added trimOnSearch option, to make searching for multiple words easier

rml1997:
Added an option trimOnSearch, to make searching for multiple words easier. 
Previously trim was run at every run, getting rid of the space between words


wenzhixin:
Thanks, can you help to update the documentation: https://github.com/wenzhixin/bootstrap-table/blob/master/docs/_i18n/en/documentation/table-options.md :)

PR #420, opened 2015-01-12
Set elemetSize to control table element size, sm or lg

l0pez:
Set data-element-size to 'sm' or 'lg' to contolr element size. If undefined default size is present.


wenzhixin:
Hi @l0pez , perhaps more suitable for `iconSize`, do you think?

l0pez:
Yes, you right. It sounds better. I'll create this change.

PR #438, opened 2015-01-19
Set iconSize

l0pez:
Set iconSize to sm or lg to control icon size surrounded table.


wenzhixin:
Thanks, can you help to add an example or update the documentation?

PR #469, opened 2015-01-24
Updating the spanish documentation

djhvscf:


djhvscf:
@wenzhixin What do you think is better?? If I translated a whole folder and then commit it or commit by single file??

wenzhixin:
Both of them are OK.

PR #476, opened 2015-01-27
FAQ translated to spanish

djhvscf:


wenzhixin:
:+1: I want to link the examples of docs to: http://issues.wenzhixin.net.cn/bootstrap-table/, so needn't translate them, what do you think?

djhvscf:
I think it is a good idea... But I think that the style of that page http://issues.wenzhixin.net.cn/bootstrap-table/ can be improve.. Because it has a button without action. I already translate the examples because I did not read this comment before

PR #500, opened 2015-02-02
Add data-value attribute for sorting on <td>

dinatih:
Hello, I really like your lib compared to datatable. Could you please add this functionality ?
Here is a example of utilisation:
```
<tr><td data-value="1">1 year old</td></tr>
<tr><td data-value="2">2 years old</td></tr>
<tr><td data-value="10">10 year1 old</td></tr>
<tr><td data-value="20">20 years old</td></tr>

```
This permit a better sorting for number value displayed  in text and to wrap cell's data in a html container ```<td data-value="2"><span class="text-danger">2</span> years old</td>```

Have a good day !

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -355,7 +355,7 @@
                 // save td's id and class
                 row['_' + field + '_id'] = $(this).attr('id');
                 row['_' + field + '_class'] = $(this).attr('class');
-                row['_' + field + '_data_value'] = $(this).data('value');
+                row['_' + field + '_data_value'] = $(this).attr('data-value');
             });
             data.push(row);
         });
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -355,6 +355,7 @@
                 // save td's id and class
                 row['_' + field + '_id'] = $(this).attr('id');
                 row['_' + field + '_class'] = $(this).attr('class');
+                row['_' + field + '_data_value'] = $(this).data('value');
             });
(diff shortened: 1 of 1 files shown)


wenzhixin:
Hi @dinatih , in this case, maybe we can use `sorter` column option to sort the data?

wenzhixin:
I think we can use sorter function, for example:
```
sorter: function (a, b) {
    a = parseInt(a, 10);
    b = parseInt(b, 10);
    ...
}
```
Close this pull-request, thanks anyway.

dinatih:
Hello, really sorry I completely forget to answer you.

Yes we could use the sorter function, but what I love with your project is the use of declarative html5 data attributes for writing the less js code possible.
I use a lot of columns that use number data, but the data displayed is not pure number for lot of case, there is 2 examples : 
- For age, or delay to current date from a past date, display as "1 years ago" but with a useful value of "13-02-14" for sorting correctly with "1 month ago" cell). 
- Permit to wrap cell data into span (badges or label) without the need to use a sorter function as for example : 
```
sorter: function (a, b) {
    a = parseInt($(a).text(), 10)
    b = parseInt($(b).text(), 10);
    ...
}
```

Do you see my point ?

I am french, so my english maybe not so good :s sorry.

wenzhixin:
Thank you, I will consider this feature, #546

dinatih:
Alright! Your new approche is exactly what I want to do in first place, keep the data-* attributes unchanged. Thank you!

Issue #510, opened 2015-02-05
onCheck/onUncheck events not working with methods


macias3:
Hi,
I've discovered some inconsistency.

Events 'onCheck' and 'onUncheck' are not firing when using method 'check' or 'uncheck'.
Events 'onCheckAll' and 'onUncheckAll' do their job when method 'checkAll' or 'uncheckAll' is called.

Take a look:
http://jsfiddle.net/L78kzfxs/2/

Thanks

wenzhixin:
Merged, close this issue.

PR #521, opened 2015-02-09
Trigger onCheck/onUncheck event handler after calling method check/uncheck

macias3:
This fixes issue https://github.com/wenzhixin/bootstrap-table/issues/510. Method check_ triggers appropriate event handlers with requested row obejct as a parameter.


wenzhixin:
Thanks.

Issue #372, opened 2014-12-26
maintainSelected not working for server side pagination


mmochetti:
Checkbox/radiobuttons selections are not being preserved when the pagination is done on server side.
Would be great to be able to keep the state of the selections after the table is refreshed.

Use case:
1) Create table to use server side pagination and to maintain checkbox states across search and paginate:
           data-url="route/to/json" 
	   data-pagination="true"
	   data-side-pagination="server" 
           data-maintain-selected="true"
2) Check checkbox of first row
3) Search for something so that the first row is wiped out
4) Clear search so that first row appears again
5) Checkbox is shown unchecked.

For this I believe it's needed to keep an array of checked/selected "id's" and check them if they are present in the rendered rows and in this array. So the attribute "data-id-field" should be set to work properly.

thedrewster:
I concur. It would be good for the state of the checkboxes to be saved upon page refresh. DataTables (datatables.net) achieves this using session and localstorage.

wenzhixin:
I think we can handle this case manually, because we do not know when we need to save the state of the rows, and the `onSearch` and `onCheck` event can be help, do you think so?

thedrewster:
Yes, you are right. I have handled this manually and it works. It's a great extension to Bootstrap. Well done and thanks.

victorvasconcelosdeaquino:
How can I do this manually? I have the same problem...
Can you show me what you do to work?

wenzhixin:
@victorvasconcelosdeaquino  #450

wenzhixin:
Added example: Maintain selected on server side: http://issues.wenzhixin.net.cn/bootstrap-table/#issues/917.html

Junglefish:
Hi

I know this ticket is closed, but I'm having the same issue with **selectboxes**. I tried adapting the given example #917 tat is for checkboxes but can't get it to work. Does anyone have a similar solution for a simple single-select selectbox?

PR #532, opened 2015-02-13
New methods to select rows by value

macias3:
Hi Wenzhixin,

I'd like to introduce new methods: checkBy and uncheckBy to select rows by value or array of values.

You can call
$("#table").bootstrapTable("checkBy", {field:"field_name", value:"value1"}) 
to check/uncheck rows by a single value
or
$("#table").bootstrapTable("checkBy", {field:"field_name", value:["value1","value2","value3"]})
to check/uncheck rows by multiple values.

In addition, onCheckAll/onUncheckAll event handlers return an array of newly/previously selected row objects.

This two commits can help in the future with https://github.com/wenzhixin/bootstrap-table/issues/372

What do you think?

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1542,6 +1542,27 @@
         this.updateSelected();
         this.trigger(checked ? 'check' : 'uncheck', this.data[index]);
     };
+    
+    BootstrapTable.prototype.checkBy = function (obj) {
+        this.checkBy_(true, obj);
+    };
+
+    BootstrapTable.prototype.uncheckBy = function (obj) {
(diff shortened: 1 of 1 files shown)


wenzhixin:
Need to update the documentation: https://github.com/wenzhixin/bootstrap-table/blob/master/docs/_i18n/en/documentation/methods.md and add examples: https://github.com/wenzhixin/bootstrap-table-examples/tree/master/methods, can you help?

macias3:
I will update the documentation and add examples. I've noticed that method "remove" uses similar parameters like my methods. I'm going to make some changes in the code I proposed to keep things clear.

PR #543, opened 2015-02-16
Enhancement: Allow User to Reorder Columns

NaughtyMC:
Reorder columns on the fly... Just requires the Sortable Interaction from jQuery UI.

### DIFF
--- a/src/bootstrap-table.css
+++ b/src/bootstrap-table.css
@@ -237,3 +237,7 @@ div.fixed-table-scroll-outer {
     height: 150px;
     overflow: hidden;
 }
+
+.placeholder { 
+    background-color: #CCC;
+}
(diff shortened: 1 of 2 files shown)


NaughtyMC:
I am not very familiar with GitHub, so this fork is missing the jQuery UI dependency. However, on an independent project this code runs perfectly.

If someone who knows how to add it to my fork (if that's the best way to do it), I would appreciate your input. Thanks!

wenzhixin:
Because this feature is depended on jQuery UI, so I think it can be added as an extension here: https://github.com/wenzhixin/bootstrap-table/tree/master/src/extensions, like other extensions, do you think so?

NaughtyMC:
Yes, that sounds good to me!

wenzhixin:
Can you help to write an extension like others?

NaughtyMC:
I will work on it when I get the chance

PR #551, opened 2015-02-19
Bsgrid

naoufalELMASBAHI:
Now it's support editable functionality.


wenzhixin:
I don't want to change the repo name, and maybe your feature can be as an extension, thanks anyway.

PR #664, opened 2015-03-30
Added option to replace jquery ajax method

yonjah:


src/bootstrap-table.js
```diff
@@ -1463,10 +1466,11 @@
...
                 sortOrder: this.options.sortOrder
-            };
+            },
+            request;
 
-        if (!this.options.url) {
-            return;
+        if (!this.options.url && !this.options.ajax) {
+            throw new Error("Using Server requires either a `url` or a custom `ajax` method");
         }
```
yonjah: Notice that I've changed this condition from silently failing to send the ajax request to explicitly throwing an error.
This will give useful information if someone forget to set the `url` or `ajax` method after setting `sidePagination` to `server`, but might break existing code if someone has previously set the `sidePagination` to `server` without setting the `url` (not realizing no request is being sent).

wenzhixin:
We need to consider this case: http://issues.wenzhixin.net.cn/bootstrap-table/#options/from-data.html, so I change it to `return`. And when server side is 'client' it need to request the ajax also.

yonjah:
I'm not sure I understand.

Why do we need to call the initServer function if `data` is being used ?
it seems like this function is not doing anything if no `url` is set

yonjah:
Whats the bug ?
why would initServer be called if the `sidePagination` is in the `client`  ?
Correct me if I'm wrong but if pagination side is `client` there should be no ajax calls being sent.

Any way if this is an issue we can do the fix as you suggest, but I would suggest throwing an error in the constructor if options require server side calls but don't set either an `ajax` or `url` this probably any way better than doing this check on each server call as I did.

PR #673, opened 2015-04-01
Fix #672

### LINKED ISSUE #672: Column Fixed Width in Percentage Problem
I defined the column using percentage, however when switch to card view and switch back to table view.
The column width from % change back px.

jsfiddle:
https://jsfiddle.net/alvin_ng0618/8eh2t597/3/

djhvscf:


wenzhixin:
Fixed column with undefined bug.

djhvscf:
Sorry, thanks!

PR #675, opened 2015-04-01
Update bootstrap-table-cookie.js

murphmar:
Added state saving for visible columns and the ability to use extension with multiple simultaneous tables.


wenzhixin:
Updated

djhvscf:
Thanks again

PR #687, opened 2015-04-05
Fix #229

### LINKED ISSUE #229: Default Sort order is wrong if data-visible="false" is set on the field referenced by data-sort-name.
I am getting really out of order sort row sorting results. I am trying to sort by an integer and the rows are ordered { 4, 4, 52, 0, 0 }. This makes no sense.

I set the 'data-sort-name="score"' and 'data-sort-order="desc"'. Score is just a column of integer values.

If I sort on any other row those are out of order too (alphabetical, etc...). But! If I click the header for the "Score" column then it sorts correctly.

Looking at this it seems to default to sorting by the first column (which is an alphabetical value). This is sorted correctly, but if I click the header on this row it is sorted wrong. Really wrong...

djhvscf:
Fixed: Default Sort order is wrong if data-visible="false" is set on the field referenced by data-sort-name.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -661,10 +661,6 @@
                     return order * value;
                 }
 
-                if (value !== undefined) {
-                    return order * value;
-                }
-
                 // Fix #161: undefined or null string sort bug.
                 if (aa === undefined || aa === null) {
                     aa = '';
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks:)

PR #693, opened 2015-04-07
getData parameter

djhvscf:
to get the current page data


wenzhixin:
Updated: http://bootstrap-table.wenzhixin.net.cn/extensions/

PR #708, opened 2015-04-08
Row state field value fix

macias3:
Using "select all items" checkbox in the table header selects only enabled rows, but doesn't set proper state field value of disabled rows.

This little fix sets proper rows state field value depending on "checked" property of inputs.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1248,6 +1248,8 @@
                                 (value && value.disabled) ? 'disabled' : undefined) +
                             ' />',
                         that.options.cardView ? '</div>' : '</td>'].join('');
+                        
+                    item[that.header.stateField] = value === true || (value && value.checked);
                 } else {
                     value = typeof value === 'undefined' || value === null ?
                         that.options.undefinedText : value;
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1456,11 +1456,11 @@
         });
     };
 
-    BootstrapTable.prototype.updateRows = function (checked) {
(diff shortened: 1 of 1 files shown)


wenzhixin:
Can you provide a jsFiddle example with the bug?

macias3:
Of course. Look at: http://jsfiddle.net/avegsy7v/3/

Steps to reproduce the bug:
1. Select all items with the table header checkbox.
2. Click "get selected" button.
3. Compare names in the alert window with selected row names.

I think that two first rows should not be in the result set of "getSelections" method.

Selecting enabled rows one by one works as expected.

wenzhixin:
Thanks, I will check it out as soon as possible.

djhvscf:
:+1:

wenzhixin:
There is a little problem when you uncheck all items, the result would be `['multiple-select']`, what do you think?

macias3:
You're right! Perhaps the 'checked' parameter in updateRows function can be completly ignored?

I've created another jsfiddle, this time with working example: http://jsfiddle.net/39yjx6t1/2/

After checking/unchecking all rows it works as expected. I also fixed the initial value of rows state field.

Do you think it's ok now?

wenzhixin:
It's OK now:).
https://github.com/macias3/bootstrap-table/blob/master/src/bootstrap-table.js#L1853, here we need to update to `this.updateRows();`

wenzhixin:
Updated.

wenzhixin:
Thanks:)

macias3:
You're welcome :)

PR #732, opened 2015-04-10
Filtered checkbox rows not being checked correctly

Ritterbitz:
When the list is filtered, the indexes are based on the core dataset not
on the filtered dataset.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1978,7 +1978,7 @@
         }
 
         var that = this;
-        $.each(this.options.data, function (index, row) {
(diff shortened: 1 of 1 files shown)


wenzhixin:
Can you add an example with this bug?

wenzhixin:
Can you add an example with this bug?

wenzhixin:
Can you add an example with this bug?

wenzhixin:
Can you add an example with this bug?

djhvscf:
Provide an example to see the bug, and then reopen this PR

PR #738, opened 2015-04-11
Fix #688: Maintain sorting on refresh flattened json

### LINKED ISSUE #688: Maintain sorting on load/refresh flattened json
Hello,

I'm trying to keep the current column sorting on load/refresh data in my table. My current code isn't working. What's the correct way to achieve this? My table loads it's data from a json file.

    var $table = $('#table'),
      columnName = "",
      columnOrder = ""; 

    $table.bootstrapTable().on('sort.bs.table', function (e, name, order) {
      columnName = '[" + name + "]';
      columnOrder = '[" + order + "]';
    });

    $table.bootstrapTable({
      sortName: columnName,
      sortOrder: columnOrder
    });

    $table.bootstrapTable('refresh', {silent: true});

dimbslmh:
I changed a few lines to fix issue https://github.com/wenzhixin/bootstrap-table/issues/688. Please review.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -602,7 +602,7 @@
         } else if (type === 'prepend') {
             this.data = [].concat(data).concat(this.data);
         } else {
-            this.data = data || this.options.data;
+            this.data = this.options.data || data;
         }
         this.options.data = this.data;
 
--- a/src/extensions/flatJSON/bootstrap-table-flatJSON.js
+++ b/src/extensions/flatJSON/bootstrap-table-flatJSON.js
@@ -27,8 +27,8 @@
         _initTable = BootstrapTable.prototype.initTable;
 
     BootstrapTable.prototype.initData = function (data) {
-        _initData.apply(this, Array.prototype.slice.apply(arguments));
         flatJSON(this, data === undefined ? this.options.data : data);
+        _initData.apply(this, Array.prototype.slice.apply(arguments));
     };
 
     BootstrapTable.prototype.init = function () {


src/extensions/flatJSON/bootstrap-table-flatJSON.js
```diff
@@ -27,8 +27,8 @@
         _initTable = BootstrapTable.prototype.initTable;
 
     BootstrapTable.prototype.initData = function (data) {
-        _initData.apply(this, Array.prototype.slice.apply(arguments));
         flatJSON(this, data === undefined ? this.options.data : data);
+        _initData.apply(this, Array.prototype.slice.apply(arguments));
```
djhvscf: I'm not sure about this change.. Did you test it with this change and without it?

djhvscf:
I think this is ok

dimbslmh:
Yes I tested it. Without the change it will flatten the data after initData() is executed. And with will do that before.

src/bootstrap-table.js
```diff
@@ -602,7 +602,7 @@
         } else if (type === 'prepend') {
             this.data = [].concat(data).concat(this.data);
         } else {
-            this.data = data || this.options.data;
+            this.data = this.options.data || data;
```
wenzhixin: Why change the line of code?

dimbslmh: this.data = data || this.options.data;

Currently the sorting of data on refresh is always done against "data" (new data) that has not been processed by the flatJSON extension.

this.data (on line 622) in BootstrapTable.prototype.initSort depends on the chosen data on line 605. If this.data is not flattened yet, the sorting could be done against a data-field with dot notation that does not exist in this.data. Which ends up with the selected sorting lost on refresh.

    this.data = this.options.data (new flattened data from flatJSON) || data (new data)

PR #760, opened 2015-04-17
Fix #583 (How to hide / filter rows?)

### LINKED ISSUE #583: How to hide / filter rows?
Which method should we use for hiding (or filtering) specific rows?

### LINKED ISSUE #750: Add/Update row shows hidden rows ?!
Hi, I think this is a bug, if it's not, it's not a really cool feature ...
Each time I add a new row to the table or update an existing one, all my hiddens rows are shown

djhvscf:
Test: https://jsfiddle.net/w50myezm/1/


achaffman:
Dennis, does this mean the 1.7 release has a bug in it and needs to be patched with your change in order to work?  

Your JSfiddle example doesn't actually set uniqueID (data-unique-id) in the table definition, how is it working?

achaffman:
Nvmd, i missed it, it's in the table def, not in the js code.  I pulled your updated code and and replaced the .min.js file with your version and it works.  I guess this will make it to the next release version.  Thanks for figuring that out!

iTouchTheSky:
@djhvscf So you have to specify `uniqueId` and `idField` to get the `hideRow` to work with the option `isIdField` ? 
Sorry, but I don't see any logic here, calling the option `isIdField` means you need to have the `idField` specified, not both ..

djhvscf:
With the idField option the show/hide row is more complicated because we have to find the checkbox in the table that has the idField set.. But with the unique id option the tr element has the id of the row..

We are not using the idField any more

achaffman:
That makes sense - the uniqueId field is replacing idField.  Thanks for clarifying.  It did seem redundant.  Just remember the getselected method returns the idField right now, will have to return the uniqueId in the future

achaffman:
Another issue I noticed today is when you do a search, hidden rows become visible again.  There is nothing keeping track of what was hidden so it makes sense.  Is there any plans to change this behavior?

djhvscf:
Yes, I think @wenzhixin is working on it

PR #776, opened 2015-04-22
Added a 'getAllSelections' method to get checked rows across all pages and filters

thetomester13:
The 'getSelections' method doesn't work properly when you enable paging and/or have filters on your table. This method will get all selected rows.


wenzhixin:
Thanks!

PR #793, opened 2015-04-27
Fix #771 and Fix #786

### LINKED ISSUE #771: Indicating a column is sortable?
One of my users has complained that its not obvious that clicking the table header will sort the table by that column.

Is there an option to add an icon or some other indication for this on columns which are sortable? For example, maybe a double-headed variant of the sort icon currently used?

Thanks.

### LINKED ISSUE #786: bug getScrollPosition
$('#table).bootstrapTable('getScrollPosition');

it can‘t return the right value ,

and this code in bootstrap-table.js

    BootstrapTable.prototype.scrollTo = function (value) {
        var $tbody = this.$container.find('.fixed-table-body');
        if (typeof value === 'string') {
            value = value === 'bottom' ? $tbody[0].scrollHeight : 0;
        }
        if (typeof value === 'number') {
            $tbody.scrollTop(value);
        }

        if (typeof value === undefined) {
            return $tbody.scrollTop();
        }
   };

that is no return,  no run in "return $tbody.scrollTop();"

because the  'type value' is 'undefined', is  ('undefined' === undefined) is false 

so I think shoud change 

      typeof value === undefined 

to

      value === undefined

djhvscf:
Example for sortable icons: http://jsfiddle.net/e3nk137y/1375/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -10,8 +10,19 @@
     // TOOLS DEFINITION
     // ======================
 
-    var cellHeight = 37; // update css if changed
-    var cachedWidth = null;
+    var cellHeight = 37, // update css if changed
+        cachedWidth = null,
+        arrowAsc = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABMAAAATCAYAAAByUDbMAAAAZ' +
+        '0lEQVQ4y2NgGLKgquEuFxBPAGI2ahhWCsS/gDibUoO0gPgxEP8H4ttArEyuQYxAPBd' +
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1968,14 +1988,15 @@
         if (typeof value === 'number') {
             $tbody.scrollTop(value);
         }
-
-        if (typeof value === undefined) {
```
wenzhixin: Here is a typo: if (typeof value === 'undefined') {

wenzhixin:
:+1:

PR #810, opened 2015-04-28
Fixes #808

### LINKED ISSUE #808: Problem with hideRow
It seems when trying to use 

```$('table').bootstrapTable('hideRow', {index: 'row-0',isIdField: true})```

Nothing happens because bootstrap-table transforms: data-unique-id="row-1" into data-uniqueid="row-1"

Therefore the filter on line 1666 (https://github.com/wenzhixin/bootstrap-table/blob/master/src/bootstrap-table.js#L1666) never matches. If I change the filter to data-uniqueid it works as expected.

Using master branch

tm1000:
-fix hideRow and showRow functionality.
-Also update documentation to reflect proper id usage


djhvscf:
Great! thanls

wenzhixin:
Great!

Issue #835, opened 2015-05-04
formatter -> get the field (column)


thibaultvanc:
Hello, 
I am using this fabulous pluggin with xeditable. 
I wish to let users edit images. I know that's not work, but i have found a work around :
-> I show the url of the image in the editable field, 
-> I show the image just below using the formatter

the pb is that I need the column variable in the formater function in order to set the data-name attribute.

Here is the formatter function

    function show_image(value, row, index) {
        return [
            '<a  data-type="text" data-value="'+ value +'" data-pk="1" data-name="" href="javascript:void(0)" class="editable editable-click">'+ value +'</a><img src="'+ value +'" style="max-height:100px">'
        ].join('');
    };

djhvscf:
Please provide a jsfiddle, we can help you

djhvscf:
Any update here??

jtrumbull:
@thibaultvanc: You cannot presently back into the field name from the `formatter` function. However, you can reverse engineer this from an additional click event.

```javascript
$(document).on('click', '.editable-click', function(e){
        
        $a = $(this), 
            $td = $a.parent('td'),
            $table = $td.parent('tr').parent('tbody').parent('table'),
            $thead = $table.data('bootstrap.table').$header,
            $theadCell = $thead.find('th:eq(' + $td[0].cellIndex + ')'),
            field = $theadCell.data('field');
        
        // Your logic here
        
});
```
Working example: http://jsfiddle.net/joeycakes/98efstkq/2/

I agree though, the field name should be accessible from the `formatter` function.
<br />
@wenzhixin and @djhvscf: Is there any logic behind passing the `header` object for the context (this), opposed to the `column` object? The latter seems more intuitive and would solve this issue. I can create a pull request, but wanted to check first.

wenzhixin:
Because I think the field name is certain, for example:
```
{
    title: 'Name',
    field: 'name',
    formatter: function () {
        // you know the filed is 'name'
    }
}
```
Maybe some cases we need the field name, I will consider whether this param is necessary.

jtrumbull:
Agreed, the field name is understood when using a specific formatter for each column. However, I believe the example wants to use a global formatter -where the column would be a variable.

wenzhixin:
Usage:
```js
{
    title: 'Name',
    field: 'name',
    formatter: function () {
        console.log(this.field)
    }
}
```

manometer114:
Hello,

[...]

PR #910, opened 2015-05-22
Fix 835: Accessing field name in formatter

jtrumbull:
This pull request is to change the context (this) of the formatter function in column options. The present context is `header`, which I believe serves no practical purpose. The change will instead pass the `column` as the context, which I feel is more intuitive.

Some key points:

- More intuitive
- Allows end developer to access field name options at `this.field`
- Makes the function uniform with the `footerFormatter`

Fix was suggested in: [Issue # 835](https://github.com/wenzhixin/bootstrap-table/issues/835)


wenzhixin:
Great!

PR #921, opened 2015-05-25
Fix bug assigned sortable arrowArrows

dimbslmh:
Set empty string value for that.options.sortName, to prevent any assigned sortable arrows being set in initBody() > getCaretHtml()


dimbslmh:
Steps to reproduce bug:

1. Single sort a column.
2. Multiple sort columns other than the previous selected column in step 1.
Result: Assigned arrow of step 1 is still active.

PR #926, opened 2015-05-26
Allow for multiple arguments

jtrumbull:
Current approach only allows for a singular argument @`_relatedTarget` passed to public
method. The new approach would allow for any amount of arguments.

__Reasoning__
I'm working on an extension and adding some additional `allowedMethods`. The methods require multiple arguments, and would ideally not have to wrap them in an object.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2251,7 +2251,7 @@
         'toggleView'
     ];
 
-    $.fn.bootstrapTable = function (option, _relatedTarget) {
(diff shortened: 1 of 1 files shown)


djhvscf:
Then I think you can delete the _relatedTarget parameter

jtrumbull:
Getting some unexpected results, will re-open when solved.

djhvscf:
Ok.. go ahead!

jtrumbull:
Nested function strikes again.

PR #934, opened 2015-05-28
Update bootstrap-table-pt-PT.min.js

burnspirit:


wenzhixin:
Needn't modify the min file :)

PR #939, opened 2015-05-29
Fix for visibleColumns Cookie

murphmar:
Need to use field instead of column number to prevent issues when a column has been added to the table after a user already has a cookie in place.


djhvscf:
Exc! Thaniks

PR #977, opened 2015-06-09
A table with a fixed header clears the filtercontrol fields after filter runs

### LINKED ISSUE #905: A table with a fixed header clears the filtercontrol fields after filter runs
A table with a fixed header clears the fields after filtercontrol runs.  So, if a user enters a value into the filter the filter works correctly,  However, after the filter completes the filter fields clear out

### LINKED ISSUE #972: Table Filter Control on modal windows with data-height
Hi @wenzhixin 

I could not see the filter row (underneath the header) when I am using the table filter control in a modal of my MVC application. F12 clearly shows that the data is existing. 
I found that if I remove the data-height attribute, then the filter works correctly.
(Of course, I have to turn on the pagination in this case)
So the issue is filter row cannot be seen when a value is set on data-height.
I am using version 1.8.0

When I tried it with 1.8.1, there is an error
Unable to get property 'options' of undefined or null reference 
Line of error -> selectControl.get(0).options

djhvscf:
Fix #905 
Fix #972


wenzhixin:
What features do `group-by` extension have? There is a typo in `groubby.js`.

djhvscf:
Sorry.. My mistake let me delete it

djhvscf:
Ready :)

wenzhixin:
Great!

PR #985, opened 2015-06-10
Fix #984 Filter Control extension broken when we hide a column

### LINKED ISSUE #984: Filter Control extension broken when we hide a column
See this example: http://jsfiddle.net/djhvscf/e3nk137y/2009/

Hide 'Name' column and the filter control is broken

djhvscf:
Fix #984 
Example: http://jsfiddle.net/djhvscf/e3nk137y/2011/


wenzhixin:
Good!

PR #993, opened 2015-06-13
Improve filter-control, cookie, multiple-sort and flat-json extensions

### LINKED ISSUE #959: Multiple-sort with multiple tables is not working
(By @rajeshbtlit)  In my scenario, i have 3 tables, and in first table multiple sort works fine no problem.
When i add it to second table, i see that the same sort modal opens (I mean the sort window of the first table)

JS fiddle here@ https://jsfiddle.net/jupitor/h2c683cn/

I have created two tables and added the multiple sort attribute for both and the second table always shows the first tables sorting columns

djhvscf:
filter-control: fix issues related to load the select control
cookie: names changed and added path, domain and secure options
multiple-sort: fix #959 
flat-json: added flatSeparator option and the code is more clear

### DIFF
--- a/src/extensions/cookie/README.md
+++ b/src/extensions/cookie/README.md
@@ -10,20 +10,38 @@ Use Plugin: [bootstrap-table-cookie](https://github.com/wenzhixin/bootstrap-tabl
 
 ## Options
 
-### stateSave
+### cookie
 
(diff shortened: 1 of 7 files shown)


wenzhixin:
Great!

PR #1007, opened 2015-06-17
Fix #980

### LINKED ISSUE #980: Detail view, when searching, returns wrong row
Using version 1.8.1, windows 7 32 bits, Browsers: All
As I said in the title, if you search and then press the + sign, you'll get the wrong info returned. As english isn't my nature tongue let me present you with a example.

    onExpandRow: function (index, row, $detail) {
        console.log(row);
    }

The above 'row' doesn't return the information that is inside that row.

Here's a fiddle!:
http://jsfiddle.net/e3nk137y/1997/

To reproduce, search for example '02' and click on the + thingy. I've console.log'd it.

OfficialBAMM

For people needing a solution asap:

    var temp = $detail.context;
    var debID = $(temp).parent().parent().children(".debID").html();

I've added a class on the ID i needed to get, maybe it's handy for someone.

tannermccoleman:
Fix #980 Detail view, when searching, returns wrong row

```initBody``` was using ```that.options.data``` (the full table data) instead of ```data``` (the current filtered data) to form the Detail view

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1452,7 +1452,7 @@
             var $this = $(this),
                 $tr = $this.parent().parent(),
                 index = $tr.data('index'),
-                row = that.options.data[index];
+                row = data[index]; // Fix #980 Detail view, when searching, returns wrong row
 
             // remove and update
             if ($tr.next().is('tr.detail-view')) {


wenzhixin:
Thanks

jogo:
I just hit this bug, when will 1.8.2 be out with the fix?

PR #1013, opened 2015-06-17
Fix #981: Grep-style filter callback function

### LINKED ISSUE #981: Grep-style filter callback function?
Is there any way to have a grep-style callback function for manual filtering? IE return false to filter out an item, return true to keep it

tannermccoleman:


wenzhixin:
Can you provide an example to show the feature like #1014?

tannermccoleman:
Sorry, I rushed trying to get this done and neglected to test it! For now it only works when the search is triggered, and it causes sorting to break. I'm trying to come up with a better solution today.

djhvscf:
Ok, please provide an example when you are done

tannermccoleman:
[Here's a working example of my filterRows option](http://jsfiddle.net/3yrbze3u/6/). 

I've created a new method called "refilterRows" which just runs ```BootstrapTable.prototype.initSearch``` and ```BootstrapTable.prototype.updatePagination``` again.

wenzhixin:
It seems have problem after sort, steps:
1. Click Starts to sort by asc
2. Enter Minimum Stars: 200
Please help to check.

src/bootstrap-table.js
```diff
@@ -969,6 +974,7 @@
         if (this.options.sidePagination !== 'server') {
             var s = this.searchText && this.searchText.toLowerCase();
             var f = $.isEmptyObject(this.filterColumns) ? null : this.filterColumns;
+            var r = (typeof this.options.filterRows === "string");
```
wenzhixin: When we via JavaScript:
```js
bootstrapTable({
    filterRows: function () {}
});
```
Here will cause error.
BTW, please use `'` instead of `"`.

wenzhixin:
Need to update this PR

PR #1014, opened 2015-06-18
Added support for title attribute in the TH and TD tag

### LINKED ISSUE #1005: Support for the HTML5 `title` attribute
To show a small descriptive text on hover, one can use the `title` attribute on a `<th>`, like so :

``` html
<th data-sortable="true" title="This column is awesome">My Header</th>
```

This attribute is lost/ignored when using `bootstrap-table`.
The `data-title` attribute overrides the contents of the `<th>` ("My Header" in the above example), which is not what I want.

How can one set the hover title ?

djhvscf:
Fix #1005 
Example [here](http://jsfiddle.net/djhvscf/e3nk137y/2112/)


EvanLiu2968:
How to support for title in the TD tag ?

PR #1022, opened 2015-06-19
Updated cookie extension with selectPage Method.

Xandertje:
Updated cookie extension to extend the selectPage Method so that it updates the state save page number cookie.


djhvscf:
Can you provide an example of this improvement?

Xandertje:
When the selectPage method is called, the pageNumber cookie is not changed. I needed to add this code in order to get it to work.

wenzhixin:
Can we merge this pull request @djhvscf

djhvscf:
Yes, we can. @wenzhixin

PR #1034, opened 2015-06-22
Fix  Reorder column order issue when column is invisble

### LINKED ISSUE #1031: Reorder column order issue when column is invisble
1- Hide the column using columns menu and then reorder column and if you visible the hidden column it will show as first column, is there any configuration that column should show as last column instead of first.
2- Is there any configuration, so column like checkbox and radio button can not be dragged?

djhvscf:
Fix #1031 
Example for #1031 [here](http://jsfiddle.net/djhvscf/e3nk137y/2181/) and [here](http://jsfiddle.net/djhvscf/e3nk137y/2194/)
Adding the group-by extension but it is in testing


haideraliarain:
Thank you

PR #1057, opened 2015-06-25
Fixes #1051 where HTML chars shown in filter-control select

### LINKED ISSUE #1051: Filter-control type 'select' gets all string with html tags

tmorehouse:
Removes HTML characters from appearing in SELECT dropdowns.


wenzhixin:
Great

PR #1059, opened 2015-06-25
Add Persian (fa-IR).

MJ-Vakili:


src/locale/bootstrap-table-fa-IR.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Bootstrap Table English translation
+ * Author: Zhixin Wen<wenzhixin2010@gmail.com>
```
wenzhixin: update the author information please.

PR #1073, opened 2015-06-26
Added Strict search option, group-by extension

### LINKED ISSUE #694: Group and Ungroup table rows based on column name
In the boot strap table, group and un-group table rows based on column name. For example in a given table below grouping is done using column  workflow name

table
groupby

### LINKED ISSUE #1066: Strict filtering required.
If I've got few columns with data for example [{"column": 1}, {"column": 14}, {"column": 18}, {"column": 21}]. So in filter I select "1", it will show me every row. In this case I propose to add strict filtering. So I am as user will be able to filter data and get rows contains only number 1.

djhvscf:
Fix #1066 
Example [here](http://jsfiddle.net/djhvscf/e3nk137y/2247/)

Example for group-by extension [here](http://jsfiddle.net/djhvscf/e3nk137y/2258/)
Fix #694


gowthamsathiya:
@djhvscf instead of specifying data-group-by-field statically as an attribute to table, can we specify dynamically such that selecting the attribute from dropdown and making groupBy over that selected attribute?

djhvscf:
@gowthamsathiya See this: https://github.com/wenzhixin/bootstrap-table/pull/1381

PR #1083, opened 2015-06-28
fixed column state cookie [issue 1078]

jetsaredim:
This should fix issue 1078 by adding the correct reference to the columns array


wenzhixin:
In latest version, have changed columns from `this.options.columns` to `this.columns`.

jetsaredim:
Ah, ok.  Version is still listed as 1.8.1 in all of the sources but that doesn't match the 1.8.1 that's out in the cdn hence my confusion.  I usually don't like to host js libs like this but I guess I'll have to make a local copy of both the extension and the main lib...  Sorry for the clutter then.

PR #1117, opened 2015-07-02
Don’t wait when applying searchText from options

### LINKED ISSUE #1116: Don't use searchTimeOut when loading data with searchText
Right now, when data loads, you can see the search query, as well whole data, after after 500ms it applies the search and the data is filtered. `searchTimeOut` should only throttle the search input.

dbkaplun:
Closes #1116


wenzhixin:
Thanks!!

PR #1129, opened 2015-07-05
Adds directive-updated event

dbkaplun:
Adds directive-updated event
Performance improvements — use event delegation


wenzhixin:
In fact, you don't need to modify other files, in addition to `src/extensions/angular/bootstrap-table-angular.js`

PR #1133, opened 2015-07-06
Cookie extension supports the filter-control extension

### LINKED ISSUE #1101: Want sample example with filter control and cookie
Hi,
Im using filter control and cookies in my bootstrap table but combining both not working please provide me any sample example with both the feature.. I already found separate examples for filter control and cookie

### LINKED ISSUE #1104: cookie extension
Hi,
is it possible to add cookie support for the resizable and reorder-columns extension. Because so can make the user his own table layout.

And any idea why my cookie ext. don't save the hidden or visible columns. All other "cookie ext." features work without problems.

thanks
andre

djhvscf:
Fix #1104 
Fix #1101 

Example [here](http://jsfiddle.net/djhvscf/081Lma9u/)

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -385,9 +385,6 @@
         },
         onResetView: function () {
             return false;
-        },
-        onPostInit: function () {
-            return false;
         }
     };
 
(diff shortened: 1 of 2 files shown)


andresendo:
any chance to save the resize and reorder extensions? :)

PR #1152, opened 2015-07-09
Fix cookieExpire to accept numbers > 9

MichaelMackus:
Without this a expire string such as "14d" gets parsed as "4d"


wenzhixin:
Thanks

PR #1165, opened 2015-07-12
Incorrect footer width when using Bootstrap 2

petr-savintsev:
When a table has a footer, there is a problem with setting the width of footer table. Here's the example: https://jsfiddle.net/2h9b8s8q
Columns in the footer table are slightly shifted. The problem is only when using Boostrap 2. With Boostrap 3 everything is OK.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1894,7 +1894,7 @@
         this.$tableBody.find('tbody tr:first-child:not(.no-records-found) > td').each(function (i) {
             $footerTd.eq(i).outerWidth($(this).outerWidth());
         });
-        if (this.$tableFooter.find('table').width() < this.$tableBody.find('table').innerWidth())
+        if (this.$tableFooter.find('table').width() != this.$tableBody.find('table').innerWidth())
         {
             this.$tableFooter.find('table').width(this.$tableBody.find('table').innerWidth());
         }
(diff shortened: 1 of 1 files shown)


wenzhixin:
Can you provide an example?

petr-savintsev:
Here is the example:
http://jsfiddle.net/2h9b8s8q

wenzhixin:
Sorry, I haven't notice your jsFiddle. Thanks.

src/bootstrap-table.js
```diff
@@ -1894,6 +1894,10 @@
         this.$tableBody.find('tbody tr:first-child:not(.no-records-found) > td').each(function (i) {
             $footerTd.eq(i).outerWidth($(this).outerWidth());
         });
+        if (this.$tableFooter.find('table').width() != this.$tableBody.find('table').innerWidth())
```
wenzhixin: I think it's better to change it to:
```js
if (this.$tableFooter.find('table').width() != this.$tableBody.find('table:eq(0)').innerWidth()) {
    this.$tableFooter.find('table').width(this.$tableBody.find('table:eq(0)').innerWidth());
}
```
When the table body has sub table maybe cause error, what do you think so?

petr-savintsev: Well as I tested a little bit, my fix is not good at all. It solves the problem for my case in Bootstrap 2, but it creates the same problem for Bootstrap 3. So I think it should be better not to use my fix and find a better solution. Maybe I should create an issue for this bug?

wenzhixin:
OK, please create an issue.

wenzhixin:
Close this PR

PR #1185, opened 2015-07-16
Support two new events in the editable plugin

hajekj:
For my project, and for others too (I believe), it would be beneficiary to add support for the two additional events in the x-editable plugin - `shown` and `hidden` so we could do even more with the plugin, since this is the easiest way to bind events to it.
The readme for plugin is also updated with the changes.

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -15,7 +15,7 @@
         onEditableSave: function (field, row, oldValue, $el) {
             return false;
         },
-        onEditableShown: function (field, row, $el) {
+        onEditableShown: function (field, row, $el, editable) {
             return false;
         },
         onEditableHidden: function (field, row, $el) {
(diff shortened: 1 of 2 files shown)


src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -34,7 +42,7 @@
             return;
         }
 
-        $.each(this.columns, function (i, column) {
+        $.each(this.options.columns, function (i, column) {
```
wenzhixin: In our latest version, because our plugin support `rowspan/colspan` table header, so we have to use `this.columns` to replace `this.options.columns`.

src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -61,7 +69,7 @@
...
```
wenzhixin: The same.

wenzhixin: The same.

hajekj:
Currently using it in my project along with [At.js](https://github.com/ichord/At.js) and it works very smoothly.

wenzhixin:
Merged it, thanks very much! 
You can help to provide an example to our [examples repo](https://github.com/wenzhixin/bootstrap-table-examples) if you have :)

hajekj:
I will try to make some examples in near future.

PR #1201, opened 2015-07-21
Reorder Row issues

nours:
Hi,

I'm using reorder-row extension to reposition entities, but I've encountered an issue upon BootstrapTable.prototype.onDrop, which do not pass the correct arguments to the onReorderRowsDrop option callback (which should be the table and the moved row, as documented), but only the last row instance.

While making a fiddle, I've seen that the plugin do not work with table having client side data, because the init function override plugs onPostBody after calling parent.

I've fixed these two issues in this pull request. Here is a fiddle with the reordering not starting : <a href="http://jsfiddle.net/c74h2zoL/1/">http://jsfiddle.net/c74h2zoL/1/</a>, and with the initial issue on a server side table : <a href="http://jsfiddle.net/5gobzuff/2/">http://jsfiddle.net/5gobzuff/2/</a>.

### DIFF
--- a/src/extensions/reorder-rows/bootstrap-table-reorder-rows.js
+++ b/src/extensions/reorder-rows/bootstrap-table-reorder-rows.js
@@ -91,7 +91,7 @@
         });
     };
 
-    BootstrapTable.prototype.onDrop = function (table, row) {
+    BootstrapTable.prototype.onDrop = function (table, droppedRow) {
         var tableBs = $(table),
             tableBsData = tableBs.data('bootstrap.table'),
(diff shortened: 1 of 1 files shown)


wenzhixin:
Help to check this PR:) @djhvscf

djhvscf:
@wenzhixin it is ok

wenzhixin:
Thanks! @nours

tm1000:
This actually breaks server side processing

tm1000:
To work with server-side it needs to be first (init) to work with client side it needs to be last (init)

Issue #1078, opened 2015-06-26
Filter-Control & Show/Hide Columns


dfmaxwell:
I seem to have two issues with the show/hide columns and it could be directly related to the filter-control since I am using that.  When Hiding/Showing a column the headers will change alignment compared to the cells below.  I've implemented the Cookie Extension so a refresh will have the column shown or missing and things will be aligned correctly.  Thanks for the quick fix on the cookie extension.

The second issue would be that after the show/hide I get inconsistent filters, most of the select elements are empty, so I can't use them without refreshing the page.

djhvscf:
can you provide a fiddle?? I think that issue is already fixed

djhvscf:
Any update here?

jetsaredim:
I submitted a pull request for this issue.  It should allow the column select state to be saved.

wenzhixin:
Please provide a jsFiddle to show your problem.

dfmaxwell:
I've not been able to get a jsFiddle working yet.  My implementation is server based and I can't expose that data externally.  My first shot at a jsFiddle also rendered completely wrong.

dfmaxwell:
I am trying to get an example running and have made some progress.  This example: http://jsfiddle.net/vbrhrbzL/ is not fully functional, The filters aren't populating from the variables, any thoughts on how to get those working so I can keep working on a jsFiddle?  Also I've noticed that even though they aren't working if I hide/show columns the HF Complete and Online are wiped out, initially they have 1 blank dropdown element, afterwards they have no dropdown element.

djhvscf:
Let me check what is happening

rhegner:
I also have issues with the filter-control extension and hidden columns. I think it is related to this issue here.

You can find a very simple jsfiddle here:
https://jsfiddle.net/6ngd2kcg/

As you can see in the HTML code I want to have a filter on "Column 2". But since "Column 1" is invisible, the filter is shown on "Column 3".

djhvscf:
This issue is fixed in the 1.10.1

PR #1317, opened 2015-08-20
Fixe #1078 - the correct selectControl is updated when show/hide column

yenyen:


wenzhixin:
The same with #1327?

yenyen:
No it's not the same that #1327. 

This pull request resolve an issue when you hide a column then it's not the right filter-control who is updated. Please go to this jsfiddle and try to hide the name's column, you'll see that the filter-control of the item price's column is missing : http://fiddle.jshell.net/Lvo3wf2q/

This pull request fixes this problem

wenzhixin:
20 Commits and 187 Files changed... I think do you make a mistake?

yenyen:
Sorry, please see this pull request : #1383

PR #1327, opened 2015-08-24
Restore focus of filter-controls after clone headers

yenyen:
Given the table height is set
When a search with a filter-control is done 
Then the headers are clone and replace 
And the focus of the filter-control is lost

This pull request try to avoid this situation, by restoring the focus after cloning the headers. The result is not perfect, because the focus is lost during a few ms. And, if the page contains multiple tables with same data-field, then it can be the false control who has focus.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1789,6 +1789,20 @@
             getScrollBarWidth() : 0;
 
         this.$el.css('margin-top', -this.$header.outerHeight());
+
+        if($(':focus').length > 0) {
+            var $th = $(':focus').parents('th');
+            if($th.length > 0) {
+                var dataField = $th.attr('data-field');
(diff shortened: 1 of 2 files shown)


wenzhixin:
Can we add this feature to `filter-control` extension? @yenyen

wenzhixin:
@djhvscf

djhvscf:
I have to double check this change

yenyen:
@wenzhixin No I don't think we can add this to the filter-control extension.
@djhvscf Yes, please. My code is not very good. But, it fixes a big issue because the filter-control extension is useless if the focus is lost after each search. Maybe you can find a better solution to this problem

djhvscf:
Close this PR. See this https://github.com/wenzhixin/bootstrap-table/pull/1370

Issue #1318, opened 2015-08-20
Reordering Columns does not switch header formatter functions.


Cowdrey:
This is apparent when using the editable plugin along with the reordering columns.  Immediately after reordering an editable and non-editable column, columns are correct. After filtering any column, the initBody(true) function fires, and it switches all header formats to the the original format of that column index. 
 I am currently working on a solution, i think it should be be as simple as :
```javascript
var temp = that.header.formatters[colA];
that.header.formatters[colA] = that.header.formatters[colB];
that.header.formatters[colB] = temp;
```
will update if it works.

Cowdrey:
Possible duplicate of https://github.com/wenzhixin/bootstrap-table/issues/1313

Cowdrey:
Created a fix for this, I don't know why I thought a swap was all I needed when you have to shift all the columns.

In the bootstrap-table-reorder-columns.js where you set the dragtable options, I changed the beforeStop function to take a variable:

```javascript
beforeStop: function (reorderInfo) {
```

And I added this to the top of the function:
```javascript
                var startMove = reorderInfo.startIndex - 1;
                var endMove = reorderInfo.endIndex - 1;
                var tempFormatter = that.header.formatters[startMove];

                if (startMove < endMove) {
                    for (var i = startMove; i < endMove ; i++)
                    {
                        that.header.formatters[i] = that.header.formatters[i + 1];
                    }
                }
                else
                {
                    for (var i = startMove; i > endMove ; i--)
                    {
                        that.header.formatters[i] = that.header.formatters[i - 1];
                    }

                }
                that.header.formatters[endMove] = tempFormatter;
```
Seems to be working.

djhvscf:
Can you provide a fiddle in order to test your change and your bug?? @Cowdrey

Cowdrey:
Here is the bug, you can observe it after a reordering a column by one, and then applying a filter. The formatters are applied to their original column indexes.
http://jsfiddle.net/fhsqydu4/2/

Here it is after the fix:
http://jsfiddle.net/fhsqydu4/3/

Cowdrey:
Also I tested SimaoLiu's solution in https://github.com/wenzhixin/bootstrap-table/issues/1313 , and that seems to work as well: 
http://jsfiddle.net/fhsqydu4/4/

It is a lot less code than my solution. I feel like my solution might be slightly more efficient, but it is insignificant.

Cowdrey:
[...]

Issue #1313, opened 2015-08-19
Reorder Columns extension - data formatter is applied on the wrong column after sorting


SimaoLiu:
If enabled re-ordering columns feature with the provided extension, plus a few columns are formatted with data-formatter, and sorting feature is also enabled.
To reproduce:
1) Re-order a column
2) Sort on any column

The data-formatter will be applied to a wrong column.

djhvscf:
Close this issue. Same of #1318

SimaoLiu:
I was trying to fix it by modifying the beforeStop function by adding an array to track the formatters along side the fields when re-ordering columns:
<img width="464" alt="screen shot 2015-08-23 at 1 09 57 pm" src="https://cloud.githubusercontent.com/assets/6552184/9429994/8f00af98-4998-11e5-9d13-569181ff6f76.png">

PR #1336, opened 2015-08-26
Fixed bug 1318

Cowdrey:
Fixed bug https://github.com/wenzhixin/bootstrap-table/issues/1318 with solution proposed by @SimaoLiu in https://github.com/wenzhixin/bootstrap-table/issues/1313


wenzhixin:
Thanks

Issue #1314, opened 2015-08-19
Reorder Columns does not offset header when dragging inside of scrollable container.


Cowdrey:
I also found the issue on the dragtable github: https://github.com/akottr/dragtable/issues/37

But that repository hasn't been updated in 8 months, so I don't know if it will ever be fixed on their end.

Cowdrey:
http://jsfiddle.net/fhsqydu4/5/ jsfiddle showing bug. Scroll as far right as possible and then attempt to move column.

Cowdrey:
http://jsfiddle.net/fhsqydu4/6/ here is my fix.
Added ```javascript 'style="position:relative;"' ```
to  ```javascript '<div class="fixed-table-body">' ``` in the init container function

djhvscf:
See this pull request https://github.com/wenzhixin/bootstrap-table/pull/1337

slampenny:
Tracked this issue down from the dragtable repo. @Cowdrey, you are a lifesaver. Thanks!

PR #1337, opened 2015-08-26
Fix for bug 1314

Cowdrey:
Fix for bug https://github.com/wenzhixin/bootstrap-table/issues/1314 where reordering columns inside scrollable container isn't offset correctly.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -523,7 +523,7 @@
                     '',
                 '<div class="fixed-table-container">',
                 '<div class="fixed-table-header"><table></table></div>',
-                '<div class="fixed-table-body">',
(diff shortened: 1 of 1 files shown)


wenzhixin:
Hi @Cowdrey, I think it is not a bug in our plugin, when you using this feature, you can just only to add a custom style to your css style to solve this bug:
```css
.fixed-table-body {
    position: relative;
}
```
Do you think so?

If the style does not cause other problem, it is better to add to css file: https://github.com/wenzhixin/bootstrap-table/blob/master/src/bootstrap-table.css#L50.

djhvscf:
I'm agree with @wenzhixin

djhvscf:
Can we close this issue?

PR #1365, opened 2015-09-04
ColumnSearch after click on clear button in IE

yenyen:
Internet Explorer adds a clear button in input. Unfortunately the columnSearch is not trigger when the clear button is pressed.

To fixe this, we have to listen to the 'mouseup' event trigger par the clear button (http://stackoverflow.com/questions/14498396/event-fired-when-clearing-text-input-on-ie10-with-clear-icon)

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -246,6 +246,24 @@
                 }, that.options.searchTimeOut);
             });
 
+            header.off('mouseup', 'input').on('mouseup', 'input', function (event) {
+                var $input = $(this),
+                oldValue = $input.val();
+
+                if (oldValue == "") return;
+
+                setTimeout(function(){
+                    var newValue = $input.val();
+
+                    if (newValue == ""){
+                        clearTimeout(timeoutId);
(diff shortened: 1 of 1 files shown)


src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -246,6 +246,24 @@
                 }, that.options.searchTimeOut);
             });
 
+            header.off('mouseup', 'input').on('mouseup', 'input', function (event) {
+                var $input = $(this),
+                oldValue = $input.val();
+
+                if (oldValue == "") return;
```
djhvscf: Please use the equal value and equal type validation and the correct format

if (something === thing) {
...
}

src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -246,6 +246,24 @@
...
+
+                setTimeout(function(){
+                    var newValue = $input.val();
+
+                    if (newValue == ""){
```
djhvscf: Please use the equal value and equal type validation

yenyen:
Done ;)

djhvscf:
Great!

PR #1374, opened 2015-09-06
add data-data-field option for compability with different REST backends

bashlakov:
Ability to customize this option will let more simply integration with backend. Default value is backward-compatibility with current realization.


wenzhixin:
Can you help to update the docs

bashlakov:
Yes, but I can't understand, which files I need to edit. Could you point me?
Also my English is very poor, and don't know Chinese and Spanish :(

djhvscf:
This is the document that you have to update https://github.com/wenzhixin/bootstrap-table/blob/master/docs/_i18n/en/documentation/table-options.md

bashlakov:
Ready. How I can locally see docs in browser?

wenzhixin:
Thanks!

PR #1383, opened 2015-09-08
Fixe #1078 - the correct selectControl is updated when show/hide column

yenyen:
Sorry, I sync the false branch in #1317...


wenzhixin:
Thanks!

PR #1394, opened 2015-09-09
Fix #1359 Cookie extension not working with default settings

### LINKED ISSUE #1359: Cookie extension not working with default settings
Both getCookie and setCookie changes the cookie name to lowercase before getting/setting it. However, **by default the cookie names are camelcase**, so the match always fails. Thus, none of the features work (except for `bs.table.columns`, which is lowercase by default).
One is forced to pass the `cookiesEnabled` option with all lowercase names to make it work.

djhvscf:
Fix #1359

Example for Multiple-search extension: http://jsfiddle.net/e3nk137y/3046/ 

Example for Accent-Neutralise extension: http://jsfiddle.net/e3nk137y/3055/

Example for accessibility: http://jsfiddle.net/e3nk137y/3062/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -710,6 +710,7 @@
                     sprintf(' rowspan="%s"', column.rowspan),
                     sprintf(' colspan="%s"', column.colspan),
                     sprintf(' data-field="%s"', column.field),
+                    "tabindex='0'",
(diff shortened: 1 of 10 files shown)


wenzhixin:
Good job, thanks @djhvscf .

PR #1400, opened 2015-09-10
added af_ZA locale

phillip-kruger:
Hi
Added one js file that contains the localization for af_ZA (Afrikaans - South Africa)


wenzhixin:
Great, thanks!

Issue #1270, opened 2015-08-10
Option export all dont work with server side pagination


geasi:
The option export "all" dont work with server side pagination.
The data exported is always the current page, not all the records.
It works when I changed to client side pagination, but it's not recommended to my application.

Need help!

wenzhixin:
The same with #1242.

Issue #1242, opened 2015-08-03
Unable to Export data correctly with pagination (Plugin extension)


quchie:
Hi,

I am trying to implement selection on table then export it. The selection will also have fixes from issue #197 which keeps the selection after clicking next page.

But i cant seem to get the correct export data. By the way, the getAllSelections() will always returned selected data on the current page although you have selected something on the previous page.

Here are the sample fiddle:
http://jsfiddle.net/4r6g4cfu/90/

wenzhixin:
I think we need to handle it in server side pagination.

geasi:
I have this problem too in server side pagination.

wenzhixin:
Merged.

PR #1424, opened 2015-09-15
Fix issue #1270 an #1242 

raku:
resolve export when data is loaded from server, see issue #1270


raku:
@wenzhixin, @djhvscf  can you check this patch please?

PR #1445, opened 2015-09-22
Check length when comparing in refreshOptions

robpvn:
If we don't check length in compareObjects, we won't notice when the new
options object has had a new property added. (Because compareObjects
otherways only checks that everything present in the old options exist
in the new, not the other way round.) This might be especially relevant
when you want to activate an extension like reorderableRows.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2463,7 +2463,7 @@
 
     BootstrapTable.prototype.refreshOptions = function (options) {
         //If the objects are equivalent then avoid the call of destroy / init methods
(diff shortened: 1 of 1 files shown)


djhvscf:
Can you provide a fiddle? @robpvn

dabros:
@robpvn - please provide fiddle to prove changes, then make PR for develop branch

https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md

thanks

robpvn:
Done and done, sorry for taking so long to react.

PR #1460, opened 2015-09-24
Correct find usage

chpdesign:
Hello,

I'm using this plugin with "table in table". If one table has an other in it, the th's are not in the corrent position and some td's slide out from the calculated width. Please use the "find" method correctly, because "find('td')" finds all the td's in the DOM if you use "find('>td')" it's going to only find the children, not all of the objects.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -577,7 +577,7 @@
             columns = [],
             data = [];
 
-        this.$header = this.$el.find('thead');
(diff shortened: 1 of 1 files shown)


wenzhixin:
Can you provide a jsFiddle to show the problem? Thanks.

chpdesign:
I would like to show it, but I can not initialize a demo on jsfiddle, sorry. Too complex my code for it. But i can show you, the problem on a picture.
preview

PR #1479, opened 2015-10-01
fix #1130 (table-condensed is not working)

### LINKED ISSUE #1130: table-condensed is not working
Similar to #44.

The class itself is applied, but bootstrap-table has an override rule with `!important` that breaks the condensed style:

screen shot 2015-07-05 at 07 07 03

geeyan:
Note: only tested on Chrome/45.0.2454.101.

> line eding not easy to set on web page..
image

### DIFF
--- a/src/bootstrap-table.css
+++ b/src/bootstrap-table.css
@@ -11,12 +11,12 @@
     border-radius: 1px;
 }
 
-.bootstrap-table .table,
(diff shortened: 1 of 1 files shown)


djhvscf:
Great!

wenzhixin:
Thanks

PR #1490, opened 2015-10-02
Changed array function when deciding to select all in pagination

wilky27:
Changed to use $.inArray rather than indexOf so it works in IE8 and older


wenzhixin:
the same with #1495, merged,Thanks

PR #1492, opened 2015-10-02
Fixes #1248 - IE8 does not support indexOf()

### LINKED ISSUE #1248: Bug with pagination in IE8
I have bootstrap table
```
<table class="main-table"
           data-toogle="table"
           data-pagination="client"
           data-url="/json/data1.json"
           data-show-refresh="true"
           data-page-list="[10, 25, 50, 100, ALL]">
  <thead>
   ....
 </thead>
</table>
```

Inside the file there are 11 entries. When you delete records from the file and update the table
```
  $(".main-table").bootstrapTable("refresh");
```

There is an error on the line in the bootstrap-table.js [Line 1014](https://github.com/wenzhixin/bootstrap-table/blob/master/dist/bootstrap-table.js#L1014)

The method is not supported.In other browsers such an error does not occur, only in IE8. At the same time when adding a string to a file or to remove another, further it works fine. How can I avoid this prompt?

thornomad:
IE8 does not support the Array.prototype.indexOf().  Replaced with jQuery.inArray().

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1151,7 +1151,7 @@
                 var pageLst = typeof this.options.pageList === 'string' ?
                     this.options.pageList.replace('[', '').replace(']', '')
                         .replace(/ /g, '').toLowerCase().split(',') : this.options.pageList;
-                if (pageLst.indexOf(this.options.formatAllRows().toLowerCase()) > -1) {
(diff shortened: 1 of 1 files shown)


djhvscf:
Please open a new PR against "develop" branch, please @thornomad

thornomad:
Okay, done.

PR #1502, opened 2015-10-05
Fix column issue with cookie-extension

tareksamni:
Show/Hide columns is not persisting as it should using cookie extension because of an out of date objects naming.


tareksamni:
@wenzhixin are you still following :) ??

tareksamni:
@djhvscf  you ?

djhvscf:
Hi @tareksamni 

Since we implemented the rowspan and colspan we use it in this way: (this.columns)

So if you have problem please update bootstrap-table.js file to v1.9.0

PR #1514, opened 2015-10-07
Update row uniqueid

thornomad:
What a terrible title for this!  It should be:

Expand `updateRow` params to accept `uniqueId`
-----------------------------------------------------------------

This pull request adds a new feature to the method `updateRow`.  It now allows the param of `uniqueId` in addition to `index` for the row to update.  

This is particularly useful if you are updating a row which is not currently visible to the user (therefor, no index is available for it).  It will update the row in the background and the updates will be displayed when the user changes their view.

I created a fiddle here: http://jsfiddle.net/thornomad/ooh23sdz/3/

Let me know if you have any questions and am hopeful it can be included.  I find it very useful.

### DIFF
--- a/docs/_i18n/en/documentation/methods.md
+++ b/docs/_i18n/en/documentation/methods.md
@@ -108,7 +108,7 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter);`.
         <td>params</td>
         <td>
         Update the specified row, the param contains following properties: <br>
-        index: the row index to be updated. <br>
+        index or uniqueId: the row index/uniqueId to be updated. <br>
         row: the new row data.
         </td>
     </tr>
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2216,10 +2216,21 @@
     };
 
     BootstrapTable.prototype.updateRow = function (params) {
-        if (!params.hasOwnProperty('index') || !params.hasOwnProperty('row')) {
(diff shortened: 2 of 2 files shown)


sketchman-nyt:
Just had to implement this same feature myself, hope this is merged soon.

:+1:

thornomad:
Seems I'm not the only one who wanted this feature! @sketchman-nyt #1523 

These should accomplish the same thing; my pull request simply combines the functionality into the current `updateRow` function (allowing for an extra `param`) while @sketchman-nyt adds a new function.

The #1523 pull request uses `indexOf` which is not compatible with IE<9; if you are accepting that request (instead of this one) please consider changing that line to use `$.inArray()`

Avivbh:
This update is great. Could also be useful to update "updateCell" method in the same way.

wenzhixin:
Consider #1523.

thornomad:
Yea - I saw #1523 and now #1592 ... that is a fine way to do it.  I had kept the same function method name (`updateRow`) because I saw some other methods accepted either `index` or `uniqueId` and was keeping with that sentiment.  Really doesn't matter though ... totally up to you!

wenzhixin:
Merged in #1592, because there are `getRowByUniqueId` and `removeByUniqueId` methods, thanks!Merged in #1592, because there are `getRowByUniqueId` and `removeByUniqueId` methods, thanks!

PR #1526, opened 2015-10-09
Expand `filterBy` to accept and array of values

thornomad:
This pull requests allows a user to specify an array of values when using `filterBy`; this is really useful when you need to create a more complex filter.

    $table.bootstrapTable('filterBy', {
        age: [10, 11, 12]
    });

I also updated the documentation and created a fiddle to test it:

http://jsfiddle.net/thornomad/1xrtz4do/


wenzhixin:
Thanks!

PR #1527, opened 2015-10-09
Use default settings when cookie settings dont exist

tm1000:
When no cookie setting exists use the default setting provided by bootstrap table (or the table itself) for sort name and sort order


wenzhixin:
Please commit to the develop branch :)

PR #1531, opened 2015-10-10
Use default settings when cookie settings dont exist

tm1000:
When no cookie setting exists use the default setting provided by bootstrap table (or the table itself) for sort name and sort order


tm1000:
To develop as requested in #1527.

This brought over the CONTRIB,md changes as well. Hope that is ok

wenzhixin:
Thanks!

PR #1572, opened 2015-10-19
sortName fails in cookie extension

KlapZaZa:
There is an issue from the cookie extension.

at line 249:
"this.options.sortName = sortOrderNameCookie ? sortOrderNameCookie : undefined;"

The option will always be "undefined" in the first table load. You can't pre-set a column to be sorted in "start-up".

By change this to the default sort options:
"this.options.sortName = sortOrderNameCookie ? sortOrderNameCookie : this.options.sortName"

This will take the defined column or choose "undefined".

Best regards
KlapZaZa


wenzhixin:
Please commit to develop branch, thanks!

wenzhixin:
Merge in #1531, thanks!

PR #1593, opened 2015-10-23
Feature/keep subtable open on initBody

saidketchman:
Currently, when initBody is called, all subtables shown via detailView are closed. This change makes it so the subtables continue to stay opened when the table is rerendered.

In general I think a better solution would be to allow for individual rows to be updated without re-rendering the entire table, but that seems to be much bigger in scope.

I'm open to suggestions on how to improve this code. I currently have it working in a private instance of bootstrap-table and it's working great. I originally just hacked this together but I've tried refining it a bit to make it suitable for merging, but I'm not claiming it's the golden standard yet :)


wenzhixin:
Can you provide a jsFiddle example?

sketchman-nyt:
Sure, here you go: https://jsfiddle.net/kgL1rocs/1/

1) Open a subtable
2) Click the Update Row button
3) Subtable should stay open

All subtables should always stay open now. Old behavior would cause subtable to close.

src/bootstrap-table.js
```diff
@@ -1558,6 +1560,47 @@
...
+
+            if (detailViewColPos > -1) {
+
+                var existingRowIndex;           //Current item in the existing table
+                var $existingRow;               //jquery obj of the existing table item row
+                var $existingRowDetailView;    //jquery obj of the existing table item row's detailview
+                var $detailRow;                 //the new detail row we'll append to html
+
+                //if the detailview was open previously, keep it open
+                existingRowIndex = $.inArray(that.getRowByUniqueId(item[that.options.uniqueId]), that.options.data);
```
wenzhixin: In many cases, do not set the `uniqueId` option, it will cause an bug: https://jsfiddle.net/wenyi/kgL1rocs/2/.

sketchman-nyt: Ah good catch, let me see if I can think of a workaround in that case.

djhvscf:
@sketchman-nyt any update here?

sketchman-nyt:
Sorry been a while, I'll see about taking another look at this soon.

wenzhixin:
Close this PR because there is no any response, thanks for your contribution!

PR #1595, opened 2015-10-24
Cookie extension does not use default sort settings

goodkiller:
Cookie extension does not use sortOrder and sortName if it's defined in init.


wenzhixin:
Merged in #1531, thanks.

PR #1611, opened 2015-10-29
adds a numericOnly sort option

duanemay:
We could expand this extension to offer other types of sort. I didn't rename it for backward compatibility.


wenzhixin:
Please commit to develop branch, thanks!

PR #1615, opened 2015-10-30
adds a numericOnly sort option

duanemay:
This change adds a numericOnly sort option. I added it to the existing natural-sorting package, since they both deal with sorting of the columns. We could expand this extension to offer other types of sort. I didn't rename it for backward compatibility.

### DIFF
--- a/src/extensions/natural-sorting/bootstrap-table-natural-sorting.js
+++ b/src/extensions/natural-sorting/bootstrap-table-natural-sorting.js
@@ -2,10 +2,11 @@
  * @author: Brian Huisman
  * @webSite: http://www.greywyvern.com
  * @version: v1.0.0
- * JS function to allow natural sorting on bootstrap-table columns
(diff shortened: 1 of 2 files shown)


src/extensions/natural-sorting/bootstrap-table-natural-sorting.js
```diff
@@ -44,4 +45,16 @@ function alphanum(a, b) {
     }
   }
   return aa.length - bb.length;
+}
+
+function numericOnly(a, b) {
+    function stripNonNumber(s) {
+        s = s.replace(new RegExp(/[^0-9]/g), "");
+        return parseInt(s.toLowerCase());
```
wenzhixin: Use parseInt(s, 10) instead

duanemay: good catch, updated

wenzhixin:
Thanks!

PR #1620, opened 2015-10-30
Add `title` attribute to pagination controls defining the page number the control option will take you to

u01jmg3:
Add `title` attribute to pagination controls defining the page number the control option will take you to

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1287,10 +1287,16 @@
             if (this.options.pageNumber <= 1) {
                 $first.addClass('disabled');
                 $pre.addClass('disabled');
+            } else {
+                $first.attr('title', 1);
+                $pre.attr('title', (this.options.pageNumber - 1));
             }
             if (this.options.pageNumber >= this.totalPages) {
                 $next.addClass('disabled');
                 $last.addClass('disabled');
+            } else {
+                $next.attr('title', (this.options.pageNumber + 1));
+                $last.attr('title', this.totalPages);
             }
             if (this.options.smartDisplay) {
                 if (this.totalPages <= 1) {


dabros:
Though it is simple in that it just shows page numbers, its not entirely clear whats going on or looks nice at all...

@u01jmg3 - this fiddle points to your commit here, as pointing to your 'master' did not show that commit

https://jsfiddle.net/dabros/5a5umzvc/

I dont mind the idea of seeing the total number of pages, though as explained in the linked github issue, the location string at bottom left seems more than sufficient.

If there was a pull request regarding showing total page number, then i strongly suggest instead using the standard method:

1 2 3 ... 7

Shouldnt be a hard change to make, if @u01jmg3 wants to quickly mock that up i think it likely it would work quite well : )

u01jmg3:
> Though it is simple in that it just shows page numbers, its not entirely clear whats going on or looks nice at all...

- What part isn't clear? It's a title tooltip just like those used on other buttons in Bootstrap Table. There is no look to it as that is governed by the browser. Styling is inconsequential. It serves an accessibility purpose.

> This fiddle points to your commit here, as pointing to your 'master' did not show that commit
https://jsfiddle.net/dabros/5a5umzvc/

- As per the contributing guidelines, it states I should commit to the develop branch which I have done. I have therefore made no changes to my version of 'master'. The fiddle, for me, does not contain my commit.

> I dont mind the idea of seeing the total number of pages, though as explained in the linked github issue, the location string at bottom left seems more than sufficient.

- It isn't sufficient. How can I fit the total into the text "Showing 1 to 10 of x rows"?
   - "Showing 1 to 10 of x rows of y pages"? That does not read well plus you want the information when you're using the pagination controls, not reading the text at the opposite side of the page.

> If there was a pull request regarding showing total page number, then i strongly suggest instead using the standard method:
1 2 3 ... 7
Shouldnt be a hard change to make, if u01jmg3 wants to quickly mock that up i think it likely it would work quite well : )

- I will see what I can do but suddenly a simple change is escalating into more based on the preference of one person. Are others happy for the pagination controls to be changed in this way (1 2 3 ... 7)?

u01jmg3:
Is this what you were looking for?

demo

dabros:
@u01jmg3  - i did edit my post a few times in first few min, i get feeling if fiddle not showing your commit that you saw it then - i fixed that very quickly.

And as i said, its just my 2 cents and if author wants to pull simple version than go for it.

Its not my preferance for '..' format, but rather consistent user experience as that is the common form - ive never seen titles on page anchors before nor even heard of that idea.

Regarding 'not sufficient', again IMHO but "Showing X of Y rows, M per page" - that really more than covers the total page count as you know exactly how many records, which subset you are viewing, how to expect each page and therefore simple math to guess how many more records and/or pages to go.

And yes, that gif shows what i mean, looks great to me. 

Just using F12 to edit anchor contents shows different widths with padding, cant see it being more or less an issue than without '...' though, as that doesnt change the offset amount in any way from without '...'

u01jmg3:
> i did edit my post a few times in first few min, i get feeling if fiddle not showing your commit that you saw it then - i fixed that very quickly but prob should have put EDIT note up, sorry.

No worries, the Fiddle was more for your benefit than mine.

> And as i said, its just my 2 cents and if author wants to pull simple version then go for it.

@wenzhixin: which solution do you prefer to letting the user know the total number of pages?

> However, its not just my preferance for '..' format, but rather consistent user experience as that is the common practice - ive never seen titles on page anchors before nor even heard of that before. Its not 'clear' as its just a number showing up over a button - but that doesnt happen elsewhere so you have to pause and think, even if you notice it.

Agree that the '1 2 3 ... 7' format for pagination controls is more common.

> I work with many non-savy users and they almost never even think to hover over anything for tooltips, and i cant see them knowing or thinking to look there for it. A localisation string option could allow simple 'X pages' or similar format, but still doesnt make users more inclined to seek it out by hovering there.

Tooltips are very handy and user testing would show you this. For example, a title attribute is essential when using font icons as button text. Not everyone associates a particular icon with a function so adding a title attribute means they can hover over a button they are not sure about and get a tooltip detailing what the button does. Or indeed adding the page number, as a tooltip, to a last page button to indicate where the user will be taken.

title

> Regarding 'not sufficient', again IMHO but "Showing X of Y rows, M per page" - that really more than covers the total page count as you know exactly how many records, which subset you are viewing, how many to expect each page and therefore simple math to guess how many more records and/or pages to go. I just cant see a use case when that isnt enough.

Users do not want to have to calculate in their head the total number of pages or 'guess'!

> And yes, that gif shows what i mean, looks great to me.

Thanks

> Just using F12 to edit anchor contents shows different widths with padding, cant see it being more or less an issue than without '...' though, as that doesnt change the offset amount in any way from without '...'

- When page numbers are in double digits going from 9 to 10 then 11, etc. can cause jumping - e.g. the user could have their mouse over the previous or next page button but the jump has caused their mouse to no longer be hovered over it.
  - Simple logic to counter this could be to check if `total pages > 9` then add padding to page number buttons `<= 9`.
  - The width of the `...` button is more or less in between the width of a `9` and `10` button but similar logic could be applied.

wenzhixin:
Thanks @u01jmg3 and @dabros, I will consider to use the [tooltips](http://getbootstrap.com/javascript/#tooltips) feature.

u01jmg3:
@wenzhixin - are you interested in this new layout for the pagination controls? I have code waiting if you'd like a PR?

demo

wenzhixin:
:+1:

Issue #1496, opened 2015-10-03
inconsistency, rows or records


timoline:
Hello,
maybe it's me but there is some inconsistency.
Is it rows or records?

Showing 1 to 10 of 800 rows 10  records per page

wenzhixin:
rows or records? Which one do you think is more appropriate?

timoline:
hmmm, rows ?

dabros:
yes, rows seems more appropriate

related: http://dba.stackexchange.com/questions/31805/what-is-the-difference-between-a-record-and-a-row-in-sql-server

dabros:
@wenzhixin - if looks good to you, maybe close this issue?

Prob best to scan to make sure i didnt miss any that use both 'rows' and 'records' in same file.

timoline:
In line 436 of bootstrap-table.js, there is still records per page and I see more "records" in this file where it should be "rows"

djhvscf:
@timoline I will open this issue but can you open a new PR with the changes that we should apply ?

PR #1655, opened 2015-11-09
'records' to 'rows' as per #1496

dabros:
#1496

Starting looking at all languages, but too hard to ensure translation accuracy so just searched for those using both 'records' and 'rows' and forced consistency of 'rows' as per discussion


wenzhixin:
#1668

Issue #1649, opened 2015-11-06
Column events break with multiple columns with the same field name


wcatron:
When creating a table with multiple columns with the same field name (example 'fork_count') the events do not apply to the proper column. They try to apply to the first instance of the field name.

Example first table works, second fails.
https://jsfiddle.net/e3nk137y/3554/

This has worked in the past as evident by this example which uses version 1.6
https://jsfiddle.net/sf36oa42/3/

dabros:
@wcatron - fix your post, your first fiddle doesnt show the issue, and the second points to master, so i think you got them mixed up.

Im not ruling this out as a bug or not, but to work just remember you dont even need **data-field** if your using a **formatter**, as the entire row is passed in.

See my update: https://jsfiddle.net/dabros/sf36oa42/1/

wcatron:
I've updated the original JSFiddles I must not have realized how to save them. 

>Im not ruling this out as a bug or not, but to work just remember you dont even need data-field if your using a formatter, as the entire row is passed in.

I understand that *data-field* is not required, I was able to work around the problem. However, formatters are often times (at least in my projects) reused across multiple tables representing slightly different data. operators should have the same reusability.

>Im not sure that leaving out field wont bug other code, but you can give it your own name if that does happen, which i can only expect if needed to referance by column name.

Once you relook at the fiddles again you should be able to see what I was talking about when I said that it worked previously in version 1.6. I had the same thought that it was necessary for referencing the column. However, having worked previously it seems like if they've updated their referencing and broken functionality. This is how I discovered the issue when I updated my old code that worked.

>Internally it uses some calculateObjectValue or similar that takes into account formatters. However, maybe sortName would be enough to cover all other use cases?

You lost me here. I'm not an expert on the internals of bootstrap. I've dabbled in extending some of the search stuff but other than that I was really looking for some help.

Thanks for taking a look at this!

dabros:
@wcatron - just some quick notes

I make reusable formatters too, but its simple to place some kind of `if(typeOf(val) === 'undefined')` with a switch if further handling needed, and/or 'convenience handlers' aka 'alias' for those formatters that call them but pass in some extra param. You can also define formatter via js so can make those aliases into anonymous functions there and just call your reusables with optional param passed in.

**sortName** is an optional column param that might be needed if you ommit the 'data-field' entirely, or if the 'data-field' doesnt actually match up to a sortable field name.

[...]

PR #1663, opened 2015-11-10
Fix duplicate data field issues with events and visibility.

wcatron:
As identified in issue #1649 this fixes issues with setting events for columns where the data-field parameter is set to the same field. Although this field is optional using it should not break expected functionality. Further more from the issue I identified, a similar problem is also seen when setting the visible parameter. It seemed like the way that the column information was being fetched using the `getFieldIndex` method was on purpose as if using that method added functionality. I mostly removed extra functions and used indexes that were already calculated and available.


wcatron:
Format Issues will try again, for some reason this has a lot of commits associated with it.

PR #1664, opened 2015-11-10
Fix data field

wcatron:
As identified in issue #1649 this fixes issues with setting events for columns where the data-field parameter is set to the same field. Although this field is optional using it should not break expected functionality. Further more from the issue I identified, a similar problem is also seen when setting the visible parameter. It seemed like the way that the column information was being fetched using the getFieldIndex method was on purpose as if using that method added functionality. I mostly removed extra functions and used indexes that were already calculated and available.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1460,7 +1460,7 @@
                     data_ = '',
                     rowspan_ = '',
                     title_ = '',
-                    column = that.columns[getFieldIndex(that.columns, field)];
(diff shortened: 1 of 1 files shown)


djhvscf:
:+1:

wcatron:
When I merged my changes from the correct branch it brought over all changes from the development branch as well. PR #1679 is much cleaner.

wenzhixin:
So can we close this, and test your change in #1679?

Issue #1629, opened 2015-11-02
Issue sorting columns containing the same value when browsers have implemented an unstable `Array.prototype.sort(comparefn)`


u01jmg3:
The property of a sorting algorithm that leaves elements that compare as equal in their original list order is called stability. The JavaScript spec specifically allows for implementations to use unstable sort algorithms; no version of Chrome adopts a stable sorting implementation.

- [From the spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.11):

> The sort is not necessarily stable (that is, elements that compare equal do not necessarily remain in their original order).

It is possible to get a stable sorting from a non-stable sort function. Before sorting you get the position of all the elements. In your sort condition, if both elements are equal, then you sort by the position.

---

Below is a modified `initSort()` solving the issue described above - would you be interested in a PR for this change? 

```js
if (!Object.keys) {
    Object.keys = function(obj) {
        var keys = [];

        for (var i in obj) {
            if (obj.hasOwnProperty(i)) {
                keys.push(i);
            }
        }

        return keys;
    };
}

BootstrapTable.prototype.initSort = function () {
    var that = this,
        name = this.options.sortName,
        order = this.options.sortOrder === 'desc' ? -1 : 1,
        index = $.inArray(this.options.sortName, this.header.fields),
        length = Object.keys(this.data).length;

    // Add current position to data object in order to enable a stable sort
    // for times when values are compared as equal
    for (var key in this.data) {
        if (this.data.hasOwnProperty(key)) {
            // Save position
            this.data[key].position = key;
        }
    }

    if (index !== -1) {
        this.data.sort(function (a, b) {
            if (that.header.sortNames[index]) {
                name = that.header.sortNames[index];
            }
            var aa = getItemField(a, name),
                bb = getItemField(b, name),
                value = calculateObjectValue(that.header, that.header.sorters[index], [aa, bb]);

            if (value !== undefined) {
                return order * value;
            }

            // Fix #161: undefined or null string sort bug.
            if (aa === undefined || aa === null) {
                aa = '';
            }
            if (bb === undefined || bb === null) {
                bb = '';
            }

            // Found to be equal
            if (aa === bb) {
                    var aa_position = getItemField(a, 'position'),
                        bb_position = getItemField(b, 'position');

                if (order == 1) {
                    aa = (length - aa_position),
                    bb = (length - bb_position);
                } else {
                    aa = aa_position,
                    bb = bb_position;
                }
            }

[...]

PR #1678, opened 2015-11-13
Enable stable sort - when elements are equal sort by their position

u01jmg3:
- See issue #1629
  - Example of problem: https://jsfiddle.net/u01jmg3/hffu9h72/
    - Sort on Item Name column
  - Example of fix: https://jsfiddle.net/u01jmg3/hffu9h72/1/
    - Sort on Item Name column

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -854,8 +854,6 @@
                 }
                 var aa = getItemField(a, name),
                     bb = getItemField(b, name),
-                    aa_position = getItemField(a, 'position'),
-                    bb_position = getItemField(b, 'position'),
                     value = calculateObjectValue(that.header, that.header.sorters[index], [aa, bb]);
 
                 if (value !== undefined) {
@@ -872,6 +870,9 @@
 
                 // Found to be equal
                 if (aa === bb) {
+                    var aa_position = getItemField(a, 'position'),
+                        bb_position = getItemField(b, 'position');
+
                     if (order == 1) {
                         aa = (length - aa_position),
                         bb = (length - bb_position);
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -837,7 +837,17 @@
         var that = this,
             name = this.options.sortName,
             order = this.options.sortOrder === 'desc' ? -1 : 1,
-            index = $.inArray(this.options.sortName, this.header.fields);
+            index = $.inArray(this.options.sortName, this.header.fields),
+            length = Object.keys(this.data).length;
+
+        // Add current position to data object in order to enable a stable sort
+        // for times when values are compared as equal
+        for (var key in this.data) {
+            if (this.data.hasOwnProperty(key)) {
+                // Save position
+                this.data[key].position = key;
+            }
+        }
 
         if (index !== -1) {
             this.data.sort(function (a, b) {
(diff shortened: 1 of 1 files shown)


wenzhixin:
Have you tested the code in IE browser?

src/bootstrap-table.js
```diff
@@ -837,7 +837,17 @@
         var that = this,
             name = this.options.sortName,
             order = this.options.sortOrder === 'desc' ? -1 : 1,
-            index = $.inArray(this.options.sortName, this.header.fields);
+            index = $.inArray(this.options.sortName, this.header.fields),
+            length = Object.keys(this.data).length;
```
wenzhixin: IE do not support this method.

src/bootstrap-table.js
```diff
@@ -846,6 +856,8 @@
                 }
                 var aa = getItemField(a, name),
                     bb = getItemField(b, name),
+                    aa_position = getItemField(a, 'position'),
+                    bb_position = getItemField(b, 'position'),
```
wenzhixin: I don't know why here use position field, If haven't use this field? Please provide a working jsfiddle with your changed code, thanks.

src/bootstrap-table.js
u01jmg3: See 4b4ede7 for IE8 fix

src/bootstrap-table.js
u01jmg3: Fixed with b6bcb8b

wenzhixin:
in this case, it is a bug https://jsfiddle.net/hffu9h72/2/

u01jmg3:
Anything else I need to change for you to merge?

wenzhixin:
@u01jmg3, my suggestion is to add this feature to an extension like: `bootstrap-table-stable-sort.js`, because add a `position` field to every row is not reasonable, your origin data is: `[{name: 'bootstrap-table'}]` and then changed to `[{name: 'bootstrap-table', position: 0}]`, will let user feel confused.

BTW, the type of `this.data` is array, so you can you use:
```
length = this.data.length;
...
$.each(this.data, function (i, row) {
    if (!row.hasOwnProperty('position')) { // or use _postion instead
        row.position = i;
    }
});
```

What do you think?

u01jmg3:
Including a bug fix as an **optional** extension would not be my choice of action but then this is your library.

Including the original position of each data item is the only way of feasibly overcoming this issue and maintaining a stable sorting algorithm.

I do agree it adds marginal bloating to the data array - would `_postion` make more sense to the user? Perhaps I could look at removing this key post-ordering?

u01jmg3:
Still waiting to find out what you would like to do - I'd like to get this PR closed off

djhvscf:
@wenzhixin @u01jmg3 what about if we use the index of the item in the data array provided instead of add a new property to each item?? Because it can creates a new bugs and that is not the idea

wenzhixin:
How to use the index of the item? @djhvscf

wenzhixin:
Added an option: `sortStable`
https://github.com/wenzhixin/bootstrap-table/commit/44cb312df059184ebb12cb338f87990b454973a0
https://jsfiddle.net/wenyi/hffu9h72/3/
Thanks @u01jmg3

fvaletas:
The fix (sortStable  option) does not work for a custom sorter

PR #1679, opened 2015-11-13
Calculate column Index differently for events/toggle. Fix #1649

### LINKED ISSUE #1649: Column events break with multiple columns with the same field name

wcatron:
Second attempt at fixing #1649 by calculating the index for the columns. Kept @wenzhixin comments in mind and the same bugs are not present. I think this is just a starting point. If this doesn't break anything I think there are more places that the index could be calculated based on the given index instead of the key name.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -55,6 +55,18 @@
         return index;
     };
 
+    var getFieldIndexFromColumnIndex = function (columns, fieldIndex) {
+        $.each(columns, function (i, column) {
+          if (!column.visible) {
+            fieldIndex--;
+          }
+          if (i == fieldIndex) {
+            return false;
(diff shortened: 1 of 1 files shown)


wenzhixin:
Can you commit this to the same PR #1664?

wcatron:
I tried but I think I did more damage than good. Sorry.

wenzhixin:
Thanks @wcatron

Issue #1677, opened 2015-11-13
When Server + PaginationSwitch on, it still pass pageNumber and pageSize


alvin0618:
Hi guys,

This problem I faced recently.
```
options:   
sidePagination: 'server',   
showPaginationSwitch: true,   
```
When I click paginationSwitch button, it still passing the params - pageNumber and pageSize to backend end point. 
```
{"pageNumber":1,"pageSize":10,"sortOrder":"asc"}
```

Now, my work around solution is the check the button whether has class `glyphicon glyphicon-collapse-down icon-chevron-down`. In `queryParams` implementation, if has down button then send pageNumber and pageSize, if not overwrite the pageNumber and pageSize to - 1.

I'm not sure this is the best solution, anyone can suggest even better way to cope with this?

dabros:
@alvin0618 - im not seeing that behaivour in http://issues.wenzhixin.net.cn/bootstrap-table/, hit F12 and watch network requests to see what is passed.

Interestingly, the $_GET query is `?order=asc&limit=10&offset=10` or `?order=asc`, which makes me think its due to an extension or `queryParam` or similar extra code on top of the base. They also could be sent via POST or similar, if you changed query method, but default is 'GET'.

Please setup a fiddle that uses whatever extensions and param you have, i suggest using one of the example fiddles from https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#bug-reports (table from url example) and adding whatever extra you need to make it display this behaivour.

alvin0618:
Hi dabros,

I am able to simulate the problem already. If you define `queryParams` and `queryParamsType`, then you will face the problem that I found.

```
                queryParams: function(params) {
                	var pageNumber = params.pageNumber;
                	var pageSize = params.pageSize;
                	if ($('.fixed-table-toolbar button[name="paginationSwitch"] i').hasClass('icon-chevron-up')) {
                		pageNumber = -1;
                		pageSize = -1;
                	}
                	
	 				return {
	 					pageNumber: pageNumber,
	 					pageSize: pageSize,
	 					searchText: params.searchText,
	 		            sortName: params.sortName,
	 		            sortOrder: params.sortOrder
	 		        };
	 			},
	 			queryParamsType: '',
```

dabros:
Here is what i needed to emulate this

http://jsfiddle.net/dabros/bosqoo29/

Will try to look into it later today, but no promises as pretty busy atm

[...]

PR #1680, opened 2015-11-13
fixed paginationSwitch for server-side #1677

dabros:
as see in #1677, server-side didnt check against this.options.pagination, so pagination-switch became ignored unless crude queryParam workaround. 

tested via http://jsfiddle.net/dabros/bosqoo29/4/ as no environment atm and couldnt get it to link in jsfiddle, though i done that in past, but its a friday ;)

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1749,10 +1749,9 @@
             },
             request;
 
-        if(this.options.pagination)
-        {
-        	params.pageSize = this.options.pageSize === this.options.formatAllRows() ?
-                    this.options.totalRows : this.options.pageSize;
+        if(this.options.pagination) {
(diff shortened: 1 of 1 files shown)


src/bootstrap-table.js
```diff
@@ -1742,16 +1742,20 @@
...
-                    this.options.totalRows : this.options.pageSize,
-                pageNumber: this.options.pageNumber,
+            params = {                
                 searchText: this.searchText,
                 sortName: this.options.sortName,
                 sortOrder: this.options.sortOrder
             },
             request;
 
+        if(this.options.pagination)
```
wenzhixin: please update to `If (...) {` and update tab to 4 spaces, thanks :)

dabros: @wenzhixin - strange tab was there if just using online editor, but redid indents again and moved '{' onto same line

Issue #1613, opened 2015-10-29
Misaligned colums in footer


LoZio:
I have a simple table that needs a total of one column. I enable table footer and define the formatting function that sums it all and delivers it back. It all works fine but the columns in the footer row are not aligned with the ones in the table. Inspecting the table (as in the pic) seems that the widths are calculated outside of the table itself up to the containing page. I have no tags for defining column or table widths, all is default except of course data-show-footer="true". Setting both width to "50%" results in uneven table nonetheless.

![untitled2](https://cloud.githubusercontent.com/assets/869890/10819189/fe087406-7e46-11e5-827f-1ded0d7ebaeb.png)

![untitled](https://cloud.githubusercontent.com/assets/869890/10819012/137ea1bc-7e46-11e5-88fc-bc3a005865a9.png)

The generated HTML
![h](https://cloud.githubusercontent.com/assets/869890/10819256/5f87afb2-7e47-11e5-96b6-332e44452a82.png)

Any idea?

LoZio:
Using FF 41, Chrome 46 and IE11

likeavirgil:
I have the same problem :/

wcatron:
Same here. [http://jsfiddle.net/88bmyvqv/1/](jsfiddle)

wcatron:
Is it possible that this was caused by fixing #639?

LockTar:
PR #1684 seems to fix the bug. The only thing that I found was when you expand the screen, the footer columns width get wrong again. But I think that this is a problem because it is not in the actual table `tfoot` but in a `div` beneath it.
See for example this fiddle with this PR http://jsfiddle.net/hbzctayu/4/ and for without this PR http://jsfiddle.net/hbzctayu/3/.
Try to change the result window in the fiddle to see the strang width again.

wcatron:
So it looks like #1684 fixes half the problem. Hopefully @djhvscf or someone can pull that in. Then relook at resizing, it might be something you have to add on like #639 's original solution.
```
$( window ).resize(function() {
   $('#table').bootstrapTable('resetView');
});
```
This is based off your fiddle with the PR. [jsFiddle](https://jsfiddle.net/hbzctayu/5/)

wenzhixin:
Merged.

qianL93:
if this issue fixed? i have the same problem @ 1.11.0

mtngnn:
This issue still exists I am using vers 1.18.2

PR #1684, opened 2015-11-13
Fix padding in footer.

wcatron:
Fixed padding issues with footer that was making the columns too wide. As addressed in #1613. Concerned about removing the `!important` from line 20. But otherwise the problem still persisted and was overwriting the solution.


wcatron:
How can I get the Travis CI build to re-run? In the details it shows it could resolve host github.com so apparently the test server was having problem.

wcatron:
Closed and reopened to ru run the Travis CI build which failed due to connection issue, not my code.

PR #1691, opened 2015-11-15
Added extension.json files

djhvscf:


wenzhixin:
Updated: http://bootstrap-table.wenzhixin.net.cn/extensions/

PR #1720, opened 2015-11-21
Added check for null values on existsOptionInSelectControl

SteveyPugs:
.toString() values were crashing

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -41,8 +41,13 @@
     var existsOptionInSelectControl = function (selectControl, value) {
         var options = selectControl.get(selectControl.length - 1).options;
         for (var i = 0; i < options.length; i++) {
-            if (options[i].value === value.toString()) {
-                //The value is nor valid to add
+            if(value){
+                if (options[i].value === value.toString()) {
+                    //The value is nor valid to add
+                    return false;
+                }
+            }
+            else{
                 return false;
             }
         }


src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -41,8 +41,13 @@
     var existsOptionInSelectControl = function (selectControl, value) {
         var options = selectControl.get(selectControl.length - 1).options;
         for (var i = 0; i < options.length; i++) {
-            if (options[i].value === value.toString()) {
```
wenzhixin: Why don't you use:
```
if (!value || options[i].value === value.toString()) {
    return false;
}
```
instead of:
```
if(value){
    if (options[i].value === value.toString()) {
        //The value is nor valid to add
        return false;
    }
}
else{
    return false;
}
```

wenzhixin:
Please provide a jsFiddle to show the problem you said, thanks!

dabros:
@SteveyPugs - please see comments above and provide fiddle to prove changes and original problem 

https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md

thanks

dabros:
@SteveyPugs - still waiting

djhvscf:
@dabros Can you create a new PR with this change? and then I will close this issue because @SteveyPugs did not answer anything

djhvscf:
I will just wait for your confirmation in order to close this issue @dabros

SteveyPugs:
Sorry for the ridiculously late response.

This appears to be fixed.

http://jsfiddle.net/steveypugs/cg664553/1/

I made a fiddle for the case but the [CDN](https://cdnjs.com/libraries/bootstrap-table) I was using is using v1.10.x, where as the [repo](https://github.com/wenzhixin/bootstrap-table/tree/master/src/extensions/filter-control) shows v1.0.x

Am I missing something in finding the latest copy?

SteveyPugs:
I have found the updated version. Was looking in the incorrect folder.

https://github.com/wenzhixin/bootstrap-table/tree/master/dist/extensions/filter-control

dabros:
@djhvscf - im not following this issue myself, just helping wenzhixin by bumping inactive threads

Whoever is following can look at latest response, if you need opinion ask then :)

SteveyPugs:
If you passed a NULL value and used a select filter control in 1.0.x version it would break the code cause it was trying to do toString() to NULL. I think this is fixed though as my jfiddle proves otherwise. I also noticed the function has been re-written with a different name, so I think we can close this issue.

djhvscf:
Ok... I will close this .See #2027

SteveyPugs:
Welcome @djhvscf! Sorry again for the late response.

SteveyPugs:
Just curious though, why is that version not updated to the latest version?

djhvscf:
which version??

SteveyPugs:
This one:

https://github.com/wenzhixin/bootstrap-table/blob/master/dist/extensions/filter-control/bootstrap-table-filter-control.js

This still says 1.0.0

wenzhixin:
@djhvscf do you think how to manage the extension version?

PR #1727, opened 2015-11-23
Bugfix -- filterControl -- trim values for select fields

ioctaptceb:
fixes issue: #1728 -- full description can be found there

solution: just trim the value before setting it.

added a couple of formatting changes according to the linter as well

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -14,6 +14,7 @@
     });
 
     var addOptionToSelectControl = function (selectControl, value, text) {
+        value = $.trim(value);
         selectControl = $(selectControl.get(selectControl.length - 1));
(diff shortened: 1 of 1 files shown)


djhvscf:
Great! Thanks @ioctaptceb

djhvscf:
Fix #1728

src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -11,6 +11,7 @@
     var sprintf = $.fn.bootstrapTable.utils.sprintf;
 
     var addOptionToSelectControl = function (selectControl, value, text) {
+        value = value.trim();
```
wenzhixin: Please use `$.trim(value)` instead because IE8 does not support this function.

wenzhixin:
Thanks

Issue #1742, opened 2015-11-26
use the hideRow function must be two parameter


wangy09:
use the hideRow function must be two parameter,index and uniqueId  ,but I just want to use one parameter such as index or uniqueId。if I use one of them,it can not hide the row.Finally I found a bug in the bootstrap-table.js  2269 line,cange ' if (!params.hasOwnProperty('index') || !params.hasOwnProperty('uniqueId'))' to ' if (!params.hasOwnProperty('index') && !params.hasOwnProperty('uniqueId'))' ,success!

dabros:
I can confirm the example at http://issues.wenzhixin.net.cn/bootstrap-table/#methods/showRow-hideRow.html appears to show this bug too, but im sure it didnt in the past

dabros:
@wenzhixin, @djhvscf - i am somewhat concerned that we referance variable not in current version.

Please see my notes on the PR #1744

thanks

wenzhixin:
Merged.

PR #1743, opened 2015-11-26
showRow & hideRow param checks, see #1742

dabros:
As seen in #1742, the param checks now require 'both' param, instead of option of which is most relevent.

Not sure when this snuck in, but easy fix.


wenzhixin:
Thanks @dabros

PR #1744, opened 2015-11-26
hideRow & showRow does NOT use isIdField!!

dabros:
As noted in #1742, doco is in touch with current code.

This feels like it could be an issue, Im making this PR just to make doco in line with code, but it would be best if someone could also trace when this happened.

If there 'was' an `isIdField` param, then when and why did it go away, and are older versions still using it???


dabros:
Noticed this commit, havent had chance to look much at it though yet so sharing here: https://github.com/wenzhixin/bootstrap-table/commit/960cfcc40a0a7983ceef6f6cecd89e90dde5d16e#diff-95c58f988cb3038d8e3b2a5176112424

wenzhixin:
I forgot to update the docs, thanks @dabros.

PR #1745, opened 2015-11-27
respect the searchFormatter option

bsa01:


wenzhixin:
@djhvscf what do you think?

djhvscf:
I think it is ok because in the current onSearch method you have that validation @wenzhixin

wenzhixin:
OK

Issue #1747, opened 2015-11-27
filterControl with cookies enabled does not erase cookies on clearFilterControl


ioctaptceb:
Issue: if you have cookies enabled with filterControl, and you clear the filter search fields with the trash icon, then refresh the page, all of the previous settings will be there, which is not the UX you want to experience if you just deleted those settings.

set search:
<img width="777" alt="set search options" src="https://cloud.githubusercontent.com/assets/2945260/11444088/f8289064-9522-11e5-8844-d0b7bb8bb233.png">

clear search using trash icon:
<img width="776" alt="clear search with trash button" src="https://cloud.githubusercontent.com/assets/2945260/11444117/3359c324-9523-11e5-8a31-e1d00af103e7.png">

reload page, search options are still there except for the very first control
<img width="779" alt="reload page search is reset" src="https://cloud.githubusercontent.com/assets/2945260/11444097/1007a8e6-9523-11e5-89b0-2c24a6a01c1e.png">

Cause:
i can confirm the following, though i am new to the code base and may be missing some important points -- the only field that is reset in the cookies is the very first field, which corresponds to `$(controls[0]).trigger(controls[0].tagName === 'INPUT' ? 'keyup' : 'change'); `

https://github.com/wenzhixin/bootstrap-table/blob/master/src/extensions/filter-control/bootstrap-table-filter-control.js#L431

-- this makes sense, as the cookies update with an `onSearch()` function, that listens for `keyup` events.

ioctaptceb:
So the issue is a bit weirder. clearing the cookies using a separate function and deleting them does not work. After the timeout expires: https://github.com/wenzhixin/bootstrap-table/blob/master/src/extensions/filter-control/bootstrap-table-filter-control.js#L432 -- for a period of time, resetting the cookies does not work. I can delete them manualy, but their values are returned somewhere around the delay + 500ms. this creates a long wait period. 

However, it also doesn't make sense to me why the timeout is there. There are no initBody() or other init functions running that would require the timeout at this moment in the code. the click that is triggered by this timeout is actually what starts the process, not the other way around.

the extra timeout is not a good solution to clear cookies, because in that time the user may navigate away from the page. (this timeout can be quite long). 

to iterate over the controls and click each one will end up resending the initBody and other init functions multiple times and is also not a solution

What I see as the solution is to not use the timeout for the click, and instead use the timeout for a clear cookie function, that searches for and removes *all* bs.table cookies for a given table

[...]

PR #1751, opened 2015-11-27
fix for filterControl clear function

ioctaptceb:
related issue: #1747 

Solution: use the timeout for a clear cookie function, that searches for and removes all bs.table cookies for a given table. more detail in the ticket

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -119,7 +119,7 @@
                     cookie = cookie.split(".").pop();
                 }
 
-                if (cookies.indexOf(cookie) === -1) {
+                if ($.inArray(cookie, cookies) === -1) {
                     cookies.push(cookie);
                 }
             });
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -109,6 +109,24 @@
         }
     };
 
+    var collectBootstrapCookies = function cookiesRegex() {
+        var cookies = [],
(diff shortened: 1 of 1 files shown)


djhvscf:
Can you provide a fiddle with this change?? I mean in order to test it and be sure that this change works perfectly @ioctaptceb

ioctaptceb:
http://jsfiddle.net/ov75xrjc/1/

to test: type '3' in the star column. list will filter
reload the page, list should still be filtered
click the garbage, should not be filtered
reload page, list should not be filtered

ioctaptceb:
also here is a fiddle for the bug:
http://jsfiddle.net/g4jxgc33/2/

to test: type '3' in the star column. list will filter
reload the page, list should still be filtered
click the garbage, should not be filtered
reload page, list should not be filtered but is **still filtered**

ioctaptceb:
this also allows clearing sort order to a default

djhvscf:
Great!! @ioctaptceb  thanks!

src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -109,6 +109,24 @@
...
+        var cookies = [],
+            foundCookies = document.cookie.match(/(?:bs.table.)(\w*)/g);
+
+        if (foundCookies) {
+            $.each(foundCookies, function (i, cookie) {
+                if (/./.test(cookie)) {
+                    cookie = cookie.split(".").pop();
+                }
+
+                if (cookies.indexOf(cookie) === -1) {
```
wenzhixin: IE8 doesn't support this, use inArray instead.

ioctaptceb: yep, sorry for the delay

Issue #1749, opened 2015-11-27
filterControl + searchbox + cookies = bug


ioctaptceb:
issue: when inputing a filter into a filterControl input box, then refreshing the page, the search box is populated with the same info as the last entered filterControl input. Sometimes this breakes the filtering, either giving too many results or too few.

cause: filterControl uses onSearch the same way that the search input box uses it. 
https://github.com/wenzhixin/bootstrap-table/blob/master/src/extensions/cookie/bootstrap-table-cookie.js#L321
the cookie is reset every time this function is run.

djhvscf:
I'm not sure that I'm understanding this issue.. Provide a fiddle in order to see the bug. @ioctaptceb

ioctaptceb:
you can see the issue here:
http://jsfiddle.net/fnvk8xwa/3/

to see the issue, in the second column titled 'stars' type in the following string: 'boot'
now, you should see that the table has no elements in the star column with the text 'boot' since the column has only numbers

now reload the page
you will see that boot is searched for in all columns

wenzhixin:
Merged

PR #1752, opened 2015-11-27
fix searchtext cookie set by column filters

ioctaptceb:
related Issue #1749 

fix: only save cookie if the onSearch function comes from the search text field. 

bug fiddle: http://jsfiddle.net/fnvk8xwa/3/
fix fiddle: http://jsfiddle.net/y32egy35/4/

steps to reproduce

> to see the issue, in the second column titled 'stars' type in the following string: 'boot'
> now, you should see that the table has no elements in the star column with the text 'boot' since the column has only numbers

> now reload the page
> you will see that boot is searched for in all columns and is in the searchtext box

in the fixed version, the last point does not happen, there is no text in the searchtext box and the filter is only in the column filter

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -317,8 +317,12 @@
     };
 
     BootstrapTable.prototype.onSearch = function () {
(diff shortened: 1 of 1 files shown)


djhvscf:
Great solution! @ioctaptceb

wenzhixin:
Thanks

PR #1692, opened 2015-11-15
Columns can now be initialized after receiving remote data

iancamp22:
Hello,

I'm trying to use the data received from an ajax call to define the column header titles. However, modifying this.columns inside responseHandler() results in no data being shown. I suspect this is because initHeaders() is called somewhere between receiving the data and executing responseHandler(). Adding this.initHeaders() and this.initSort() inside the load() function fixes the issue.

PR #1768, opened 2015-12-02
Columns can now be initialized after receiving remote data

iancamp22:
Pasted from PR #1692:

Hello,

I'm trying to use the data received from an ajax call to define the column header titles. However, modifying this.columns inside responseHandler() results in no data being shown. I suspect this is because initHeaders() is called somewhere between receiving the data and executing responseHandler(). Adding this.initHeaders() and this.initSort() inside the load() function fixes the issue.

Bug JSFiddle: https://jsfiddle.net/1ehuLmw0/ 
Working JSFiddle (using my branch): https://jsfiddle.net/dzd4ts93/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2110,9 +2110,11 @@
             data = data.data;
         }
 
+        this.initHeader();
         this.initData(data);
         this.initSearch();
         this.initPagination();
+        this.initSort();
         this.initBody(fixedScroll);
     };
 


wenzhixin:
Thanks @iancamp22, but I think it is not a good idea to call `initHeader` every time when we use `load` function, your case is very special, we can use a workaround do solve your problem: https://jsfiddle.net/1ehuLmw0/1/. What do you think?

iancamp22:
Okay, I understand. I thought I tried that workaround, but perhaps I wasn't calling `initHeader` the correct way. It seems to work fine now. Thanks!

PR #1781, opened 2015-12-07
getItemField check null

pengzhang:
getItemField  json value check null


wenzhixin:
Thanks! But this bug has been solved in develop branch: https://github.com/wenzhixin/bootstrap-table/commit/b0f0354e61d3be7a3528d18be187914b3aa209d7
Issue: https://github.com/wenzhixin/bootstrap-table/issues/1757

PR #1797, opened 2015-12-11
add iconSizeClass

dreamerworks:
IconSize is not changing according to tableGrid IconSize option. So, I added an extra option "exportIconSize" that allow to define an icon size.
It would be much better, if we managed to inherit the main table iconSize option and have the possibility to override it in the plugin options. Instead of defining an extra option.

### DIFF
--- a/src/extensions/export/bootstrap-table-export.js
+++ b/src/extensions/export/bootstrap-table-export.js
@@ -44,11 +44,14 @@
             var that = this,
                 $btnGroup = this.$toolbar.find('>.btn-group'),
                 $export = $btnGroup.find('div.export');
-
             if (!$export.length) {
+                var iconSizeClass= '';
+                if (this.options.exportIconSize) {
+                    iconSizeClass = ' btn-'+ this.options.exportIconSize; 
+                }
                 $export = $([
                     '<div class="export btn-group">',
-                        '<button class="btn btn-default dropdown-toggle" ' +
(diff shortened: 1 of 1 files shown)


dabros:
@dreamerworks - i played around with a fiddle and proved that using `this.options.iconSize` there should inherit not just the `BootstrapTable.DEFAULTS` value, but the user defined value, if present.

In my test though, iconSize didnt show up at all unless passed, but that maybe old bootstrap-table.js version.

Proof this yourself, then please remove `exportIconSize` option.

If you wanted, you could replace it with more generic extensions of the base icon options, and check for overrides else use base, but that seems overcomplicated as there is no use case for using different icon sets/styles for standard vs export.

dreamerworks:
Hi, 
I have done as you requested.
http://jsfiddle.net/dreamerworks/e3nk137y/4032/
This jsFiddle proves that there is an issue with button icon size and shows a way to fix this. This is a better way than the one I previously proposed.

I kindly ask you to review this once again.
Thanks

dabros:
@wenzhixin - looks alright to me, but pls do confirm as pretty tired atm

wenzhixin:
Needn't to add new option: https://github.com/wenzhixin/bootstrap-table/commit/48a40761751c59118a671de8be68b121f4187a2c, jsFiddle: http://jsfiddle.net/e3nk137y/4225/, thanks anyway.

PR #1798, opened 2015-12-11
Server side paginations uses this.options.searchText, fixes #617

### LINKED ISSUE #617: server-side pagination not displaying pages or row count ; search not working ; sort not working
Hi.  I'm new here.  I just switched my pagination over to server-side, with the default parameters, as indicated on the example page, and a bunch of javascript related things aren't working.  

- My table of 32 rows displays all the rows in one table, with no pages.  It also now returns at the bottom

'Showing 1 to 0 of 0 rows'  , with no page controls on the right-hand side.  

My table is set set up like

    <table class='sastable' id='table' data-toggle='table' data-classes='table table-condensed table-bordered' data-sort-name="{{'plate' if stage=='3d' else 'mjd'}}" data-sort-order='asc'
		data-show-columns='true' data-pagination='true' data-search='true' pageSize='25' 
		data-select-item-name='input' data-toolbar='#toolbar' data-id-field='id' 
		data-side-pagination='server' data-page-list="[10, 25, 50, 100]" name='sastable'>
     </table>

Am I missing something in the setup to activate it?  I could not find another issue related to this one.

- The column sorting no longer works anymore.  

 - The search is also not working.  I updated my bootstrap-table-flatJSON.js as indicated in this thread
https://github.com/wenzhixin/bootstrap-table/issues/434, but this did not fix the problem.  

Am I missing some javascript somewhere to get the server-side pagination set up properly?  

Thanks for your help.

glorang:


djhvscf:
We use a "private" variable in order to have the original values (in the options object) clean.. Maybe the change would be that when we are using server side pagination use the this.searchText

wenzhixin:
Thanks, merged in #1890

PR #1796, opened 2015-12-11
Fixe #1795

yenyen:
Given I have sort a table
When I append, prepend or update rows
Then the sort is not applied

This pull request fixes this issue by calling "initSort" just before "initBody"

PR #1811, opened 2015-12-17
Refreshsort after append rows (same as #1796 but in develop branch)

yenyen:
Same as #1796 but in the develop branch


wenzhixin:
Thanks!

PR #1444, opened 2015-09-22
Fixed ReorderRows init when reorderable is false

robpvn:
When the ReorderabledRows property is false or not present, the rest of
bootstraptable was not inited properly. (Since commit 6201df0bb)

PR #1815, opened 2015-12-18
Fixed ReorderRows init when reorderable is false

robpvn:
When the ReorderabledRows property is false or not present, the rest of
bootstraptable was not inited properly. (Since commit 6201df0)

This is a repeat of PR #1444. Sorry for leaving a driveby PR and then forgetting about it for so long.

@tm1000 provided the fiddles to show that it works:
Fiddle that proves it's broken: http://jsfiddle.net/5gobzuff/3/
Fiddle that Proves this is the proper fix: http://jsfiddle.net/5gobzuff/4/


tm1000:
+1

wenzhixin:
Thanks!

PR #1816, opened 2015-12-18
 Check length when comparing in refreshOptions

robpvn:
This is a proper rehash of PR #1445, to the right branch.

Here's the fiddle that shows why it's a problem: https://jsfiddle.net/5rzajcgc/4/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2579,7 +2579,7 @@
 
     BootstrapTable.prototype.refreshOptions = function (options) {
         //If the objects are equivalent then avoid the call of destroy / init methods
-        if (compareObjects(this.options, options, false)) {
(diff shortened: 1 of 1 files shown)


djhvscf:
Thanks @robpvn , but what do you think about that function?? I mean the logic is ok, can it be improved?

robpvn:
Well, the method is supposed to see if two objects are equal or not, but if you tell it to not compare lengths then it will miss changes where b has more options than a. So it doesn't actually always give you the right result if you pass false on the third option. 

Also, checking the length first saves a lot of time if the objects have a different size, because then you don't have to go through all the properties and compare them, you already know that the objects are different. In general I can't see any case where it would be helpful to not compare lengths.

wenzhixin:
Thanks!

Issue #1832, opened 2015-12-24
is there a way to control editable options with data-attributes ?


promogest:
how can i use something like : 
```<th data-field="valore_unitario"
      data-width="3%"
      data-editable="true"
      data-editable-mode="inline"
      data-sortable="true">```

thanks

fsauter:
Hey, I was in need of the same and added this functionality in my fork. I may open a pull request. But I'm sure that my code needs improvement.

https://github.com/fsauter/bootstrap-table
https://github.com/fsauter/bootstrap-table/commit/638cebc72296e9be226d62e79d88565da68badfb

It enables you to add table and specific column data attributes. All editable data attributes must be prefixed with "editable" (see below). Note that table wide options are used for every column but can be overriden.

````html
<table id="dispatch-type-list"
  data-url="data/url.json"
  data-id-field="id"
  data-editable-emptytext="This field is empty"
  data-editable-url="/my/editable/update/path">
  <thead>
    <tr>
      <th class="col-md-1" data-field="id" data-sortable="true" data-align="center">#</th>
      <th class="col-md-3" data-field="name" data-editable="true">Name</th>
      <th class="col-md-7" data-field="description" data-editable="true" data-editable-emptytext="This will override the empty text option of the table.">Description</th>
      <th class="col-md-1" data-align="center" data-formatter="formatter.actions">Actions
      </th>
    </tr>
  </thead>
</table>
````

promogest:
thanks , it works

dabros:
@promogest - please update topic title to include '[solved]' and close issue please

promogest:
it is solved only because @fsauter did a patch and commit a pull request.

PR #1835, opened 2015-12-24
Enabled data attributes for editable column.

fsauter:
I need to set editable options via data attributes. Since #1832 shows that I'm not alone I decided to open a pull request. I'm really sure that my code needs improvement but it solves the issue:

https://jsfiddle.net/fsauter/6da007fm/

It enables you to add table and specific column data attributes. All editable data attributes must be prefixed with "editable" (see below). Note that table wide options are used for every column but can be overriden.

````html
<table id="dispatch-type-list"
  data-url="data/url.json"
  data-id-field="id"
  data-editable-emptytext="This field is empty"
  data-editable-url="/my/editable/update/path">
  <thead>
    <tr>
      <th class="col-md-1" data-field="id" data-sortable="true" data-align="center">#</th>
      <th class="col-md-3" data-field="name" data-editable="true">Name</th>
      <th class="col-md-7" data-field="description" data-editable="true" data-editable-emptytext="This will override the empty text option of the table.">Description</th>
      <th class="col-md-1" data-align="center" data-formatter="formatter.actions">Actions</th>
    </tr>
  </thead>
</table>
````

So it would be nice to see, that something like this will find its way into the editable extension. Can you help me to improve and integrate it?

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -47,14 +47,34 @@
                 return;
             }
 
+            var editableOptions = {}, editableDataMarkup = [], editableDataPrefix = 'editable-';
+
+            var processDataOptions = function(key, value) {
+              // Replace camel case with dashes.
+              var dashKey = key.replace(/([A-Z])/g, function($1){return "-"+$1.toLowerCase();});
+              if (dashKey.slice(0, editableDataPrefix.length) == editableDataPrefix) {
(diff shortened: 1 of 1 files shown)


promogest:
thanks

wenzhixin:
Thanks! Can you help to update the README file: https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/editable/bootstrap-table-editable.js?

fsauter:
Sure, I will do this within the next two weeks.

PR #1845, opened 2015-12-31
Sort formatted number, eg: 45,646,206

javalover123:


dabros:
@javalover123 - to make easier, pls describe use case where this failed and why this needed, helps wenzhixin if long days ;)

javalover123:
Sort formatted number 45,646,206 and 145,646,206, the original result is 145,646,206 first, but 145,646,206 is larger.

The cause is treat formatted number as string, not number.

dabros:
@javalover123 - so you use regex every time to remove commas, then check if the reformatted string $.isNumeric

Up to @wenzhixin to decide whether this approach has no real performance factors and should be treated as core requirement, but i personally think that this type of formatting is an outside use case and exactly why the `sorter` column option exits, which always custom sort function.

Maybe if you could provide a more widely applicable column-option that handled more than one specific use case, or if this use case was more common, then it would be a **PR**

But as I said, just my opinion, not up to me - and if no real performance impact then maybe - but part of me also thinks that something like "11, 454, 43" is not ALWAYS a formatted number, but can be a sequence of numbers - so I really do feel that this is a `sorter` use case, not non-optional core requirement.

javalover123:
I just replace , to empty string, exclude other word like space, so 11, 454, 43 can't compare as number with my code.

Otherwise, you want treat 11, 454, 43 as string, you can use <span>11, 454, 43</span>.

dabros:
@javalover123 - my points still stand

This is a deployed product, meaning that updates should enhance capabilities and fix bugs, not force users to change working code to acheive same desired outcome they already achieved.

'maybe', oh so 'maybe' as an optional param, but never would this qualify as non-optional core code if 'potentially' ALL users have to then go back to successfully working deployments and use formatters to achieve the same working outcome, due to ONE user not wanting to use a formatter

Its just not a common enough use case, so just use a formatter, like the X number of people before you with similar issues have done, to avoid breaking anyone elses implementation.

wenzhixin:
I agree with @dabros, in your case you can use `sort` to custom your sorter, thanks anyway.

PR #1847, opened 2015-12-31
bugfix: nested object null make the script fail

xm-yang:
when nested object is null, the script will fail.


wenzhixin:
Fixed in develop branch, thanks!

PR #1860, opened 2016-01-06
Rendering detail view directly with detailFormatter

ludup:
Adds a third parameter to detailFormatter method passing the jQuery element of the new cell that will contain the detail view. This allows a developer to render directly into the cell which can be advantageous in some circumstances. If the method returns with the element still empty the string returned from detailFormatter will be appended, maintaining backwards compatibility with previous versions.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1667,10 +1667,13 @@
                 that.trigger('collapse-row', index, row);
             } else {
                 $this.find('i').attr('class', sprintf('%s %s', that.options.iconsPrefix, that.options.icons.detailClose));
-                $tr.after(sprintf('<tr class="detail-view"><td colspan="%s">%s</td></tr>',
-                    $tr.find('td').length, calculateObjectValue(that.options,
-                        that.options.detailFormatter, [index, row], '')));
-                that.trigger('expand-row', index, row, $tr.next().find('td'));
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1667,10 +1667,13 @@
...
             } else {
                 $this.find('i').attr('class', sprintf('%s %s', that.options.iconsPrefix, that.options.icons.detailClose));
-                $tr.after(sprintf('<tr class="detail-view"><td colspan="%s">%s</td></tr>',
-                    $tr.find('td').length, calculateObjectValue(that.options,
-                        that.options.detailFormatter, [index, row], '')));
-                that.trigger('expand-row', index, row, $tr.next().find('td'));
+                $tr.after(sprintf('<tr class="detail-view"><td colspan="%s"></td></tr>', $tr.find('td').length));
+                var $element = $tr.next().find('td');
+                var content = calculateObjectValue(that.options, that.options.detailFormatter, [index, row, $element], '');
+                if($element.length == 1) {
```
wenzhixin: Please update to: `if ($element.length === 1) {`

src/bootstrap-table.js
```diff
@@ -1667,10 +1667,13 @@
...
+                  $element.append(content);
```
wenzhixin: Use `4 spaces` instead.

ludup:
Changes requested have been committed.

wenzhixin:
thanks

PR #1861, opened 2016-01-06
Added example to docs for rowStyle and cellStyle

thornomad:
It wasn't immediately clear what should be returned by the rowStyle and cellStyle functions; this example should help clarify.

### DIFF
--- a/docs/_i18n/en/documentation/column-options.md
+++ b/docs/_i18n/en/documentation/column-options.md
@@ -224,7 +224,15 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
         value: the field value.<br>
         row: the row record data.<br>
         index: the row index.<br>
-        Support classes or css.
+        Support classes or css. Example usage:<br>
+<pre>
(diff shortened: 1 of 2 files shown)


dabros:
@thornomad - thanks for that

Though btw, maybe another quick commit to add code block / pre / whatever to html in `sortName`, to be consistent with above inline html, as well as character "e" outsite table (click 'view whole file')

Just pass-by-observations ;)

thornomad:
@dabros - I can add another commit for `sortName` ... sure.

As for the "e" ... I haven't a clue where that is coming from and think it may be generated from Markdown by some strange twist of fate.  It was there before I got to it.  Tell me how to remove it and I will.

http://bootstrap-table.wenzhixin.net.cn/documentation/#column-options

D

wenzhixin:
Thanks!

wenzhixin:
"e" problem: https://github.com/wenzhixin/bootstrap-table/commit/48c188d0ad2fb28509f0ede1e44113cb8340ee69

Issue #1498, opened 2015-10-03
Prevent re-initialization on search, sort, cell updates, ...


marank:
Consider the following jsFiddle: https://jsfiddle.net/4pthhsp9/3/ (I hope it works this time :))

As you can see, I want to activate the input fields based on the selected rows. Moreover, I want to store the input fields' value in a seperate, hidden column. And on top of that, I'd like to use [jquery.maskedinput](https://github.com/digitalBush/jquery.maskedinput) on the input fields.

Sorting, searching and calling 'updateCell' causes the table to re-init itself, resulting in the input fields getting reset.

Is there any way to prevent the table from re-init itself anytime, but instead (for example) manage the rows as jquery objects?

sketchman-nyt:
:+1: I'm having this same problem with updating my table, it gets completely reinitialized with initBody() being called. This closes any open detail views I have open. It would be nice to have a way to say update in place rather than reinitializing the entire table.

coffy:
+1 for not re-rendering of table after "updateCell" call

coffy:
Hi, @marank and @sketchman-nyt , you can try new enhancement #1883 in development branch just addd <code>{reinit: false}</code> when calling updateCell method. Remember you have to update table cell HTML manually, it won't be redrawn from underlying data.

wenzhixin:
Merged in develop, thanks for @coffy

jcavandoli:
I got the same problem with the `append` method.
If you append a new row, the initBody() will redraw the whole table and reset the custom html in each cells.

Ericzhiluo:
the auto-initialization in 'remove' 'update' 'append'  is very annoying as it resets HTML changes you made. Reinitialize the whole table when only one row is updated is unnecessary.

djhvscf:
We're working in an improvement where we will update just the row(s) updated. This will increase the performance

arpitjain1616:
Is there any update on this?

I am not able to make **reinit: false** work with an Input box in the table and updating the cell value with 'updateCellByUniqueId' method. Waiting for the response. @wenzhixin @aureliendavid @djhvscf @dabros @matusferko

PR #1883, opened 2016-01-14
Added new param for updateCell method to avoid table reinitialization

coffy:
New parameter 'reinit' for updateCell method. Providing <code>{reinit:false}</code> when calling updateCell method,  initSort() and initBody() methods are not executed. New parameter helps to solve issue #1498

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2390,8 +2390,7 @@
         }
         this.data[params.index][params.field] = params.value;
 
-        var reinit = params.reinit;
-        if (params.reinit===false || params.reinit === 'false') {
+        if (params.reinit === false || params.reinit === 'false') {
             return;
         }
         this.initSort();
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2391,7 +2391,9 @@
         this.data[params.index][params.field] = params.value;
 
         var reinit = params.reinit;
-        if (reinit===false || reinit==='false') return;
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -2389,6 +2389,9 @@
             return;
         }
         this.data[params.index][params.field] = params.value;
+
+        var reinit = params.reinit;
+        if (reinit===false || reinit==='false') return;
```
djhvscf: Please format this code.

```javascript
if (!params.reinit || params.reinit === 'false') {
    return;
}
```

coffy: This would break compatibility with current/previous behaviour. Existing **updateCell** calls without this param would not work as before. **reinit** has to have value set. undefined can't pass this condition.
Do you suggest different approach?

src/bootstrap-table.js
```diff
@@ -2389,6 +2389,10 @@
             return;
         }
         this.data[params.index][params.field] = params.value;
+
+        if (params.reinit === false || params.reinit === 'false') {
```
wenzhixin: I think we needn't to check string type: `params.reinit === 'false'`

coffy: agree, will change

martinburch:
Am I calling this option correctly? The update works, but the expanded row still collapses.

```
$myTable.bootstrapTable('updateByUniqueId', {
				id: row['id'],
				row: row,
				reinit: false
			});
```

PR #1895, opened 2016-01-15
Add tooltip for filter-control toolbar button.

SergeyOzeryaniy:
Add tooltip for filter-control toolbar button.
Localize tooltip text for RU.


wenzhixin:
Thanks!

Issue #1902, opened 2016-01-18
Change grunt from a dependency to a devDependency


kohenkatz:
When you run `npm install bootstrap-table`, it also installs `grunt` (and a bunch of `grunt-contrib-*` stuff).  However, since Grunt is a build tool and not a real dependency, it (and its plugins) should be listed in the `devDependencies` options in `package.json` instead of in the main `dependencies`.  That way, those of us who use the files in the `dist` folder and/or use our own build scripts do not need to download Grunt.

Thanks!

wenzhixin:
Any pull request is welcome.

dabros:
@kohenkatz  - **PR** always welcome

https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#pull-requests

PR #1916, opened 2016-01-21
Load Grunt only in development

kohenkatz:
This PR is a fix for #1902.

The benefit of this is that developers of the library still get Grunt when they checkout or download the source and run `npm install`, while end users don't get Grunt when they run `npm install bootstrap-table`.  End users don't need Grunt to be installed by this library because either they already have it on their own or they are not using it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "directories": {
     "doc": "docs"
   },
-  "dependencies": {
+  "devDependencies": {
     "grunt-contrib-copy": "^0.8.0",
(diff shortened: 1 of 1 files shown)


wenzhixin:
Now there are two `devDependencies`, shall we merge them together?

kohenkatz:
Oops. I'll fix that.

kohenkatz:
Fixed

wenzhixin:
Merged, thanks!

PR #1921, opened 2016-01-21
Fixed  key-events on toolbar elements

daciot:
Fixed  key-events, when the  's', 'r', 't', 'p', 'left', and 'right' keys are press on toolbar elements, the key-events it will not run.

### DIFF
--- a/src/extensions/key-events/bootstrap-table-key-events.js
+++ b/src/extensions/key-events/bootstrap-table-key-events.js
@@ -32,7 +32,7 @@
                     $toggle = that.$toolbar.find('button[name="toggle"]'),
                     $paginationSwitch = that.$toolbar.find('button[name="paginationSwitch"]');
 
-                if (document.activeElement === $search.get(0)) {
+                if (document.activeElement === $search.get(0) || !$.contains(document.activeElement ,that.$toolbar.get(0))) {
                     return true;
                 }
 


djhvscf:
Provide an example with this change, please

daciot:
Notice, on first example "NO FIXED", we can't write letter 'R' because, key-event is trigged. On the other hand on second example "FIXED", when we focused in components on toolbar the letter 'R' is write.

video_1453841656

dabros:
@daciot - please link to live code where we can confirm this bug, because im not seeing it anywhere.

Either a fiddle (https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#bug-reports) or a current example (http://issues.wenzhixin.net.cn/bootstrap-table/index.html)

Thanks

daciot:
https://jsfiddle.net/f6huhpdo/

dabros:
@daciot - thanks, so it is specifically when custom toolbar using standard names and client side code

daciot:
No, both sides server and client. The error happend when we use plugin key-events and custom toolbar same time.

dabros:
@daciot - thanks for the clarification and the spot :+1:

wenzhixin:
Thanks @daciot

Issue #1936, opened 2016-01-27
a 'table-no-bordered' with a 'hidden' column show unwanted horizontal scrollbar


gwjwin:
Hi, I am using this (http://issues.wenzhixin.net.cn/bootstrap-table/index.html#issues/463.html) method to search invisible columns, but if I set the table as 'table-no-bordered', the 'No matching records found' shows with unwanted horizontal scrollbar.

Here is the example:
https://jsfiddle.net/gwjwin/gt0Ladad/1/

dabros:
The issue appears to be this rule:

```css
.bootstrap-table .table.table-no-bordered > thead > tr > th, .bootstrap-table .table.table-no-bordered > tbody > tr > td {
    border-right: 2px solid transparent;
}
```

There is no matching rule when not `.table-no-bordered`.

Im guessing that it was made to simulate padding or something that is now no longer present in standard layout.

dabros:
@gwjwin - please see post above

gwjwin:
Thanks for the reply, I am using 'class="hidden"' because I want search it, but 'data-visible' columns works now, and I was confused with the example in this page: 
http://issues.wenzhixin.net.cn/bootstrap-table/index.html#issues/463.html
It says "By default, cannot search invisible column.", and I am sure I cannot search 'data-visible' columns when I posted this issue, but now I can. Is the script updated?

wenzhixin:
@gwjwin You can try this option: `searchable`.

gwjwin:
I agrees with wenzhixin that using 'data-searchable' is more reasonable. So the example page is out-dated. We CAN search invisible columns by default now, and we can disable search by setting 'data-searchable'. 
Issue closed.

dabros:
@gwjwin - thanks for spotting that, doco is the hardest to keep in sync.

So long as this is confirmed, I will make **PR** with doco change

https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#pull-requests

dabros:
@gwjwin - no need for **PR**

The doco is not out of step:

http://bootstrap-table.wenzhixin.net.cn/documentation/#column-options

<table>
[...]

PR #1937, opened 2016-01-27
Update css to fix no-border issue: #1936

dabros:
**EDIT - IMPORTANT**

Just noticed the way he hiding that column in https://jsfiddle.net/gwjwin/gt0Ladad/1/

Instead, use `data-visible` like seen in https://jsfiddle.net/dabros/gt0Ladad/2/, and no issue

@wenzhixin - close this if you happy its his fault for not reading doco.

<br/>
---
**Original post content**
---

As seen in #1936, issues with scroll bars in no-border table

---

The issue appears to be this rule:

```css
.bootstrap-table .table.table-no-bordered > thead > tr > th, .bootstrap-table .table.table-no-bordered > tbody > tr > td {
    border-right: 2px solid transparent;
}
```

There is no matching rule when not `.table-no-bordered`.

If you just remove this (or change to `border-right:none`) the only side effect i can see is when there are results, small border appears between tbody > tr > td, which this 2px border covers (and ofcourse you lose a little bit of padding, though still 8px from td rule).

So to keep that 2px border padding but avoid scroll issue i added this rule:

```css
.bootstrap-table .table.table-no-bordered > tbody > tr > td:last-child {
    border-right: none;
}
```

This commit atm includes both rules, to keep the 2px border for whatever reason it was first created, though you could just change the 2px to `border-right:none` if that approach is preferred.

### DIFF
--- a/src/bootstrap-table.css
+++ b/src/bootstrap-table.css
@@ -25,6 +25,10 @@
     border-right: 2px solid transparent;
 }
 
+.bootstrap-table .table.table-no-bordered > tbody > tr > td:last-child {
+    border-right: none;
+}
+
 .fixed-table-container {
     position: relative;
     clear: both;


djhvscf:
@dabros so I understand that this PR can be closed, right?

dabros:
@djhvscf - if your happy with my conclusion that `data-visible` causes no issues, yes

wenzhixin:
I think it's better to merged this PR.

PR #1938, opened 2016-01-27
Remove bogus conditions that will always be true

ioctaptceb:
Testing if a column has `((!column.radio) || (!column.checkbox))` will
always return true, as even if the column is a radio button or
a check box (or anything else for that matter) this condition will
always be true.


ioctaptceb:
Closed as this is included in #1940

PR #1940, opened 2016-01-27
Refactor filter control select initialization

ioctaptceb:
Extract the conditions in which the Select dropdown for
the input type in the filter control is initialized. Moved
this into its own function (based on the single responsibility
principle) and in general made the code more reader friendly.

The formatted field value will also appear as a formatted field
value (before it was unclear what the relationship between value
and text in addOptionToSelectControl was).

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -385,19 +385,17 @@
 
                 value = $.fn.bootstrapTable.utils.calculateObjectValue(that.header, that.header.formatters[j], [value, item, i], value);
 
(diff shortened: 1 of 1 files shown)


djhvscf:
Great! Thanks for this change! @ioctaptceb but we'd like to see this change working... Please provide a fiddle with it

ioctaptceb:
hiya,

here is a jsfiddle showing that the select functionality still works as expected: http://jsfiddle.net/fu67wkpv/

Question: there is a travisci setup, are there any plans for a test suite? it would make it easier to ensure that features work as expected.

wenzhixin:
Thanks

Issue #1605, opened 2015-10-26
BootstrapTable.prototype.onColumnSearch makes 3 requests


pilb:
Hi,

Using bootstrap-table-filter-control.js, on server side, 3 requests were sent to server.

Thanks.

dabros:
@pilb - more detail please

Ideally include all relevant header and response info from each, or at least url and summary of it so we see what they were

Also please include copy of your html and js related to this plugin

thanks

pilb:
Please find example here : https://jsfiddle.net/gr24ouqo/4/.

Try to enter text into column filter and you will see 3 requests on network. 

```js
BootstrapTable.prototype.onColumnSearch = function (event) {
        copyValues(this);
        var text = $.trim($(event.currentTarget).val());
        var $field = $(event.currentTarget).parent().parent().parent().data('field')

        if ($.isEmptyObject(this.filterColumnsPartial)) {
            this.filterColumnsPartial = {};
        }
        if (text) {
            this.filterColumnsPartial[$field] = text;
        } else {
            delete this.filterColumnsPartial[$field];
        }

        this.options.pageNumber = 1;
        this.onSearch(event); // First Request
        this.updatePagination(); // Second Request
        this.trigger('column-search', $field, text); // Third Request
    };
```
Thanks,

Pilb

dabros:
No time to chase up further atm, but it appears to make as many duplicate requests as there are columns, even if filter attribute not set on some.

dabros:
mayble related: #1501

javieralfaya:
Hi Dennis,
You have closed this issue and related issue 1501, but ¿It has created a patch for this problem?.
Best regards.

djhvscf:
Until now this issue is opened because we don't have any patch for it.. If you have one please open a new PR @javieralfaya

javieralfaya:
Hi Dennis, this is may patch: 

Table filter control when pagination on server submit response twice, i commented this.update Pagination(); into bootstrap-table-filter-control.js and problem is resolved:

BootstrapTable.prototype.onColumnSearch = function (event) {
    copyValues(this);
    var text = $.trim($(event.currentTarget).val());
    var $field = $(event.currentTarget).parent().parent().parent().data('field')

    if ($.isEmptyObject(this.filterColumnsPartial)) {
        this.filterColumnsPartial = {};
    }
    if (text) {
        this.filterColumnsPartial[$field] = text;
    } else {
[...]

PR #1942, opened 2016-01-27
refactor filter cookies extension to avoid dbcalls

ioctaptceb:
There was an incompatibility between the cookies extension and the
filter control extension, which resulted in the following behavior:
on page load, every time that a cookie was found, it would be added to a list of
filter, and it would query the database again. This would result in (for
example) a series of three filters to force three separate database
calls, each database call accumulating the same params from the last.

This update fixes this issue. it also incorporates the change from issue
 #1605 -- in which updatePagination is called by onSearch(), but both
are included in the onColumnSearch() function, causing the table to
double all calls to the database.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1121,6 +1121,7 @@
             this.data = f ? $.grep(this.options.data, function (item, i) {
                 for (var key in f) {
                     if ($.isArray(f[key])) {
+                        // TODO: remove this extra if statement
                         if ($.inArray(item[key], f[key]) === -1) {
                             return false;
                         }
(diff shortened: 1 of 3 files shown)


ioctaptceb:
jsfiddle can be found here: http://jsfiddle.net/cfgmkpp8/

djhvscf:
Thanks @ioctaptceb

ioctaptceb:
@djhvscf can you guys take a look at this?

wenzhixin:

djhvscf:
Sure @wenzhixin :)

wenzhixin:
Thanks

Issue #1941, opened 2016-01-27
Allow to access getVisibleFields or making a method getVisibleColumns


enayet-repo:
I am using bootstrap-table plugin - http://bootstrap-table.wenzhixin.net.cn

There is a method for accessing hidden columns - getHiddenColumns but there is no getVisibleColumns method. Actually I want to search on the visible columns only for this I need the visible column names. 

I think searching is meaning for the visible columns only, otherwise users may be confused with the searched result.

I have searched in the library and found there is a method - getVisibleFields. But it is not accessible. I think it should be allowed to make access.

Thank you...

Stackoverflow link - http://stackoverflow.com/questions/34981291/wenzhixin-bootstrap-table-getvisiblecolumns

dabros:
@enayet-repo - there are ways around that, by using selectors to find hidden columns or using `GetOptions`['columns'], but i agree they not the easiest way.

If you do want to make it public accessible, look at how other methods have been made the same and make a **PR** using develop branch (https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L2111)

https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#pull-requests

djhvscf:
Ref #1943

PR #1943, opened 2016-01-28
added getVisibileColumns method for different utility.

enayet-repo:
There is a method for accessing hidden columns - getHiddenColumns but there is no getVisibleColumns method. 
I have searched in the library and found there is a method - getVisibleFields. But it is not accessible.
Actually adding getVisibleColumns method has more utilities especially to search on the visible columns only.
I think searching is meaning for the visible columns only, otherwise users may be confused with the searched result.


djhvscf:
Fix #1941

wenzhixin:
Sorry, this PR need to update.

wenzhixin:
And please help to update the docs.

enayet-repo:
Updated. Thanks

wenzhixin:
Update docs: https://github.com/wenzhixin/bootstrap-table/blob/master/docs/_i18n/en/documentation/methods.md

enayet-repo:
Updated, thanks again.

djhvscf:
@enayet-repo this **PR** contains a lot of commits that will break our code.. Can you create a new **PR** or revert those commits?? Please just include your new method and the documentation for that method

PR #1955, opened 2016-01-29
Remove extra db call to server when possible

ioctaptceb:
Bootstrap table does a db call for initial data to the server. This
causes the client to make two db calls. one when the client initializes( gets default sorting), and another when filtering the database based on the cookies that are saved on the browser. this change removes that extra initial db call under three
conditions:

1) the cookies option is set to true
2) there are cookies associated to this table
3) all select dropdown filters have default values

this change also updates the documentation for filterControl, in order
to add the proper usage for the property `filterData`.

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -151,6 +151,7 @@
     var BootstrapTable = $.fn.bootstrapTable.Constructor,
         _init = BootstrapTable.prototype.init,
         _initTable = BootstrapTable.prototype.initTable,
+        _initServer = BootstrapTable.prototype.initServer,
         _onSort = BootstrapTable.prototype.onSort,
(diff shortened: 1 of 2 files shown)


djhvscf:
It sounds great!! But I was wondering why you are using a method getCookies but it method only return the cookies ser by filter control extensión.. I think it can be more generic

wenzhixin:
I agree with @djhvscf, I think we needn't to change the code of `bootstrap-table.js`.

ioctaptceb:
Ah, good catch. I will move this over to the cookie extension.

ioctaptceb:
Updated!

wenzhixin:
Great, can you provide an example?

djhvscf:
@ioctaptceb any update here?? We already merged a PR with the filterData documentation added

ioctaptceb:
example is here. http://jsfiddle.net/ym6ztamu/1/

wenzhixin:

PR #1962, opened 2016-02-01
Methods 'updateRow' and 'updateByUniqueId' can now take either a row …

david-maison:
Methods 'updateRow' and 'updateByUniqueId' can now take either a row or an array of rows as parameter.

This change allows an array of rows to be updated at once.

The alternative way is to call N times the "updateRow" for each updated row, however the "initSort" and "initBody" will also be done N times.

Allowing these methods to take an array of rows as parameter will remove the useless overhead cost of "initSort" and "initBody" calls.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2306,19 +2306,24 @@
     };
 
     BootstrapTable.prototype.updateByUniqueId = function (params) {
-        var rowId;
+        var that = this;
+        var allParams = $.isArray(params) ? params : [ params ];
 
-        if (!params.hasOwnProperty('id') || !params.hasOwnProperty('row')) {
-            return;
-        }
+        $.each(allParams, function(i, params) {
+            var rowId;
 
-        rowId = $.inArray(this.getRowByUniqueId(params.id), this.options.data);
+            if (!params.hasOwnProperty('id') || !params.hasOwnProperty('row')) {
+                return;
+            }
 
-        if (rowId === -1) {
-            return;
-        }
+            rowId = $.inArray(that.getRowByUniqueId(params.id), that.options.data);
 
-        $.extend(this.data[rowId], params.row);
+            if (rowId === -1) {
+                return;
+            }
+            $.extend(that.data[rowId], params.row);
+        });
+
         this.initSort();
         this.initBody(true);
     };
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -2306,19 +2306,24 @@
...
 
     BootstrapTable.prototype.updateByUniqueId = function (params) {
-        var rowId;
+        var me = this;
+        var allParams = params instanceof Array ? params : [ params ];
 
-        if (!params.hasOwnProperty('id') || !params.hasOwnProperty('row')) {
-            return;
-        }
+        allParams.forEach(function(params) {
```
djhvscf: Please use the $.each from jQuery in order to support IE <8

src/bootstrap-table.js
```diff
@@ -2306,19 +2306,24 @@
     };
 
     BootstrapTable.prototype.updateByUniqueId = function (params) {
-        var rowId;
+        var me = this;
```
djhvscf: Use `var that = this` in order to follow the standard. **Don't forget change the variable name in the function**

src/bootstrap-table.js
```diff
@@ -2335,10 +2340,16 @@
...
 
     BootstrapTable.prototype.updateRow = function (params) {
-        if (!params.hasOwnProperty('index') || !params.hasOwnProperty('row')) {
-            return;
-        }
-        $.extend(this.data[params.index], params.row);
+        var me = this;
+        var allParams = params instanceof Array ? params : [ params ];
+
+        allParams.forEach(function(params) {
```
djhvscf: Please use the $.each from jQuery in order to support IE <8

src/bootstrap-table.js
```diff
@@ -2335,10 +2340,16 @@
     };
 
     BootstrapTable.prototype.updateRow = function (params) {
-        if (!params.hasOwnProperty('index') || !params.hasOwnProperty('row')) {
-            return;
-        }
-        $.extend(this.data[params.index], params.row);
+        var me = this;
```
djhvscf: Use `var that = this;` in order to follow the standard. **Don't forget change the variable name in the function**

david-maison:
Thanks for the comments. Commit updated.

src/bootstrap-table.js
```diff
@@ -2306,19 +2306,24 @@
     };
 
     BootstrapTable.prototype.updateByUniqueId = function (params) {
-        var rowId;
+        var that = this;
+        var allParams = params instanceof Array ? params : [ params ];
```
wenzhixin: Please use `$.isArray(params)` instead

david-maison:
Commit updated (use "$.isArray" instead of "instanceof Array")

wenzhixin:
Great! Thanks.

PR #1974, opened 2016-02-05
Fix bug: select-control breaks when column name is already an existing class name

adambyer:
An error occurs on line 151 if you are already using the name of a column as a class name. Example: in my case I have a column (db field) named "status" and I also have an existing class named "status". So the selector on line 147 finds elements other than the select element that was dynamically assigned the "status" class. And the 1st matching element is not the select, so options is undefined. To fix I appended the product name to ensure uniqueness.


djhvscf:
Please provide an example with the issue and without the issue @adambyer

adambyer:
Broken:
https://jsfiddle.net/6rtyweex/

Fixed:
https://jsfiddle.net/oLttvabv/4/

wenzhixin:
Thanks, merged [9e7532b](https://github.com/wenzhixin/bootstrap-table/commit/9e7532b62c420f1a433cd9e3b9c0d7bb2422fe5e).

PR #1977, opened 2016-02-06
Cancel previous xhr request on server side pagination

D0d0:
Cancel previous xhr request to server.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1846,7 +1846,7 @@
         if (this.options.ajax) {
             calculateObjectValue(this, this.options.ajax, [request], null);
         } else {
-            if(this._xhr){
+            if(this._xhr) {
                 this._xhr.abort();
             }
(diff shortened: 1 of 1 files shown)


wenzhixin:
can you provide an example?

719media:
+1, i am not sure if this needs an example. just have an ajax request take 10+ seconds on a forced sleep or something and then notice them pile up as you paginate.

src/bootstrap-table.js
```diff
@@ -1846,7 +1846,10 @@
         if (this.options.ajax) {
             calculateObjectValue(this, this.options.ajax, [request], null);
         } else {
-            $.ajax(request);
+            if(this._xhr) {
```
wenzhixin: `if (this._xhr && this._xhr.readyState !== 4) {`

wenzhixin:
@D0d0

Issue #1986, opened 2016-02-10
Additional Arguments in a Formatter?


tannermccoleman:
Is there any way to pass an additional argument to a formatter? The field is an array, and each item in the array has a month. I want it to display the total value of the items from that month. It's messy to create 12 different formatters. Is there another way?

dabros:
@tannermccoleman - share a few of your '12' formatters, your columns list and a subset of your data.

What might be easiest is just keeping those 12 formatters, but make them conveinience handlers for same endpoint, so only one function does actual work and just needs to be called with certain param.

tannermccoleman:
[Here's a fiddle with a really basic example](https://jsfiddle.net/tannermccoleman/2srpp9d7/). I'm already using one central function for the business logic, but it would be ideal to not have to write a factory to create a different formatter for each and every column. There are other cases (such as Date columns) where passing an additional parameter (such as date format) would cut down on a lot of needless, near useless code.

dabros:
Searched for few min to try and recall something, found #1856 and here is your answer

```javascript

App.receivableCounterAny = function (value, row, index) {
  console.log(this);
  return App.receivableCounter(this.monthid, value, row, index);
}

```

You just need to add a `data-monthid` via any means to the column, as any extra `data` attributes become part of the column object, accesible via `this`

tannermccoleman:
@dabros Perfect! I hadn't noticed the mention of `this` context in the documentation. Thank you so much!

tannermccoleman:
I ran into a bit of a new issue: When the same field and formatter is used, the `this` context does not update from the first occurrence. [In this example, January is used for all 12 fields](https://jsfiddle.net/tannermccoleman/sk5yzabf/2/).

dabros:
@tannermccoleman - just saw your **PR**, but here is workaround if needed: https://jsfiddle.net/dabros/1z3cf1z0/

tannermccoleman:
[Here is my original example with my pull request](https://jsfiddle.net/tannermccoleman/q5dkq122/)

wenzhixin:
Fixed this issue in PR: #1679.

https://jsfiddle.net/wenyi/sk5yzabf/4/

wenzhixin:
BTW, I also agree with the solution of @dabros.

PR #1988, opened 2016-02-12
Fix context for cols with same field and formatter

tannermccoleman:
Fix issue discovered in #1986 where columns of the same field and formatter would only ever send the context of the first column to said formatter


wenzhixin:
Merged the same problem in #1679, thanks.

PR #1992, opened 2016-02-15
Fix #1991 - Call to `resetFooter` after save

### LINKED ISSUE #1991: Footer doesn't update after X-Editable change
I have a table that has 6 months and quantities for each month when I update the quantity the footer row does not update to the new sum until I refresh the page.

image

Updating the column seems to do a string concatenation of all the quantity values. i am using the oneditsave function and passing in the RefreshOptions method with the columns. Result shown below:

image

dabros:
Footer contents updated via call to `resetFooter` (which itself checks for `showFooter`) following call to save changes.
Fix for issue reported in #1991 
fiddle: http://jsfiddle.net/dabros/Lr056k01/1/


djhvscf:
Fix #1991

wenzhixin:
Thanks!

PR #2006, opened 2016-02-18
Fix #1578 : group-columns not updating on data-visible

### LINKED ISSUE #1578: show-columns and group-columns are incompatible
I think it would make more sense to have the items in the first row included in the show-columns menu.
This would allow hiding a column with all its subitems.

dabros:
Fix #1578 - group-columns with varying data-visibility of any columns never updated colspan nor hid uneeded group-column/parents.

See usage at http://jsfiddle.net/dabros/92a57s7L/10/ (where local copy disabled and external resources posting to a specific commit is used)

Im still unsure whether my `vColumns` code (in `SetFieldIndex`) is most efficient approach, but setting parents there and overwriting `data-visible` & `colspan` in `initTable` is best approach i can think of.

I also take the assumption that lowest level (last group with parents) decide overall `data-visible` of the parents, but that makes sense as fiddle proves this works in usual approach and if you hide parent group-column but have child visible, that is user error so ofcourse parent group-column should be visible.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -737,7 +737,7 @@
                     $.each(column.parents,function(k,lookupStr) {              
                     	var lookupKeys = lookupStr.split("."),
                       	    parentCol = that.options.columns[lookupKeys[0]][lookupKeys[1]];
-                        parentCol.colspan += (column.visible) ? 1 : -1;
+                        parentCol.colspan += parentCol.colspan += (column.visible) ? 1 : (parentCol.colspan > 0) ? -1 : 0;
                         parentCol.visible = (parentCol.colspan > 0);
                   	});
                 }
@@ -799,6 +799,10 @@
                     }
 
                     visibleColumns[column.field] = column;
+                    
+                } else if (!column.visible) {
+                    // Fix #1578 - needed to pickup on multi-level th visibility
+                    return;
                 }
 
                 html.push('<th' + sprintf(' title="%s"', column.titleTooltip),
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -732,12 +732,12 @@
             
         $.each(this.options.columns, function(i, columns) {
             $.each(columns, function(j, column) {
-                column.colspan = 0;
+                column.colspan = 1;
                 if(typeof column.parents !== 'undefined') {
                     $.each(column.parents,function(k,lookupStr) {              
                     	var lookupKeys = lookupStr.split("."),
                       	    parentCol = that.options.columns[lookupKeys[0]][lookupKeys[1]];
-                        parentCol.colspan += parentCol.colspan += (column.visible) ? 1 : (parentCol.colspan > 0) ? -1 : 0;
+                        parentCol.colspan += (column.visible) ? 1 : (parentCol.colspan > 0) ? -1 : 0;
                         parentCol.visible = (parentCol.colspan > 0);
                   	});
                 }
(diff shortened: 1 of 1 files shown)


djhvscf:
I'm not sure if your **PR** is working because if I hide some columns the `TH` is not hidden

dabros:
@djhvscf - yeh, i see it

i did think i had it working fine at a point, maybe my clearing it, or initial visibility perhaps...

will fix soon

dabros:
@djhvscf - so it wasnt my change, rather that the `initHeader` `column.visible` check was only inside a `fieldindex` check, so didnt affect interim levels.

Also a sanity check to the colspan computation to better handle multi-level multi-col combinations, as they obviously more problematic.

Not sure how that skewed my initial testing, but thanks for the spot.

Making change now, here are fiddles:

- inline changes: http://jsfiddle.net/dabros/92a57s7L/12/

dabros:
Testing now

djhvscf:
@dabros if I hide the B columns the AB header is hidden also, which is incorrect

dabros:
@djhvscf - thanks, but i was posting then testing as issues kept coming up

Found that my test results skewed because of `fieldIndex =  getFieldIndexFromColumnIndex(that.columns, i);` replacing `fieldIndex = $.inArray(field, that.getVisibleFields());` in one use case, but not any other use case for `getVisibleFields()`. It kept throwing off certain tests, so my changes never married up with my jsfiddle only version (as i was testing with both side-by-side and didnt think to test for new code in develop)

Please advise why we need this function to differ for that one call, and whether i need to work around it or why it should differ so

https://github.com/wenzhixin/bootstrap-table/commit/4fbc9304d14ba6a3b2d48007e8100776df8e39cc

http://jsfiddle.net/dabros/92a57s7L/14/

Thanks

dabros:
```javascript

    var getFieldIndexFromColumnIndex = function (columns, fieldIndex) {
        $.each(columns, function (i, column) {
          if (!column.visible) {
            fieldIndex--;
          }
          if (i == fieldIndex) {
            return false;
          }
        });
        return fieldIndex;
    };
```

My thought is that this function is inherently broken as it doesnt seem to regard multi-level in that fieldindex might belong to a column with `rowspan > 1` in previous columns index/row.

djhvscf:
@dabros exactly that I was thinking about.. I think this function does not support the multi-level header.. remove it.. @wenzhixin do you recall something about this function `getFieldIndexFromColumnIndex`

dabros:
@djhvscf - looks like its more you than @wenzhixin ;)

https://github.com/wenzhixin/bootstrap-table/pull/1664

I cant make too much sense out of it, so can you please read then see if a problem?

djhvscf:
Oh! Yes.. Now I remember the issue.. I think @dabros you have to update that function in order to fix our root issue here

dabros:
@djhvscf - would it be better to stop using `column.fieldname` in similar cases (and havent to compute multi-level visibility each and every time we need it for those cases) and instead have `column.uid` which would be either "fieldname.instance" or just "instance".

I noticed that #1988 has `fieldCounts` which looks to be similar, though unsure if its saved to column object or not.

If that change still in current version, then its very discordant as we use `fieldCounts`, `getVisibleColumns` and `getFieldIndexByColumnIndex` for same basic purpose.

djhvscf:
@dabros any update here??

dabros:
@djhvscf - I started solving some tricky issues, took longer than expected and work picked up so now have 8 open issues/tabs. At least 2 of them (as well as this one) my last post was a question to other users.

For this issue, please scan my posts and give thoughts on the question I asked about how best to tackle this fieldname to avoid the numerous issues and potential bugs, as it is also a common theme through related bugs for extensions.

PR #2012, opened 2016-02-20
Add data variable to post-body trigger

thornomad:
The `onPreBody` event is passed the `data` variable, however, the `onPostBody` is not.  Am not sure if this is by design or just hadn't been implemented.

In my one use case, I want to execute the method `mergeCells` in `onPostBody()` and need access to the `data` variable.


wenzhixin:
Thanks

thornomad:
My pleasure!  Anything I can do to help since bootstrap-table has made my life so much easier!

wenzhixin:
:)

Issue #1984, opened 2016-02-09
Table elements inside bootstrap table gets select also


ewsun:
There is javascript error when table td cell contains other regular table elements( not bootstrap table elements ).  That is because in BootstrapTable.prototype.initTable function,  it selects all children elements not the immediate child.

Example:
<pre><code> this.$el.find(<b>'tbody tr'</b>).each(function () {
    var row = {};

    // save tr's id, class and data-* attributes
   ...
    $(this).find(<b>'td'</b>).each(function (i) {
        var field = that.options.columns[i].field;
        ...
    });
   ...
});</code></pre>
This will select all the tds which include the tds in its child elements.

It should use immediate selector:
<pre><code>
this.$el.find(<b>'>tbody>tr'</b>)<br>
$(this).find(<b>'>td'</b>)<br></code></pre>

djhvscf:
Can you open a new **PR**, @ewsun ?

ewsun:
Sorry, I am new to GitHub.  This is my first issue report. :).  Did you mean create a new Pull Requests?

djhvscf:
Exactly.. Create it please and provide a fiddle with the issue and another one with the issue fixed

dabros:
https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#pull-requests

djhvscf:
@ewsun any update here?

Falseee:
JsFiddle with issue : http://jsfiddle.net/1k715pjz/1/
JsFiddle without issue : http://jsfiddle.net/n9w053h5/1/

PullRequest: #2037

wenzhixin:
Merged

PR #2037, opened 2016-02-27
use a better selector

Falseee:
#1984


djhvscf:
Fix #1984

wenzhixin:
Thanks @Falseee

PR #2038, opened 2016-02-27
Patch 3 - New Feature for Editable extension

dreamerworks:
With this update, the we get the ability to allow some editable cells to become non editable.
Scenario:
**Columns**  Person, Married (y/n), Spouse name
**Data**
John, y, Mary
Paul, n, -

Imagine that the column "Spouse name"  is editable. In the case of Paul, he is not married, so we don't want the "Spouse name" to be editable. Instead we want to be able to provide some other html like "-" or "Not applicable".

For this, we just need to add to the current editable options a new property, like this:
```
editable: {
    type:"text";
    noedit:function(value, row, index) {
        if (row.married=='y') {
          return false;  // return false if you want the field editable.
        } else {
          return '<span class="fa fa-ban" title="single people have no spouse."></span>';
        }
    }
}
```


Falseee:
Please add a JsFiddle, so we can check if your solution/code is working.
Why you have so many crazy commits with the same bad commit message ?

dreamerworks:
Hi, here is my [jsFiddle](https://jsfiddle.net/dreamerworks/kwk7dot9/18/).
Sorry about all those commits, I thought I could fix the code directly on gitHub in one edit.... turns out I failed all those times you see there. Then I fixed it on my local machine and made the final commit.

wenzhixin:
Thanks for your fiddle, I think it is better to use `noeditFormatter` instead, what do you think @djhvscf @dabros?

* Please help to update the `README.md` file.
* [Please squash your commits together](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history)

dabros:
@wenzhixin - do you mean the param name, in which case i agree with that vs 'noedit'

It perhaps edges over `isEditable` and `isEditableFormatter` name as it is focused on the negative, only really needed if conditions should disable editing.

Though i feel `isEditable` would make more sense if could return `bool` OR `string`, with a string being the fail message shown when trying to edit perhaps...

However the fiddle is broken atm, so i cant say much about the solution itself.

Notably, i would be interesting in looking more into `disabled` and similar https://vitalets.github.io/x-editable/docs.html features to better implement this feature via UI feedback.

dreamerworks:
I see what you mean, I changed the code to noEditFormatter as it was clearly better than "noEdit". The problem I see with the name "isEditable" is that it may be counter intuitive to say:
```
editable: {
  isEditable: function(...) {... return false;}
}
```

At the moment I am kind of lost between the "display" and "formatter"... will need to study a bit more. 
The only contribution I can give, is that the contents of the "not editable" cell should be able to be programmed having access to the "row" data and should allow for arbitrary html to be returned.

wenzhixin:
Thanks, I will consider what @dabros said, and add it to the README.

PR #2049, opened 2016-03-02
Add support for filter control select list data to be passed in as JSON string and filter control starts with search

fenichelar:
For my application, `var:variable` would not work as the variable is not accessible using `window`. This allows an object to be passed in like so:

    var data = {
      key1: 'text1',
      key2: 'text2'
    };
    filterData = 'jso:' + JSON.stringify(data);

`filterStartsWithSearch = true;` makes the search be a 'starts with' search. The default is false. If `filterStrictSearch = true` then `filterStartsWithSearch` is ignored, which makes sense because strict search implies starts with.

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -222,6 +222,12 @@
                             addOptionToSelectControl(selectControl, key, variableValues[key]);
                         }
                         break;
+                    case 'jso':
+                        var variableValues = JSON.parse(filterDataSource);
+                        for (var key in variableValues) {
+                            addOptionToSelectControl(selectControl, key, variableValues[key]);
+                        }
+                        break;
                 }
             }
         });
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -325,7 +325,8 @@
         filterControl: undefined,
         filterData: undefined,
         filterDatepickerOptions: undefined,
-        filterStrictSearch: false
+        filterStrictSearch: false,
+        filterStartsWithSearch: false
     });
 
     $.extend($.fn.bootstrapTable.Constructor.EVENTS, {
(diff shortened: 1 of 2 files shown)


djhvscf:
Provide an example, please @fenichelar

fenichelar:
[http://jsfiddle.net/fenichelar/bchswsn2/](http://jsfiddle.net/fenichelar/bchswsn2/)

djhvscf:
Great!!

djhvscf:
@fenichelar I misunderstood the `filterStartsWithSearch`option.. Can you explain us a bit more?

fenichelar:
Let's say you type in 'dog' in the filter control box.

`filterStrictSearch = true` means if the item is EXACTLY equal to 'dog', for example 'dog'.
`filterStartsWithSearch = true` means if the item STARTS WITH 'dog', for example 'dog', 'dogs are cool', 'dogdog', etc.
The default is if the item CONTAINS `dog`, for example 'dog', 'dogs are cool', 'dogdog', 'catdog', 'cats like dogs', etc.

Make sense?

fenichelar:
If you set `filterStrictSearch = true` and `filterStartsWithSearch = true`, it uses `filterStrictSearch` because everything that would pass 'exactly equal to' would also pass 'starts with'.

djhvscf:
Ok.. I saw that `name` column has the `filterStartsWithSearch = true` but if search `1` the table is not filtered as expected @fenichelar

djhvscf:
And also see this `PR` https://github.com/wenzhixin/bootstrap-table/pull/2050 because I applied some changes

fenichelar:
I'm not following... Are you saying the filtering isn't working?

fenichelar:
If I type "1" in the filter control input for "name", no results are found. That is the expected behavior as none of the names start with the character "1". One of them ends with the character "1", but it doesn't start with "1".

djhvscf:
Aaaah OK OK. . Sorry, now I understand it

wenzhixin:
Thanks!

src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -222,6 +222,12 @@
                             addOptionToSelectControl(selectControl, key, variableValues[key]);
                         }
                         break;
+                    case 'jso':
+                        var variableValues = JSON.parse(filterDataSource);
+                        for (var key in variableValues) {
+                            addOptionToSelectControl(selectControl, key, variableValues[key]);
+                        }
+                        break;
                 }
```
djhvscf: This switch should not be here.. I remove it in a **PR** @wenzhixin

PR #2056, opened 2016-03-05
fixes #2055 - recursively search `data`

### LINKED ISSUE #2055: Search only examines top level item elements
If you are using a nested object structure in your `data` the search will not penetrate past the first level.  If you are using a data field that is, for example, `location.city` it will not search it.

See: https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L1190-L1214

Because it is using `for key in item` it is only checking the top level of the object.  There needs to be a recursive iteration to go through each level of the data.

It might be better than to go through each item in the `data` to instead start with the `keys` in the header fields and approach it that way.  Folks may have a lot of data in their response that isn't being display in the table. 

#### fiddle: http://jsfiddle.net/thornomad/qt69a08f/
I used the base fiddle and just nested the data once; as you can see search does not work.

thornomad:
If a `field` is referencing a nested value (e.g., `location.city`) it was not being searched; only top level fields were.  This should fix that (#2055).

Here are the jsfiddles:

Problem: http://jsfiddle.net/thornomad/qt69a08f/
Solution: http://jsfiddle.net/thornomad/b106ogvb/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1187,20 +1187,33 @@
             }) : this.options.data;
 
             this.data = s ? $.grep(this.data, function (item, i) {
-                for (var key in item) {
-                    key = $.isNumeric(key) ? parseInt(key, 10) : key;
-                    var value = item[key],
-                        column = that.columns[getFieldIndex(that.columns, key)],
-                        j = $.inArray(key, that.header.fields);
+                for (var j = 0; j < that.header.fields.length; j++) {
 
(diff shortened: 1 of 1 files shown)


Falseee:
The fiddles have the same content

thornomad:
Thank you!  I guess when I changed the `bootstrap-table.js` resource I
didn't save it; try it again.  it should be working now!

wenzhixin:
Thanks.

PR #2059, opened 2016-03-07
Fix methods updateRow and updateByUniqueId

david-maison:
### Problem
In a grid with the "search" feature, the `updateRow` and `updateByUniqueId` methods doesn't correctly update the data when something has been writen by the user in the "search box".

The technique issue (from my point of view, maybe I'm wrong) is the use of `this.data` instead of `this.options.data` in these both methods. From what I understand, `this.data` is used for data after being sorted, filtered... and `this.options.data` for the original data.

If what I understand is correct, then I saw `this.data` used in other methods (like `insertRow`), but I'm waiting for a confirmation about if I'm wrong or not.

#### Fiddle link:
- with the bug (current master): https://jsfiddle.net/Lightsephi/fger316d/
- without the bug (this branch): https://jsfiddle.net/Lightsephi/cjkddhk0/

#### Action to follow:
- enter "2" in the search box
- click on the "Update row" button
- observe results (1)
- clear the search box content
- observe again the results (2)

#### Expected behavior:
at (1), the second line should appear with the new value "2"
at (2), both lines are shown correctly with "2" in the value column

#### Actual behavior:
at (1), nothing change
at (2), the second line has still "3" in the value column. The row hasn't been updated.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2392,9 +2392,10 @@
             if (rowId === -1) {
                 return;
             }
-            $.extend(that.data[rowId], params.row);
+            $.extend(that.options.data[rowId], params.row);
         });
 
+        this.initSearch();
         this.initSort();
         this.initBody(true);
     };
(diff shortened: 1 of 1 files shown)


djhvscf:
@david-maison I think this fix is ok because the `this.data` array is when the data is filtered and the `this.options.data` is the original data array and we should update the original data array

What do you think, @wenzhixin?

wenzhixin:
Thanks @david-maison @djhvscf

PR #2063, opened 2016-03-09
Adjusted the clicked column logic to take into account visibility.

charettes:
An incorrect field is detected as soon as there an invisible column before the clicked cell's column.

Here's a test case to reproduce:

https://jsfiddle.net/ojf0035g/2/

Note how the first column (`Item ID`) is configured to be non clickable while it's not the case for the second one (`Item Name`) since it inherits the table `data-click-to-select="true"` option. If you try to click on a cell of the `Item Name` column you'll notice nothing happen.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1728,7 +1728,7 @@
                 $tr = $td.parent(),
                 item = that.data[$tr.data('index')],
                 index = $td[0].cellIndex,
-                field = that.header.fields[that.options.detailView && !that.options.cardView ? index - 1 : index],
+                field = that.getVisibleFields()[that.options.detailView && !that.options.cardView ? index - 1 : index],
                 column = that.columns[getFieldIndex(that.columns, field)],
                 value = getItemField(item, field, that.options.escape);
 


charettes:
Is there any way to provide a regression test?

djhvscf:
@charettes See this example using your [changes](https://github.com/charettes/bootstrap-table/blob/clickToSelect-visible-columns/src/bootstrap-table.js): https://jsfiddle.net/djhvscf/ojf0035g/3/

charettes:
@djhvscf, it works perfectly, am I missing something?

djhvscf:
No, everything is ok.. just waiting for @wenzhixin

charettes:
Here's a more detailed case to make review easier:

[Before](https://jsfiddle.net/ojf0035g/12/)
[After](https://jsfiddle.net/ojf0035g/14/)

djhvscf:
Your examples are very helpful @charettes

wenzhixin:
I found this problem is fixed in the develop branch: https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L1800, thanks @charettes @djhvscf

PR #2066, opened 2016-03-09
escape special characters, add additional translations

rmjoia:
to prevent encoding issues, added html encoded special characters, also, since some translations were showing in Kanji, added the missing translations to Portuguese pt-PT


wenzhixin:
Thanks

PR #2075, opened 2016-03-12
Fixes #2074 - return field from visible cells

### LINKED ISSUE #1841: onClickCell response strange when hide the column before the column bind with onClickCell
Three Column: id name city ...
column is bound with onClickCell
when i hide the column id, the onClickCell event is only triggered on column city
why?

### LINKED ISSUE #2074: onClickCell provides parameters for adjacent cells when columns are not visible
I believe this is the same problem described in #1841 - although an example was not provided.

Here is a fiddle: http://jsfiddle.net/thornomad/6mhjovgj/

The first column is hidden: `data-visible="false"`.  If you click on a cell it will return the cell from the column previous to it.  If you hide a second column (e.g., `stars`) it will return a value two columns previous to it.  If no columns are hidden, it works as expected.

**Any cell clicked _after_ a hidden column will return the cell value minus the number of hidden columns prior to it.**

thornomad:
See issue #2074 - should fix that problem.

Most likely also fixes #1841 (if I understand the question correctly).

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1728,7 +1728,8 @@
                 $tr = $td.parent(),
                 item = that.data[$tr.data('index')],
                 index = $td[0].cellIndex,
-                field = that.header.fields[that.options.detailView && !that.options.cardView ? index - 1 : index],
+                fields = that.getVisibleFields(),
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks!

PR #2087, opened 2016-03-16
field name with dot character

raku:
Prevent error when field name is like this "field.name".
I escape field name so filter inputs can be found using jquery, because you are searching an object in jquery like $(.field.name) you will get undefined.


djhvscf:
This PR is not valid because you made the changes in the dist folder

djhvscf:
Open a new PR in the correct folder

raku:
ok, I will push it tonight

PR #2088, opened 2016-03-17
add constant in order to get it from angular scope

xidui:
get the jQuery table element in the angular scope:
```js
var app = angular.module("some_module", ['bsTable']);
app.controller('some_controller', ['$scope', 'uiBsTables', function($scope, uiBsTables){
    // you can then operate the tables via uiBsTables variable.
}]);
```


wenzhixin:
Thanks!

Issue #2120, opened 2016-03-28
fail to render editable columns after 'refreshOptions'


luryson:
https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/editable/bootstrap-table-editable.js#L63
When first time initializing the table here, `column.formatter` has been  overrode.'refreshOptions' method still call initTable function while column.formatter is no longer the function I defined. 
So I got this:
 
![image](https://cloud.githubusercontent.com/assets/7643503/14080743/8658dfe4-f539-11e5-983a-34cde7d015af.png)
for the first time, and this after 'refreshOptions': 
![image](https://cloud.githubusercontent.com/assets/7643503/14080773/a41fb480-f539-11e5-8aa4-22fa674ca1fc.png)
check this: https://github.com/luryson/bootstrap-table/commit/761f8517100b95a2df4952f35f5b0dfd3a7a9e8f

djhvscf:
sounds good.. can you provide an example?

airtonjr:
Maybe same thing here:

Export And Editable Plugin can't work together. Possible bug? #2116

luryson:
@airtonjr If you have better way to solve this problem, please let me know

airtonjr:
@luryson I am trying with no success.

luryson:
@airtonjr  I solved my problems temporarily  https://github.com/luryson/bootstrap-table/commit/761f8517100b95a2df4952f35f5b0dfd3a7a9e8f .I am not sure your problem is exactly the  same one, just try it. I store the formatter function I defined in `column` as  `column._formatter`. When 'refreshOptions' method is called, `column._formatter` will be checked and called firstly.

luryson:
@djhvscf http://runjs.cn/detail/jod6mf1k here is the example. Every time I press 'press me' button I got this:
![image](https://cloud.githubusercontent.com/assets/7643503/14098671/01188910-f5b1-11e5-90d6-fd1e51c0bd6b.png)  
![image](https://cloud.githubusercontent.com/assets/7643503/14098681/12d73f66-f5b1-11e5-9049-3db9866b4e1b.png)

djhvscf:
ok, but can you provide an example of your solution?

luryson:
@djhvscf  Sorry.  The example above has been modified with my solution.http://runjs.cn/detail/jod6mf1k

djhvscf:
great... lets create a new **PR**

airtonjr:
[...]

PR #2123, opened 2016-03-29
editable columns error when refreshOptions

luryson:
failed to render editable columns after 'refreshOptions'. https://github.com/wenzhixin/bootstrap-table/issues/2120. I stored the formatter function defined in `columns` option as  column._formatter. When 'refreshOptions' method is called, `column._formatter` will be checked and called firstly.

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -60,6 +60,9 @@
 
             $.each(that.options, processDataOptions);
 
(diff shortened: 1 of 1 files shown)


djhvscf:
Can you provide an example with this change

luryson:
ok. Here are examples.
before: http://runjs.cn/detail/gp3swfyj
after: http://runjs.cn/detail/jod6mf1k

wenzhixin:
Thanks!

PR #2133, opened 2016-03-31
multisort fixes

719media:
I've made the multisort work with server (ajax) tables, and cleaned up the display to make it less cluttered


719media:
going to move this to another branch

PR #2136, opened 2016-03-31
Integrate bootstrap4 classes

719media:
Integrate bootstrap4 classes.
Also moved the data('bootstrap.table') storage so that the data attribute is accessible immediately. This prevents the need for using setTimeout() when overriding the ajax options, for example, which is very confusing and unwieldily.


djhvscf:
Can you provide an example, please?

719media:
i just did one more commit to use some new properties:
paginationHAlignClass
paginationDetailHAlignClass
searchAlignClass
buttonsAlignClass
toolbarAlignClass

My personal opinion is to deprecate the old properties (it isn't really more difficult to have people use "pull-right" vs. "right" for bootstrap 3 compatibility.

Anyway:
fiddle using bootstrap4 with 1.10.1 version of bootstrap-table
https://jsfiddle.net/719media/493tbzfa/1/

fiddle using bootstrap4 with patched version:
https://jsfiddle.net/719media/493tbzfa/2/

wenzhixin:
Close ref #3428. Thanks for your contribution!

PR #2138, opened 2016-03-31
toolbar order

719media:
For backwards compatibility, I wasn't able to make it as simple as I would have liked.

Two new properties: toolbarOrder and toolbarItems.

toolbarItems: stores toolbarTemplates (and later on, maybe more information?) that can be overridden by overriding options (added bonus feature).

toolbarOrder is used to define the order.

the name of the four "built-in" toolbar items are kept the same (columns, paginationSwitch, refresh, toggle) so that variables can be predicted from these names (e.g. showColumns, showPaginationSwitch, showRefresh, showToggle)

it also makes it so that future plugins don't have to override initToolbar to get plugins into the system, they can just add a toolbar template to the option and a corresponding showToolberName option.

Any other ideas/etc. to make better just let me know


719media:
fiddle using 1.10.1 bootstrap-table and four built-in toolbar items
https://jsfiddle.net/719media/fgq94gbv/1/

fiddle using 1.10.1 bootstrap-table and toolbarOrder
https://jsfiddle.net/719media/fgq94gbv/3/
In the patched example, I included an external plugin (multi-sort) to show that it still works with external plugins. However, external plugins would need to register their plugin with toolbarItems for it to be able to be ordered.

djhvscf:

Issue #2141, opened 2016-04-01
Reset sorting order


jonatanklosko:
There's no option to provide the most natural behaviour. I mean the sort state should be reseted, so: user clicks first sort arrow that sorts let's say ascending, then clicks another arrow, and then the first one again; the user would expect that the sort order is ascending again. So only clicking two times on the same arrow sorts descensing in this case.
[That's code which I wrote to deal with this.](https://github.com/jonatanklosko/worldcubeassociation.org/blob/45cb3f0ab097b5e2f1fafcba0da4d9810038e608/WcaOnRails/app/assets/javascripts/application.js#L245-L256)

djhvscf:
This solution looks good!! Lets open a new **PR**, please @jonatanklosko

PR #2145, opened 2016-04-01
Add rememberOrder option to table

jonatanklosko:
Applies the changes indicated in #2141.
This provides more natural and intuitive behaviour. When the option is set to false, on a first click on a sort arrow for a column that is not currently sorted, table always gets sorted in the same order, thus only when followed by a second click it gets sorted in the opposite order.

### DIFF
--- a/docs/_i18n/en/documentation/table-options.md
+++ b/docs/_i18n/en/documentation/table-options.md
@@ -87,8 +87,8 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
         <td>rememberOrder</td>
         <td>data-remember-order</td>
         <td>Boolean</td>
-        <td>true</td>
-        <td>Set <code>false</code> to reset the order to the default (set by column's order attribute) on every sort action for a column that is not currently sorted.</td>
+        <td>false</td>
+        <td>Set <code>true</code> remember the order for each column.</td>
     </tr>
     <tr>
         <td>iconsPrefix</td>
--- a/docs/_i18n/en/documentation/table-options.md
+++ b/docs/_i18n/en/documentation/table-options.md
@@ -84,6 +84,13 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
         <td>True to get a stable sorting. We will add <code>_position</code> property to the row.</td>
     </tr>
     <tr>
+        <td>rememberOrder</td>
+        <td>data-remember-order</td>
+        <td>Boolean</td>
+        <td>true</td>
+        <td>Set <code>false</code> to reset the order to the default (set by column's order attribute) on every sort action for a column that is not currently sorted.</td>
+    </tr>
+    <tr>
         <td>iconsPrefix</td>
         <td>data-icons-prefix</td>
         <td>String</td>
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -286,6 +286,7 @@
         sortName: undefined,
         sortOrder: 'asc',
         sortStable: false,
+        rememberOrder: true,
         striped: false,
         columns: [[]],
         data: [],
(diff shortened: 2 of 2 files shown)


djhvscf:
Please provide an example

jonatanklosko:
https://jsfiddle.net/oxL0adyx/2/

docs/_i18n/en/documentation/table-options.md
```diff
@@ -70,6 +70,13 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
         <td>Defines the column sort order, can only be 'asc' or 'desc'.</td>
     </tr>
     <tr>
+        <td>rememberOrder</td>
+        <td>data-remember-order</td>
+        <td>Boolean</td>
+        <td>true</td>
```
jfly: I feel like everyone is going to want to set this option to false, and this is very subtle behavior that many people won't even realize that they want. I realize it would be a "breaking" change, but how about defaulting this option to false? I personally don't see any value in this option, so I'd even vote for removing it entirely and unconditionally switching to the new behavior.

jonatanklosko: I definitely agree with you and waiting for maintainers response.

wenzhixin: I agree with you. Let's change to `false`.

src/bootstrap-table.js
```diff
@@ -1006,7 +1007,11 @@
             this.options.sortOrder = this.options.sortOrder === 'asc' ? 'desc' : 'asc';
         } else {
             this.options.sortName = $this.data('field');
-            this.options.sortOrder = $this.data('order') === 'asc' ? 'desc' : 'asc';
+            if(this.options.rememberOrder) {
```
wenzhixin: `if (this.options.rememberOrder) {`

src/bootstrap-table.js
```diff
@@ -1006,7 +1007,11 @@
...
+                this.options.sortOrder = $this.data('order') === 'asc' ? 'desc' : 'asc';
+            } else {
+                this.options.sortOrder = this.options.columns[0].find(function(option) { return option.field === $this.data('field'); }).order;
```
wenzhixin: Please update [find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) method.

| Feature | Chrome | Firefox (Gecko) | Internet Explorer | Edge | Opera | Safari |
| --- | --- | --- | --- | --- | --- | --- |
| Basic support | 45.0 | 25.0 (25.0) | No support | 12 | No support | 7.1 |

wenzhixin:
This branch has conflicts that must be resolved
 Use the command line

jonatanklosko:
As @jfly wrote it could be better to remove the option and switch to the current behaviour, but that's you choice of course.

jonatanklosko:
@wenzhixin Is there anything more to do to let it be merged up?

jonatanklosko:
@wenzhixin @djhvscf Could you provide any response, please? Also the label 'needs example' is no longer appropriate, I suppose (https://jsfiddle.net/oxL0adyx/4/).

jfly:
Bump. Any update here? I realize there are conflicts again, but it's not really fair to ask @jonatanklosko to resolve the conflicts just to be ignored again.

djhvscf:
Hi @jonatanklosko @jfly , I will try to solve the conflicts

jonatanklosko:
Actually those were pretty straightforward, so I just made a quick rebase.

djhvscf:
@wenzhixin I think we merge this **PR**

wenzhixin:
Thanks for your contribution! @jonatanklosko @djhvscf

jfly:
It looks like this didn't make it into release `0.11.1`. Any ideas when the next release containing this will be? Thanks!

PR #2151, opened 2016-04-03
Update bootstrap-table-uk-UA.js

juliyvchirkov:
petty update to turn the Ukrainian localization more accurate


wenzhixin:
Thanks!

PR #2155, opened 2016-04-03
Fix the clickToSelect to work correctly with links

### LINKED ISSUE #2154: Clicking links while using clickToSelect option
At the moment when we enable the clickToSelect option for checkboxes and have some links in our table, clicking them causes that a row is selected. It's a little bit confusing if a user for example middle-click on a link to open it in a new tab and the row gets selected. [Here's my workaround.] (https://github.com/jonatanklosko/worldcubeassociation.org/blob/sortable-registrations/WcaOnRails/app/assets/javascripts/application.js#L268-L271)

jonatanklosko:
Fixes #2154. As described the row should be selected on click only when the target is not a link, so user can easily open a link in a new tab with middle-click without selecting the checkbox.
Example: https://jsfiddle.net/20rxmcxt/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1670,8 +1670,8 @@
             that.trigger(e.type === 'click' ? 'click-cell' : 'dbl-click-cell', field, value, item, $td);
             that.trigger(e.type === 'click' ? 'click-row' : 'dbl-click-row', item, $tr);
 
-            // if click to select - then trigger the checkbox/radio click
-            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect) {
+            // if click to select - then trigger the checkbox/radio click unless the target is a link
+            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect && !$(e.target).is('a')) {
                 var $selectItem = $tr.find(sprintf('[name="%s"]', that.options.selectItemName));
                 if ($selectItem.length) {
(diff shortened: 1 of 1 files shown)


src/bootstrap-table.js
```diff
@@ -1670,8 +1670,8 @@
             that.trigger(e.type === 'click' ? 'click-cell' : 'dbl-click-cell', field, value, item, $td);
             that.trigger(e.type === 'click' ? 'click-row' : 'dbl-click-row', item, $tr);
 
-            // if click to select - then trigger the checkbox/radio click
-            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect) {
+            // if click to select - then trigger the checkbox/radio click unless the target is a link
+            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect && !$(e.target).is('a')) {
```
jfly: What about other things users can click on: `<input>`, `<button>`, `<textarea>`s? I don't know how many of those stopPropagation on the click event.

wenzhixin:
I am agree with @jfly, for your problem, you can set the `clickToSelect` column option to `false`: https://jsfiddle.net/wenyi/20rxmcxt/1/.

Thanks anyway.

jfly:
Ohh, interesting. That's not as fine of granularity as just ignoring the `<a>` tags, though. You could imagine a paragraph of text with one measly link it in. Turning off click-to-select on that entire column would be throwing out the baby with the bathwater.

Issue #2146, opened 2016-04-02
Contributing Fix for Filter Control V1.0.0


jamesmikesell:
I've made some changes/fixes to the bootstrap-table-filter-control.js v1.0.0 to:
1.) Populate select lists with all unique column values rather than just those on the first page(s) (if paging is enabled)
2.) Improved performance when creating select lists (large datasets took used to take several seconds)

My questions
1.) Are you interested in pulling these changes?
2.) Do they go against the wenzhixin/bootstrap-table repo (i didn't see a specific fork for the filter control)

Hope this is the right place to ask this question.  Also, bear with me, this will be my first contribution on github.

djhvscf:
This is the right place where you can create the PR...so go a headset and create the PR

jamesmikesell:
Created pull request #2147

wenzhixin:
Merged

PR #2164, opened 2016-04-06
Not all values are populating in select filter, and related performance improvements

jamesmikesell:
This was in reference to #2146

Examples:

Original V1.0.0
https://jsfiddle.net/rider2112/thar3xu5/

Fixed Version
https://jsfiddle.net/rider2112/ugs4jx6r/

Logic changes:
Looping through all data in the table instead of just the visible data.
Using for loop for improved performance over $.each
Creating a map with only unique values.
Upon finding all unique column values, the adding them to the select
control.
Sort select control after all values have been added.

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -11,13 +11,29 @@
     var sprintf = $.fn.bootstrapTable.utils.sprintf;
 
     var addOptionToSelectControl = function (selectControl, value, text) {
-    	value = $.trim(value);
-    	selectControl = $(selectControl.get(selectControl.length - 1));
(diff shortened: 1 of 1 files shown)


djhvscf:
@jamesmikesell you must resolve the conflict that this **PR** has

jamesmikesell:
@djhvscf @wenzhixin  So is my only option to make this change against whatever the most recent version of [bootstrap-table-filter-control.js](bootstrap-table/src/extensions/filter-control/bootstrap-table-filter-control.js) is in the development branch, which at time of writing is 9fc1aeed25683042c53a872595e086bfa6a52e1f ?  And, if i do that, that will make it more or less impossible to merge just this change into the master branch, right?  

Again, forgive my ignorance, im new to GIT (only have a SVN background).

PR #2180, opened 2016-04-13
Some bug fixes

### LINKED ISSUE #2179: Bug Report About Bootstrap-table Filter control extension
I add the 'Clear Filters' button,and when I click the button,there is an error in the console.

image

djhvscf:
Example: https://jsfiddle.net/djhvscf/fp25opbb/1/


src/bootstrap-table.js
```diff
@@ -1956,7 +1956,7 @@
         }
         request = $.extend({}, calculateObjectValue(null, this.options.ajaxOptions), {
             type: this.options.method,
-            url: this.options.url,
+            url: (url === undefined ? this.options.url : url),
```
wenzhixin: url || this.options.url

djhvscf: :facepunch: yes, you're right.. :o: 

Let me change it

src/bootstrap-table.css
```diff
@@ -196,7 +196,7 @@
     line-height: 1.428571429;
 }
 
-.fixed-table-toolbar .bars,
+.fixed-table-toolbar .bars-bs-table,
```
wenzhixin: I think here is better to use `.bs-bars`, we use like `.bs-`... for example `.bs-checkbox` 
Do you think so? @djhvscf

wenzhixin:
Great, thanks @djhvscf

PR #2186, opened 2016-04-13
Upgrade to sort orders so the text can be changed from locales

ErwannNevou:
This change makes this possible : 

```js
$.fn.bootstrapTable.locales['fr-FR'] = {
    [ ... ]
    sortOrders: function () {
      return {
        asc: 'Croissant',
        desc: 'Décroissant'
      }
    },
    [ ... ]
 }
```

wich is useful :+1:

### DIFF
--- a/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
+++ b/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
@@ -200,7 +200,7 @@
         formatDuplicateAlertDescription: function() {
             return "Please remove or change any duplicate column.";
         },
-        sortOrders : function () {
+        formatSortOrders : function () {
             return {
               asc: 'Ascending',
               desc: 'Descending'
@@ -337,7 +337,7 @@
             $multiSortName.append('<option value="' + column.field + '">' + column.title + '</option>');
         });
 
-        $.each(this.options.sortOrders(), function(value, order) {
+        $.each(this.options.formatSortOrders(), function(value, order) {
             $multiSortOrder.append('<option value="' + value + '">' + order + '</option>');
         });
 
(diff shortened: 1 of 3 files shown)


djhvscf:
Can you provide an example?

ErwannNevou:
Hello,

Here is an example (All changes are in the multi-sort popup): 
First without my changes :  [No changes](https://jsfiddle.net/u2ta4utm/6/)
In this, we can't translate 'Ascending' and 'Descending' without changing the code in bootstrap-table-multiple-sort.js which is annoying. 

But I wanted to be able to translate it from the locales like the other strings in the popup.

Second with my changes : [With changes](https://jsfiddle.net/34tp1htj/3/) 
In this one, we can change the translation of 'Ascending' and 'Descending' without changing the code in bootstrap-table-multiple-sort.js. So we can change all the strings of the popup from the locales.

The multi sort still works, I just changed the translation 

PS : The translatons are in French

djhvscf:
Thanks! Lools good

src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
```diff
@@ -203,6 +199,12 @@
         },
         formatDuplicateAlertDescription: function() {
             return "Please remove or change any duplicate column.";
+        },
+        sortOrders : function () {
+            return {
+              asc: 'Ascending',
+              desc: 'Descending'
+            }
```
wenzhixin:
```
return {
    asc: 'Ascending',
    desc: 'Descending'
};
```

src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
```diff
@@ -203,6 +199,12 @@
         },
         formatDuplicateAlertDescription: function() {
             return "Please remove or change any duplicate column.";
+        },
+        sortOrders : function () {
```
wenzhixin: It's better to use `formatSortOrders`

wenzhixin:
Please help to update the README

ErwannNevou:
Done

wenzhixin:
Thanks!

PR #2216, opened 2016-04-24
+ OS X Finder internal system files

juliyvchirkov:
+ OS X Finder internal system files .DS_Store, .localized & Icon


wenzhixin:
Thanks!

Issue #2220, opened 2016-04-26
Exporting selected rows does not work when data-pagination-side is set to server


sampeka:
Experiencing a problem with the export plugin. The selected rows are not obeyed when data-side-pagination is set to server - here's a jsfiddle illustrating the issue: http://jsfiddle.net/e3nk137y/6074/. 

## Steps to reproduce
1. Select "Export Selected" in the dropdown menu at the top left.
2. Select some rows
3. Export as csv

david2sim:
I have the same problem

dabros:
So the core issue appears to be how server side pagination doesnt like `.load` of `getAllSelections`, in that I mean it seems to completely ignore it.

http://jsfiddle.net/dabros/axsuvcg1/

See `export` at https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/export/bootstrap-table-export.js#L99

See `getAllSelections` at https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L2578

See `load` at https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L2321

My fiddle shows current testing progress, and we can assume `load` works correctly with client side as seen in http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/export.html (breakpoints in source panel would be all thats needed to completely confirm that)

I might get round to this again later, but busy schedule so anyone else feel free to pursue further :)

sampeka:
Thanks @dabros, that was really helpful. If I get a chance tonight I'll try and put together a PR with your pointers, but feel free to beat me to it

sampeka:
@dabros do you have any ideas how to fix the "export all" function when the `data-side-pagination` is set to server?

dabros:
@sampeka - ive seen #2220, can you please provide a fiddle with changed file included in js panel (be sure to check the 'wrap' option) and all other conditions like `data-side-pagination` set for testing?

https://github.com/wenzhixin/bootstrap-table/blob/master/CONTRIBUTING.md#bug-reports

I can then take a look and ideally contain changes in js panel with `// CHANGED` comments so quicker testing process

heenaMahajan:
how to "export all" function when the data-side-pagination is set to server?it current export current page records

dabros:
@heenaMahajan - please be cautious posting to closed issues, if you still looking for a solution then create new issue and link it to here

PR #2232, opened 2016-04-28
Fix for #2220

sampeka:
Quick fix for #2220, but looks like overall this needs a bit of refactoring since it's a workaround. This only solves the problem with exporting selected data - exporting all still does not work. 

The problem this has solved is that `load` expects the data in the format I've written there when `data-side-pagination=true`. Ideally it would be best to avoid needing to check that condition since the data has already been standardised on the first call to `load`, but I thought for now it's best not to touch the code base that much when there are no unit tests etc. If there is a more permanent fix that doesn't require a lot of refactoring, let me know and I'll amend it.

### DIFF
--- a/src/extensions/export/bootstrap-table-export.js
+++ b/src/extensions/export/bootstrap-table-export.js
@@ -100,6 +100,15 @@
                         var data = that.getData(),
                             selectedData = that.getAllSelections();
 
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks!

PR #2237, opened 2016-04-29
Fix #1556: Properly set main entry in package.json

### LINKED ISSUE #1556: Add Common JS support
Any chance of adding Common JS support so this can be used with Webpack and Browserify?

herrcykel:
This makes it possible to use boostrap-table with webpack or browserify with a simple
```javascript
require('bootstrap-table');
```


wenzhixin:
Merged: https://github.com/wenzhixin/bootstrap-table/commit/a5cf1fe1261e4954672ed5579a627ab15594c53f

PR #2240, opened 2016-04-30
Fix #2239 - natural-sorting not work when data including numeric number, null or undefined.

### LINKED ISSUE #2239: natural-sorting not work when data including numeric number, null or undefined.
Hi.

The extension natural-sorting throw the following exception when the data including numeric number.
`bootstrap-table-natural-sorting.js:21 Uncaught TypeError: t.charAt is not a function`

And when the data including null or undefined, the plugin throws similar exception too.

Please confirm jsfiddle: 
https://jsfiddle.net/t02uk/dm8151ff/

Thank you.

t02uk:
Fix #2239. Here is jsfiddle.

Problem: https://jsfiddle.net/t02uk/dm8151ff/
Solution: https://jsfiddle.net/t02uk/o3Lvxvfn/

### DIFF
--- a/src/extensions/natural-sorting/bootstrap-table-natural-sorting.js
+++ b/src/extensions/natural-sorting/bootstrap-table-natural-sorting.js
@@ -29,8 +29,18 @@ function alphanum(a, b) {
     return tz;
   }
 
-  var aa = chunkify(a);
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks

PR #2255, opened 2016-05-07
Adding support for custom sorters – multi-sort (sortPriority)

manometer114:
If a custom sorter exists, use it.  Otherwise, use the default sorter function.  

Currently, having a custom sorter + multi-sort gets messed up since the default sorter is not apt for all cases.

This should make it so that sortPriority = [{"sortName":"DAY_ID","sortOrder":"desc"},{"sortName":"RESPONSES","sortOrder":"desc"}] works even when the DAY_ID and RESPONSES fields have custom sorters.

### DIFF
--- a/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
+++ b/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
@@ -204,7 +204,7 @@
             return {
               asc: 'Ascending',
               desc: 'Descending'
-            }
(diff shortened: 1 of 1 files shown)


djhvscf:
Can you provide an example, please?

wenzhixin:
Any update?

djhvscf:
Any update here, @manometer114?

djhvscf:

PR #2272, opened 2016-05-13
Adding sort class table option

### LINKED ISSUE #2265: Column Sort Class [solved]
Hi,

When you sort by a particular column, could you please add a class to the td elements of that column so that it can be styled to show that this is the column we are sorting by.

eg you can clearly see this table is being sorted by the third column.

Thanks

djhvscf:
In order to highlight the `td` elements sorted

Fix #2265 
Example: http://jsfiddle.net/djhvscf/e3nk137y/6452/


phibo23:
could you also update the data-sort-name and data-sort-order attributes on the table html so they can be used for styling the whole table or columns other than the sorted ones?

djhvscf:
@phibo23 I'm not sure what you mean, can you explain a little bit more?

@wenzhixin this **PR** needs something else?

wenzhixin:
Thanks, merged now, because I'm from the old to the new to handle the pull requests:)

wenzhixin:
Use `sprintf` to format the code: https://github.com/wenzhixin/bootstrap-table/commit/ce20422ac8f6d002ac3cc07fbc61939bb6816280

phibo23:
@djhvscf I mean the attributes that are used to pass the options to the table:
```
<table id="eventsTable"
       ...
       data-sort-name="stargazers_count"
       data-sort-order="desc"
       ...
>
```
They only represent the initial state of the table. But if they were updated on sort (which I currently do manually) they could be used to style columns other than the sorted one, e.g. if they are related to the sorted data, or other parts of the table such as the detail view.

PR #2304, opened 2016-05-25
added 'field' from description into parameter column

dabros:
added 'field' from description into parameter column

Mentioned in right column, but make more visible and consistent with other events doco

**IMPORTANT** 

@wenzhixin @djhvscf 

was looking at adding referance to when `this`could be used to access table object, as ive had to repeat it multiple times for various issues. 

Im just a little unsure of which or all events that applies to, as consistency is key for doco, and i really think pointing users to that for all relevant events can help them out alot (i certaintly didnt realise it till it was raised to me, makes absolute sense now why `$element` and such are param, but wouldnt have occured to me in advance, and other users making some mistake)


wenzhixin:
Thanks @dabros

PR #2330, opened 2016-05-31
After reordering columns, reinitialise the toolbar

iox:
This pull requests fixes a bug with the reorder columns extension.

Steps to reproduce:
- Open http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/reorder-columns.html
- Drag and drop column ID to the second position
- Click on the top right icon, to open the menu to hide/show columns
- Click the checkbox for "Item Name"
- Instead of hiding the "Item Name" column, the "ID" column is hidden

I found out that "resetView" did not update the toolbar menu. We need to run "initToolbar" too after a column reorder.

### DIFF
--- a/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
+++ b/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
@@ -144,6 +144,7 @@
                 for (var i = 0; i < ths.length; i++ ) {
                     columnIndex = $.fn.bootstrapTable.utils.getFieldIndex(that.columns, ths[i]);
                     if (columnIndex !== -1) {
+                        that.columns[columnIndex].fieldIndex = i;
                         columns.push(that.columns[columnIndex]);
                         that.columns.splice(columnIndex, 1);
                     }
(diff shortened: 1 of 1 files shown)


iox:
I close my own pull request, the issue is not solved yet. Now the headers and data are out of sync. I will investigate more and reopen the pull request when it's fixed.

iox:
Ok, I found where the problem came from: the "fieldIndex" attribute was not being updated, creating inconsistencies.

This pull request is now ready for review.

wenzhixin:
Thanks @iox

wenzhixin:
Can you provide a fiddle to show what you have fixed.

dhuneault:
Hi, 

Same Issue for me.

Try your link (explanation of ios) [http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/reorder-columns.html](url)

1. Drag and drop column ID to the second position
2. Click on the top right icon, to open the menu to hide/show columns
3. Click the checkbox for "Item Name"
4. Instead of hiding the "Item Name" column, the "ID" column is hidden and the "Item Name" column is still visible.

The problem is that the fieldIndex is not updated or reordered when dragging columns.

Thanks

wenzhixin:
Thanks for your contribution!

PR #2344, opened 2016-06-03
Update bootstrap-table.js

gdoumenc:
Trigger collapse-row before removing content


wenzhixin:
Thanks for your contribution!

PR #2078, opened 2016-03-13
Extension Multiple Sort Support Server Pagination

bleeloe:
Multisort extenson now support pagination server mode. added url array param 
`&multiSort[0][sortName]=field_one&multiSort[0][sortOrder]=asc&multiSort[1][sortName]=field_two&multiSort[1][sortOrder]=asc`

PR #2355, opened 2016-06-05
Prevent use of multiSort function after table loading when sorting server side.

owenmoony:
I'm binding to the `multiple-sort.bs.table` event to trigger `bootstrapTable('refresh')` but without this change it goes into an infinite loop of requests.


owenmoony:
Related to https://github.com/wenzhixin/bootstrap-table/pull/2078

wenzhixin:
Thanks for your contribution!

PR #2373, opened 2016-06-13
Decrement totalRows on remove if using serverside pagination

jekkos:
Bootstrap tables does not decrement the number of total rows when a row is removed with serverside pagination enabled. 

This causes an issue when someone removes all rows in the last page. In that case the calculated offset is invalid and will show an empty table as result. This fix decrements the total number of rows set by the server and will fix this behavior after a table refresh.


wenzhixin:
Thanks!

PR #2387, opened 2016-06-15
fixed page list

adambyer:
If the total number of records matches one of the numbers in the page list, the next next greater number in the page list is still displayed when it should not be. 

Example...
http://jsfiddle.net/e3nk137y/7102/

In this case there are 65 total records so the page list should be 10,25,50,65. But it also shows 100 which it shouldn't.


wenzhixin:
Thanks!

PR #2405, opened 2016-06-22
Fixed a bug in the calculation of toolbar and pagination heights

bezany:
Fixed a bug in the calculation of toolbar and pagination heights, if there is a inner div on the next line.
Demo bug: http://jsfiddle.net/xbyoLyxt/3/
Demo fix: http://jsfiddle.net/1dLt903z/1/

".fixed-table-toolbar", ".fixed-table-pagination" may contain two floating div. If inner div on the next line then function "getRealHeight" return incorrect value.
For get correct height with call outerHeight(true) need add clearfix solution (http://stackoverflow.com/a/1633170). Function "getRealHeight" is no longer used.

### DIFF
--- a/src/bootstrap-table.css
+++ b/src/bootstrap-table.css
@@ -304,3 +304,10 @@ div.fixed-table-scroll-outer {
     height: 150px;
     overflow: hidden;
 }
+
+/* for get correct heights  */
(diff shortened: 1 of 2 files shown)


wenzhixin:
Thanks!

PR #2430, opened 2016-07-05
Make 'selected' export properly using server pagination

yufan-liu:
When using server side pagination, and export method is 'selected',
the original method will download all the data in the page.
The new method will download only selected rows.
Because selected rows are wrapped in an array, it does not have 'rows' field.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2347,7 +2347,9 @@
         if (this.options.sidePagination === 'server') {
             this.options.totalRows = data.total;
             fixedScroll = data.fixedScroll;
-            data = data[this.options.dataField];
+            if (!$.isArray(data)) {
+                data = data[this.options.dataField];
+            }
         } else if (!$.isArray(data)) { // support fixedScroll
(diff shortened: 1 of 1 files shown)


wenzhixin:
It seems the same with #2220, merged the PR: #2232.

felipecarrillo100:
I had the same issue,  as a work around I did changes to the bootstrap-table-export.js
Badically I temporarilly disable the server side during the export.
Line 108:
                        var data = that.getData(),
                            selectedData = that.getAllSelections();
                        var backup_sidePagination = that.options.sidePagination;  //backup
                        that.options.sidePagination = 'client'; // set to client side before exporting
                        that.load(selectedData);
                        doExport();
                        that.options.sidePagination = backup_sidePagination;  // restore
                        that.load(data);

wenzhixin:
I think the solution of  @felipecarrillo100 is better than update `bootstrap-table.js`.

wenzhixin:
PR #2232 solved this issue. Thanks for your contribution!

PR #2460, opened 2016-07-13
Improvements in filter-control extension

### LINKED ISSUE #1540: filter-control => unset unnecessary values from select options
When we have many select filters, it would be great to update filter options and hide values which are not exists in current data view. 

I will better explain it on example: 
https://jsfiddle.net/e3nk137y/3345/

If we choose "First" : "A" than on filter "Second" option "3" shoud be disabled / deactivated .
Generally it would be great to update all filter options after each filtering action.

additional on stackoverflow: http://stackoverflow.com/questions/33101205

### LINKED ISSUE #2448: Filter control has embedded style
This is more of a request rather than a bug. The filter control from the extension has embedded style:
```
style="margin: 0 2px 2px 2px;"
```
This prevents us from overriding the style for margin using the class "filterControl". Without overriding it appears that the filter control has no spacing between it and the adjacent search field to the right. It would be easier for people to override if the style was in a CSS file instead of being embedded.

btw - the older version of the filter control did not have this embedded style.

### LINKED ISSUE #2461: Table with filterBy and updateByUniqueId: Cannot read property '_data' of undefined
I've a bootstrap table where i load data from URL. The load load Ok.
Into the table they are enable and disable useres. When the page is load i filter the data with "**filterBy**" to see enable user.
I've into the toolbox two button:

1. "**Activate Only**" To see enable users
2. "**Desactivate Only**" To see disable users 

The two button aply de funtion with "**filterBy**"

If the table have data-pagination="true" and i click to the button "**Desactivate Only**" and click to the row button into the column action, they do nothing, and will return this error:

> Cannot read property '_data' of undefined
> bootstrap-table.js:1626

You can see the issue here: http://jsfiddle.net/hcbg3k8g/11/

But, if the page don't have pagination (data-pagination="false"), the script work like a charme.
You can see the script working here: http://jsfiddle.net/hcbg3k8g/12/

I'm ussing Chrome

Thank you for your help.

djhvscf:
Fix #1540
Fix #2448 

Example for #1540: https://jsfiddle.net/djhvscf/e3nk137y/7580/

Fix #2461 
Examples
`data-pagination="false"`: http://jsfiddle.net/djhvscf/hcbg3k8g/14/
`data-pagination="true"`: http://jsfiddle.net/djhvscf/hcbg3k8g/13/

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -223,12 +223,14 @@
 
                 sortSelectControl(selectControl);
 
-                disableOptionSelectControl(selectControl, uniqueValues);
+                if (that.options.disableUnusedSelectOptions) {
+                    disableOptionSelectControl(selectControl, uniqueValues);
+                }
             }
         });
     };
 
-    var escapeID = function( id ) {
+    var escapeID = function(id) {
        return String(id).replace( /(:|\.|\[|\]|,)/g, "\\$1" );
    };
 
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -1,7 +1,7 @@
 /**
  * @author: Dennis Hernández
  * @webSite: http://djhvscf.github.io/Blog
- * @version: v2.1.2
+ * @version: v2.1.1
  */
 
 (function ($) {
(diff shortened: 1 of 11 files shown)


src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -1,7 +1,7 @@
 /**
  * @author: Dennis Hernández
  * @webSite: http://djhvscf.github.io/Blog
- * @version: v2.1.1
+ * @version: v2.1.2
```
wenzhixin: no need to update the version, because the release version is v2.1.0

wenzhixin:
https://github.com/wenzhixin/bootstrap-table/commit/c09d12cd009eeefa18f7a67cf0c2b357e37b8f2b

djhvscf:
@wenzhixin oh! You're right. Sorry. Let me change this

wenzhixin:
Thanks

PR #2463, opened 2016-07-14
Some improvements - Cookie / Filter-Control extensions

### LINKED ISSUE #2371: localStorage or sessionStorage table state saving.
Awesome work! I know there is the cookie extension, but it would be great if the table state could be remembered on localStorage or sessionStorage also, how can I archive this for active page, column selection, number of rows per page, filter control, sort name and sort order based on table-id? Would be great to choose session or local storage for saving. Million thanks for your help :)

djhvscf:
Example: https://jsfiddle.net/djhvscf/e3nk137y/7586/

Fix #2371

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ ChangeLog
 - feat(js): add `sortClass` option
 - feat(js): add `totalField` Option
 - fix(css): fix #2208, dropdown-menu style bug
+- fix(js): fix #2461, adding the initPagination call to updateByUniqueId and updateRow methods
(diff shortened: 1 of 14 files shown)


doomfrawen:
Hi,

I just downloaded this extension, and it works great but I found a small issue that I don't know how to fix. When I'm using the plug-in data-filter-control which filters specific columns (http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/filter-control.html), it does save the selected filter but it doesn't re-apply it, as if the filter would be null, therefore it shows back all the entries.

Would you happen to know how to fix that?

If you need any more information please let me know.

Thank you in advance

djhvscf:
@doomfrawen 

PR #2519, opened 2016-08-01
Side pagination on server => on change event and no key up

Flounn:
- on change event when sidePagination==='server' and no filter after load data. Decrease number of request and it's more user friendly


djhvscf:

@Flounn

PR #2521, opened 2016-08-01
locale fr : toolbar (advancedSearch) extension

Flounn:
external file locale fr-FR


djhvscf:
Hi @Flounn thanks for your contribution but this change must be done against locale files (https://github.com/wenzhixin/bootstrap-table/tree/develop/src/locale)

PR #2527, opened 2016-08-01
locale fr : export extension

Flounn:


djhvscf:
Hi @Flounn thanks for your contribution but this change must be done against locale files (https://github.com/wenzhixin/bootstrap-table/tree/develop/src/locale)

PR #2528, opened 2016-08-01
locale en-US to use other locale : export extension

Flounn:


wenzhixin:
I do not think it is necessary to set the locale, because the default locale is `en-US`.
For example: http://jsfiddle.net/e3nk137y/10681/

Flounn:
Yes sorry it's correct. I load the locale file before the tableExport.js file...

Issue #1049, opened 2015-06-24
XSS filtering on fields


alex-kolesnyk:
Please provide some possibility to escape special chars on fields.
There is not every API returns filtered values so this possibility will be very helpful.

djhvscf:
Can you provide a little example?

alex-kolesnyk:
Try to put "<script>alert()</script>" in to any field of bstable

alex-kolesnyk:
<script>alert()</script>

djhvscf:
Ok, do you have any idea how to avoid this in our plugin, @alex-kolesnyk ??

alex-kolesnyk:
Just check how Underscore escape method works, it's simple and helps a lot.

djhvscf:
Great! I'm checking that logic

toskebre:
Here you go:

var escapeObject = function(obj) {
    _.each(obj, function(value, key) {
      if (_.has(obj, key)) {
        if (_.isString(value)) {
          obj[key] = _.escape(value);
        } else if (_.isArray(value) || _.isObject(value)) {
          obj[key] = escapeObject(value);
        } else {
          obj[key] = value;
        }
      }
    });
    return obj;
  };

Then override initBody method and call escapeObject like this:

escapeObject(this.data);

After that, call default initBody inside override method.

djhvscf:
Ok.. in that case we have to copy the logic because we don't use the underscore library

wenzhixin:
Added `escape` option.

PR #2532, opened 2016-08-02
Add column level string escaping

jackjamieson:
https://github.com/wenzhixin/bootstrap-table/issues/1049

Issue 1049 was resolved by adding a table-wide escape option.  This pull request adds the ability to escape on the column level so certain columns can still have markup in them with the `data-escape` attribute on the `<th>` element.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -524,7 +524,8 @@
         cellStyle: undefined,
         searchable: true,
         searchFormatter: true,
-        cardVisible: true
+        cardVisible: true,
+        escape : false
     };
 
     BootstrapTable.EVENTS = {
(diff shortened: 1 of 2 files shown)


djhvscf:

jackjamieson:
https://jsfiddle.net/knd8vff1/ this shows the current table-escaping.
https://jsfiddle.net/2esvsz26/1/ this shows the proposed table-level escaping addition.

djhvscf:
Ok, got it... but please update the documentation

jackjamieson:
Documentation updated.

src/bootstrap-table.js
```diff
@@ -829,6 +830,10 @@
                     that.header.stateField = column.field;
                     that.options.singleSelect = true;
                 }
+                
+                if(column.escape) {
+                	that.options.escapeColumn = true;
```
djhvscf: What about this option `escapeColumn`?

jackjamieson: Woops, good catch that was from earlier while I was testing it out.  I will remove it.

jackjamieson:
Do I need to include an example within the documentation?

wenzhixin:
Thanks for your contribution!

PR #2534, opened 2016-08-03
add click to edit a row feature

horkenw:


horkenw:
I'm sorry just let me close this pull request, cause I do not think this is good enough to be a extension

PR #2563, opened 2016-08-10
Print extension - add printFormatter column option

yaronyam:
New column option 'printFormatter':  a function that formats the cell values for this column in the printed table. Function behaviour is similar to the 'formatter' column option


wenzhixin:
Thanks!

PR #2586, opened 2016-08-23
editable extension - Click edit row

horkenw:
update plugin logic to set input or select for row click to edit


wenzhixin:
Thanks for your contribution!

wenzhixin:
http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/click-edit-row.html
It seems cannot save the data when change a row @horkenw

Issue #2439, opened 2016-07-06
1.11 : filterBy with array not working


vricard:
Hi there,
Since the 1.10 to 1.11 maj, i can't filter with an array anymore : 
$('#table').bootstrapTable('filterBy', {id: [1,2,3]});
it only works with a single int
$('#table').bootstrapTable('filterBy', {id: 1});

thx for your work =3

wenzhixin:
Fixed in develop branch.

PR #2644, opened 2016-09-14
fixes #2643

### LINKED ISSUE #2643: filterBy regression - no longer supporting arrays
### `filterBy` no longer works with an array

```
$('#table').bootstrapTable("filterBy", {"forks_count": ["11","4"]});  // broken
```

See this example of it not working:

http://jsfiddle.net/thornomad/mqaybexx/
### code change at some point

It looks like the logic was changed at some point (can't find the specific commit).  I had originally submitted #1526 with a working solution but since then the code was refactored.  I am not sure if that was to fix something or just an attempt at cleanup.
### pull request submitted

http://jsfiddle.net/thornomad/9p854bm4/

thornomad:
Here is the jsfiddle showing it fixed: http://jsfiddle.net/thornomad/9p854bm4/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1242,9 +1242,12 @@
             // Check filter
             this.data = f ? $.grep(this.options.data, function (item, i) {
                 for (var key in f) {
-                    if ($.isArray(f[key]) && $.inArray(item[key], f[key]) === -1 ||
-                            !$.isArray(f[key]) && item[key] !== f[key]) {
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks! This bug is fixed in #2439

wenzhixin:
https://github.com/wenzhixin/bootstrap-table/commit/de867d379a46b377efa7eef83fdf898b9073b28c

PR #2646, opened 2016-09-15
fixes #2645

### LINKED ISSUE #2645: Removing inline javascript to avoid CSP voilation
The pagination component injects inline javascript **"javascript:void(0)"** in the anchor element to stop further navigation once clicked. Someone trying to implement [CSP](https://w3c.github.io/webappsec-csp/) and mitigate XSS issues by blocking inline javascript will receive though harmless but noisy CSP violations in the browser and on the server.

This is true for all browsers which honour CSP headers like Google Chrome 53. Setting http headers like following will results in an violation being reported in chrome developer console as well as an report being submitted on the server.

Header

```
Content-Security-Policy: script-src 'self'; report-to csp-reporting-endpoint
```

Will result on error like following in the Chrome developer console once the user click on page number to change the page for the table.

```
Refused to execute JavaScript URL because it violates the following Content Security Policy directive: "script-src 'self'. Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution.
```

Since this is an inline javascript execution it cannot be mitigated by using a sha256 or nonce value.

I will publish an PR and provide an fix as well.

mayankgoyal:
Enables setting CSP headers to block inline javascript to mitigate XSS attacks. Fix #2645

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1392,7 +1392,7 @@
                     } else {
                         active = page === that.options.pageSize ? ' class="active"' : '';
                     }
-                    pageNumber.push(sprintf('<li%s><a href="javascript:void(0)">%s</a></li>', active, page));
+                    pageNumber.push(sprintf('<li%s><a href="#">%s</a></li>', active, page));
                 }
             });
             pageNumber.push('</ul></span>');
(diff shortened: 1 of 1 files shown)


wenzhixin:
Merged in https://github.com/wenzhixin/bootstrap-table/commit/dd44699a965b1bb47797db9a0d2206d21c87cba8

PR #2679, opened 2016-09-29
Replace get field index by hash table

yenyen:
The getFieldIndex is called for every cell when an user do a search, with a big table, this can be very long.

This pull request replaces this function by an hashmap that is builds when the columns are initialized. The field is used as the key and the column index as the value. So you just need to use 
this.fieldsColumnsIndex[field] to retrieve the index. In complexity, this operation is O(1) instead of O(n).


djhvscf:
It looks good! @wenzhixin can we merge this?

wenzhixin:
Thanks for your contribution!

PR #2696, opened 2016-10-10
print extension - retain UI filtering and sort order

yaronyam:
Added boolean option - **printAsFilteredAndSortedOnUI**.  Default: _true_.
When true - print table as sorted and filtered on UI.


wenzhixin:
Thanks

PR #2717, opened 2016-10-18
Export extension: support .xlsx format

RJHsiao:
Plugin [tableExport.jquery.plugin](https://github.com/hhurz/tableExport.jquery.plugin) is already supported .xlsx format (Office Open XML Workbook, or someone called 'MS Excel 2007+ XML Format'), just modified for that.


wenzhixin:
Thanks

Issue #191, opened 2014-10-29
Remember Sort & Show-Columns with Cookie or Localstorage


nikmauro:
Hello how it possible to remember the state of sort, Show Columns Selections, Pagination and Records per page with Jquery Cookie or LocalStorage based on Table ID data-id-field="id"?

Suggestion params  data-state-save="true"

I have this Lib
https://github.com/carhartl/jquery-cookie

djhvscf:
Hey!! @wenzhixin you can close his issue!!

PR #2738, opened 2016-10-28
Bootstrap Table Defer URL Extension

rubensa:
When using server-side processing, the default mode of operation for
bootstrap-table is to simply throw away any data that currently exists
in the table and make a request to the server to get the first page of
data to display. This is fine for an empty table, but if you already
have the first page of data displayed in the plain HTML, it is a waste
of resources. As such, you can use data-defer-url instead of data-url to
allow you to instruct bootstrap-table to not make that initial request,
rather it will use the data already on the page.

### DIFF
--- a/src/extensions/defer-url/README.md
+++ b/src/extensions/defer-url/README.md
@@ -1,0 +1,17 @@
+# Table Defer URL
+
+Use Plugin: [bootstrap-table-defer-url](https://github.com/wenzhixin/bootstrap-table/tree/master/src/extensions/defer-url)
+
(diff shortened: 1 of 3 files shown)


djhvscf:
Can you share us an example, @rubensa?

rubensa:
There is already an example linked: [Bootstrap Table Defer URL Extension Example #191](https://github.com/wenzhixin/bootstrap-table-examples/issues/191)

wenzhixin:
Thanks for your contribution!

PR #2742, opened 2016-10-31
Update bootstrap-table.js

dei1562:
Added the column option showRadioTitle (boolean) the option which lets the developer decide to show the title of the columns with radio buttons. The default value is false.


wenzhixin:
http://jsfiddle.net/wenyi/e3nk137y/18060/
http://jsfiddle.net/wenyi/e3nk137y/18061/

PR #2758, opened 2016-11-09
Add method setOptions adn fix method "BootstrapTable.prototype.remove"

dimonchik-com:
For fix method append and prepend when option this.options.sidePagination === 'server'

When options.sidePagination === 'server' and we use method append or prepend not work pagination.

### DIFF
--- a/dist/bootstrap-table.js
+++ b/dist/bootstrap-table.js
@@ -2589,6 +2589,12 @@
         return this.options;
     };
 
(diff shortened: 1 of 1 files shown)


djhvscf:
Please provide a [jsFiddle](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports)

wenzhixin:
I think we can use `refreshOptions` instead, thanks for your contribution!

PR #2773, opened 2016-11-21
Allowing nested fields in group-by-v2 extensions

kanashin:


kanashin:
Not the package.json but I commited in order to use in my project

wenzhixin:
Please create A new PR, thanks.

PR #2777, opened 2016-11-22
Fix Filter Control overwrite "search" paramater

### LINKED ISSUE #2776: Filter Control overwrites "search" parameter in API call
With both "search" and Filter Control enabled, Filter Control will overwrite the "search" parameter sent to server/API.

Enable search and filterControl
Search for something (not using Filter Control)
Search for something using Filter Control

The last search will now populate both "filter" and "search" parameter sent to API/server. The original search is lost (but still populating the search input field).

Bakke:
Fixes #2776

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -630,15 +630,10 @@
             delete this.filterColumnsPartial[$field];
         }
 
-        // if the searchText is the same as the previously selected column value,
-        // bootstrapTable will not try searching again (even though the selected column
-        // may be different from the previous search).  As a work around
-        // we're manually appending some text to bootrap's searchText field
-        // to guarantee that it will perform a search again when we call this.onSearch(event)
-        this.searchText += "randomText";
-
         this.options.pageNumber = 1;
-        this.onSearch(event);
+        this.initSearch();
+        this.updatePagination();
+        this.trigger('search', text);
         this.trigger('column-search', $field, text);
     };
 


djhvscf:
@Bakke do you have any fiddle where we can see the issue and the solution?, thanks.

Bakke:
https://jsfiddle.net/p63sfsvd/5/

Sorry, I have no working API examples right now using HTTPS.

If you search using the "search"-field you get a request like: `https://jsonplaceholder.typicode.com/posts?search=test1&order=asc&offset=0&limit=10`. The "search" parameter is now "test1".

If you then filter a column you get a new request like: `https://jsonplaceholder.typicode.com/posts?search=test2&order=asc&offset=0&limit=10&filter=%7B%22title%22%3A%22test2%22%7D`. The "search" parameter is now "test2" (the same as filter) - the original value is lost.

Bakke:
Working example: https://jsfiddle.net/qbkn7991/10/

If you search for "luke", then filter "Gender"-column to "male" the request will be: `https://swapi.co/api/people/?search=luke&order=asc&offset=0&limit=10&filter=%7B%22gender%22%3A%22male%22%7D`

wenzhixin:
@djhvscf help to review this please.

DustinUtecht:
@djhvscf any update here ?

diemus:
thanks for the solution@Bakke ,I have the same problem,it seems the newest code didn't sovle this problem,

mciesiel:
Thanks for the solution, I have the problem too
As diemus said, the fix is not in the last version, can you resolve conflicts and create a new version? 
Thank you in advance

djhvscf:

Issue #2791, opened 2016-11-26
Accessibility issues while working with bootstrap-table 


eduardomourar:
In order to make the table easier to use by people with disabilities, some improvements are required. I will be sending a PR with those changes soon that will include:

- Use ARIA components
- Make it possible for a screen reader (NVDA, JAWS, etc) to be used

wenzhixin:
Merged, thanks.

PR #2792, opened 2016-11-26
Fix accessibility issues (#2791)

eduardomourar:
See issue #2791 for details.
- Fix accessibility issues on table, table-export, table-toolbar


wenzhixin:
Thanks

PR #2831, opened 2016-12-16
insertRow method doesn't works if the table has active filters

jps1984:
The insertRow method doesn't works if the table has active filters, even when the new row matches the enabled filter.
The issue can be reproduced here:
[https://jsfiddle.net/jps1984/0v3enkwe](https://jsfiddle.net/jps1984/0v3enkwe)

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2511,7 +2511,7 @@
         if (!params.hasOwnProperty('index') || !params.hasOwnProperty('row')) {
             return;
         }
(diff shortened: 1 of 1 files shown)


gboyreau:
Fixed.
In InsertRow method, you have to replace : 
"this.data.splice(params.index, 0, params.row);"
By
"this.options.data.splice(params.index, 0, params.row);"
just like UpdateRow or UpdateByUniq

djhvscf:
Great!

PR #2838, opened 2016-12-26
Master

klausthul:
Bugfix to the filter-control plugin.

It fixes following behavior: When there is a select control in one column and other columns get added/removed from the table, the items shown in the select control get duplicated.


klausthul:
Sorry did not work...

PR #2855, opened 2017-01-09
fixed racecondition

Dunstkreis:
When filling in table-search-input-fields, i allways get missing letters while typing. Thats because of the line edited in this change.
You dont' wanna "resett" the field value. in the case that it is a select field, you wanna set the value, thats okay. But not on an input.
setting the value triggers change, so we'll trigger this manually.


djhvscf:
Please provide a [jsFiddle](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports) to show this problem, thanks!

djhvscf:

PR #2882, opened 2017-02-03
Update boostrap-table.js

pokica-hu:
adding updateCellById method. to be called with {id, field, value}

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2627,6 +2627,33 @@
         this.initBody(true);
     };
 
+    BootstrapTable.prototype.updateCellById = function (params) {
+        var that = this;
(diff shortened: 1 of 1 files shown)


djhvscf:
Can you create an example with this change?

pokica-hu:
I use the new method with this function:
function updateTableCell(tableObject, id, cellName, newData){
    tableObject.bootstrapTable('updateCellById', {id: id, field: cellName, value: newData});
}

to documentation: 
Name: updateCellByID
Parameter: params
Description: update the cell specified by the id, each param contains following properties:
  id: a row id where the id should be the uniqueid field assigned to the table.
  field: field name of the cell to be updated
  value: new value of the cell

wenzhixin:
Can you help to update the documentation?

Issue #2887, opened 2017-02-08
Change data-url using refresh:url


Shpartko:
Hello!

When init table with property `data-url = "/ admin / category / load category = 10?"`, And after that change url `table.bootstrapTable ( 'refresh', {url: '/ admin / category / load category = 99?' })`, all be ok, data loaded correctly. But, if I use pagination and change page from 1 to 2, table use 1st url - _category = 10_, but not _category = 99_ ...

I think, this is not correct. Table must use last assigned url for loading data.
I don't know how I can change data url correctly for all table features work after that.

What you think to do that?

djhvscf:
Please provide a [jsFiddle](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports) to show your problem, thanks!

Shpartko:
The problem same at this: https://github.com/wenzhixin/bootstrap-table/issues/2481
Infographic: http://joxi.ru/nAyQ4G4hPgBEAZ

Solve:
   ```
 BootstrapTable.prototype.refresh = function (params) {
        if (params && params.url) {
            this.options.url = params.url;
            this.options.pageNumber = 1;
        }

        this.initServer(params && params.silent,
            params && params.query, params && params.url);
        this.trigger('refresh', params);
    };
```
If that correct architecture code, implement it in bootstrap-table master code.

Shpartko:
I don't know, do you see closed issues or not, so I reopen this issue for @djhvscf take in solve of descriptive problem and may be implement in master branch.

djhvscf:
@Shpartko lets do this, please:

1- Share us an example with your solution
2- Create a new Pull Request

I think it also can solve #2481

Shpartko:
Example of solution: https://jsfiddle.net/jamiam/5qmm063j/
My Pull Request: a91e050ab12a3c8408cdab5e33c508977cc497c1

wenzhixin:
Merged

PR #2891, opened 2017-02-09
Change data-url using refresh:url

Shpartko:
https://github.com/wenzhixin/bootstrap-table/issues/2887

Example of solution: https://jsfiddle.net/jamiam/5qmm063j/

+New feature for refresh method: set pageNumber


djhvscf:
@Shpartko I think we need to update the documentation, please

Shpartko:
Done.

wenzhixin:
Thanks @Shpartko

PR #2893, opened 2017-02-10
add align option for mergeCells

n4mlih:
no option for align so i made align option available for merging cells. When aplied to 'post-body.bs.table'. i can have mergeCells with differents align from columns header style i made..

there you can see just for last row i made align center + mergeCells it

hopefully this will usefull, thanks.

### DIFF
--- a/dist/bootstrap-table.js
+++ b/dist/bootstrap-table.js
@@ -2547,6 +2547,7 @@
             col = $.inArray(options.field, this.getVisibleFields()),
             rowspan = options.rowspan || 1,
             colspan = options.colspan || 1,
+            align = options.align,
             i, j,
             $tr = this.$body.find('>tr'),
             $td;
@@ -2567,7 +2568,7 @@
             }
         }
 
-        $td.attr('rowspan', rowspan).attr('colspan', colspan).show();
+        $td.attr('rowspan', rowspan).attr('colspan', colspan).attr('align', align).show();
     };
 
     BootstrapTable.prototype.updateCell = function (params) {


martinburch:
The HTML table cell align attribute is deprecated in HTML4.01 and removed in HTML5.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#attr-align

@n4mlih n4mlih Does using bootstrap-table's existing cell style property instead, and passing a CSS property, work for you?
```
function cellStyle(value, row, index, field) {
  return {
    css: {"text-align": "center"}
  };
}
```

n4mlih:
wow...that align already deprecated but why i can still use it??

when i do what your exactly suggest, the style attribute doesnt add to my table.. here i try to produce it in https://jsfiddle.net/9jeh3Lqk/
what i missing here??..help

thanks @martinburch

martinburch:
@n4mlih it's just the fiddle that's messed up—something to do with how the javascript is loaded and what methods are in-scope.

Changing the "load type" of the javascript to `No wrap - in <body>` in the fiddle fixed it.
image

https://jsfiddle.net/9jeh3Lqk/3/

You can also set the cellStyle formatter in the javascript instead of reading it out of the HTML attributes. Something like this:
```
$(function () {
    $('#table').bootstrapTable({
        data: data,
        columns: [{},{},{},{cellStyle:cellStyle},{cellStyle:cellStyle},{cellStyle:cellStyle},{cellStyle:cellStyle}]
    });
});
```

wenzhixin:
I think @martinburch is right, thanks anyway.

PR #2928, opened 2017-02-21
bug fixed: value write back

horkenw:
when submitted, changed value rewrite back to list

#2586


wenzhixin:
When the row field is not `name` or `price`, can it work?

horkenw:
sorry seems I make some wrong I will fixed it again.. :P I'm close this branch

djhvscf:
I think the fields of the row can't be hard coded.. They should be dynamic

@wenzhixin @horkenw

wenzhixin:
Yes, your are right @djhvscf

horkenw:
Ok, I will try to do this in another way, thanks. :>  @wenzhixin @djhvscf

PR #2938, opened 2017-02-22
Fix issue with Placeholder options

djhvscf:
when it is undefined the filter-control plugin is not rendering the control.

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -257,7 +257,7 @@
                 if (column.searchable && that.options.filterTemplate[nameControl]) {
                     addedFilterControl = true;
                     isVisible = 'visible';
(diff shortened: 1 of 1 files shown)


nicoden:
The root cause is due to COLUMN_DEFAULTS not taken into account. So the default `filterControlPlaceholder` option is undefined instead empty string.

Original code:
``` js
    $.extend($.fn.bootstrapTable.COLUMN_DEFAULTS, {
        filterControl: undefined,
        filterData: undefined,
        filterDatepickerOptions: undefined,
        filterStrictSearch: false,
        filterStartsWithSearch: false,
        filterControlPlaceholder: ""
    });
```

`$.fn.bootstrapTable.COLUMN_DEFAULTS` 
must be
`$.fn.bootstrapTable.Constructor.COLUMN_DEFAULTS`
or 
`$.fn.bootstrapTable.columnDefaults`

PR #2941, opened 2017-02-23
Feature - Support enable/disable expanding by row

ibraheemhlaiyil:
There is no current support for enabling/disabling expansion of a row on a per-row basis.

This PR adds support for this.

Option `data-detail-filter="detailFilterFunction"` or `options.detailFilter=detailFilterFunction` is used to provide a function name (or function) to call. This function is provided with the row index and row data as parameters:

`function detailFilterFunction(index, row) {
}`

The default detailFilter option function returns true to keep the existing functionality to enable expanding all rows if this option is not configured.

See fiddle for example: http://jsfiddle.net/ng0uuqy2/4/

### DIFF
--- a/docs/_i18n/en/documentation/table-options.md
+++ b/docs/_i18n/en/documentation/table-options.md
@@ -448,6 +448,13 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
         <td>Format your detail view when <code>detailView</code> is set to <code>true</code>. Return a String and it will be appended into the detail view cell, optionally render the element directly using the third parameter which is a jQuery element of the target cell.</td>
     </tr>
     <tr>
+        <td>detailFilter</td>
+        <td>data-detail-filter</td>
+        <td>Function</td>
+        <td>function(index, row) {<br>return true;<br>}</td>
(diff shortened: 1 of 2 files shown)


djhvscf:
@ibraheemhlaiyil can you please update also the documentation?

Thanks

ibraheemhlaiyil:
@djhvscf I've updated table-options.md

wenzhixin:
Thanks for your contribution! I think this is an useful feature.

pjabbott:
This feature made it into the documentation on http://bootstrap-table.wenzhixin.net.cn/documentation/, but it's only present in /src/bootstrap-table.js in the develop branch.  It is not in the zip file at http://bootstrap-table.wenzhixin.net.cn/getting-started, nor is it in /dist in either the develop or master branches.

Was this added to the documentation prematurely?  I was pulling my hair out trying to get it to work before I realized 'detailFilter' was not present at all in the /dist I downloaded.

ibraheemhlaiyil:
I was asked to add in the documentation but I think the change wasn't pulled in! Glad you find my feature useful :)

PR #2980, opened 2017-03-09
Add missing Slovak translations

D0d0:
Add missing Slovak translations into translate file


wenzhixin:
Thanks for your contribution!

PR #2987, opened 2017-03-10
Fix grouped tables with nested files

kanashin:
The modification of bootstrap-table.js is because of height attribute.


moebaca:
I incorporated these changes on my local copy and it fixed the following error:

return a[that.options.groupByField].localeCompare(b[that.options.groupByField]);

PR #3010, opened 2017-03-17
Fix loading of defaults columns options in table-filter-control

nicoden:
Defaults columns options are already defined in the code but never set.


wenzhixin:
Thanks for your contribution!

Issue #3012, opened 2017-03-17
sortStable is not stable


nicoden:
The stable sort do not maintain the relative sort correctly.

Example: [https://jsfiddle.net/nicoden/qotyfmne/6/](https://jsfiddle.net/nicoden/qotyfmne/6/)

Step to reproduce:

  1. Sort by Category and see order of rows
  2. Toggle sort of Category and see that rows order is changed

Because Category is the same for all rows, the rows order should not change with a stable algorithm.

wenzhixin:
Merged!

PR #3014, opened 2017-03-17
Fix stable sort (issue #3012)

nicoden:
Stable sort is not working as expected:

- Always set _position before sort (to keep it stable)
- Fix stable sort for 'calculatedObject' when they are equals


wenzhixin:
Thanks for your contribution!

Issue #3032, opened 2017-03-23
Cookie extension breaks paging post merging PR #2646


mayankgoyal:
After merging PR  #2646, using cookie extension breaks page navigation via the pager as it encapsulates the page navigation functions onPage* and ignores the returned values from it. We need to just propagate the returns to jQuery for it to do the right thing.  

I will publish an PR and provide an fix as well.

djhvscf:
Do you have any example??

mayankgoyal:
Withdrawing as this has incorrect usage at my end and I finally figured out what I did incorrect.

PR #3034, opened 2017-03-23
Fixing navigation reset while using cookie extension on navigation

mayankgoyal:
Fixes the Issues #3032


djhvscf:

mayankgoyal:
I think the issue I had was with using it in incorrect sequence with browserify. This was an incorrect request.

Issue #3046, opened 2017-03-27
No way to set footer values when server-side pagination is used


mrbubblesort:
I would like to set the values in the footer with an object in the ajax response when data-url & data-side-pagination="server" is set.  (Specifically, I'm loading data from an elasticsearch instance, and want to show my aggregations in the footer.)  Is there a way to do that?  For example, if this is the response:

`
{
    "total": 200,
    "rows": [ ['col1':50, 'col2':50] .... ],
    "aggs": [ 'col1' : 100, 'col2' : 200 ]
}
`

How can I set the "aggs" field in the footer?  Or is there a better way to do it?

If there isn't a way, I've made a small change that might work:
https://github.com/wenzhixin/bootstrap-table/compare/develop...mrbubblesort:develop

mrbubblesort:
Made pull request
https://github.com/wenzhixin/bootstrap-table/pull/3048

mnswiftone:
I found a solution to adding total values to the footer row of my server-sided pagination true table using existing 1.14 with no modifications needed.  The way it works is all data values returned are available in the load-success event - and if you use the footerFormatter - you can assign id'd containers (I used spans) to put the values from the dataset in the load-success event.  I have used this with my results over 16000 rows as a lot of pages, with the total row showing all column totals (returned from stored proc assigned in API returned dataset) - and not the current page only totals.

I believe this could be a solution for both #3048 and #3046

Steps:

1.  Add a footerformatter to the column(s) you wish to have a total value. i.e.
```
columns: [
 ...(other columns)...,
{
     title: 'Order<br />Amount', 
     field: 'Order_Amount', 
     sortable: true, 
     align: 'center', 
     footerFormatter: 'totalFormatter'
}]
```
2. Add footerFormatter's for each column - each having a unique html container id for your footer column (I added class text-nowrap to ensure the $ was in front of the amount, otherwise it showed as 'Total $' and amount on the next line) :
```
function totalFormatter(data, row) {
     return 'Total: <span id="totalspan" class="text-nowrap"></span>';
}
```

3. Add the value for that column as part of your API return data set (one for each column you want to show a total for) i.e. My example shows one  column, I did this for many columns, but all same concept just unique ids in footerFormatter. Here I added a totalamount and totalcount to show total $ of orders, and how many orders placed from stored proc and assigned those to return values in the API, returned in the dataset like:
```
{
     rows: ...   ,
     total: 600,
     totalamount: 5432.10,
[...]

PR #3048, opened 2017-03-27
Allow custom footer values from server

mrbubblesort:
Also mentioned here https://github.com/wenzhixin/bootstrap-table/issues/3046

Assuming this is the current expected response when using server side pagination
https://github.com/wenzhixin/bootstrap-table-examples/blob/master/json/data2.json
This change will allow support support for a third field defined by this.options.footerValuesField that will be used for the footer.  

For example, if you need to display the "true" total of _all records on all pages_ in the footer, you would need to make additional ajax calls to set that, since currently the only data sent to the footer is what already exists in the table.

So if the server returns a response like so:

```
{ 
"total": 200, 
"rows": [ 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25], 
    ['col1':50, 'col2':25]
   ], 
"aggregations": ['col1' : 750000000, 'col2' : 64500000] 
}
```

With this change, you can make the values in the "aggregations" field display in the footer like so:

```
$('#my_table').bootstrapTable({
   ....
   footerValuesField: 'aggregations',
   ....
});
```

Example:

http://jsfiddle.net/4r6g4cfu/626/

There are supposed to be 800 rows of data and in this example I want to show the total of all rows (which is 319600). However that is impossible since only the values for one page have been retrieved from the server. The total for all rows would have to be calculated server-side, hence the need to check for another field from the server.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -298,6 +298,8 @@
         paginationLoop: true,
         sidePagination: 'client', // client or server
         totalRows: 0, // server side need to set
+        footerValues: [],
+        footerValuesField: null,
         pageNumber: 1,
         pageSize: 10,
         pageList: [10, 25, 50, 100],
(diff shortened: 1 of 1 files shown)


djhvscf:

mrbubblesort:
Take a look at this then
http://jsfiddle.net/4r6g4cfu/626/

There are supposed to be 800 rows of data and in this example I want to show the total of _all_ rows (which is 319600).  However that is impossible since only the values for one page have been retrieved from the server.  The total for all rows would have to be calculated server-side, hence the need to check for another field from the server.

mrbubblesort:
@djhvscf 

I've added an example, can you please check it?

mrbubblesort:
@djhvscf

Can you please give me an update on the example?  Thanks!

djhvscf:
@wenzhixin

mrbubblesort:
@djhvscf 
@wenzhixin

Still nothing?

PR #3061, opened 2017-03-31
Keep timeout per column, not globally

### LINKED ISSUE #3063: Using datepicker filter-control on 2 identical tables causes both tables' datepicker field to be set
When putting two tables with identical column sets (but different data), both with filter-control, on the same page, then selecting a date using the datepicker from 1 table's filter, the date on the other table's filter is also set. The filter is not applied though.

https://jsfiddle.net/0yjj5r91/

This is caused by the fix for #2418, where the set of classes is used to select the field to be set.

jvdmr:
First commit fixes issue where leaving a filter field too quickly (by using tab to move to the next field) would ignore the value just entered.
Second commit fixes #3063

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1004,22 +1004,28 @@
         }
     };
 
-    BootstrapTable.prototype.onSort = function (event) {
+    BootstrapTable.prototype.onSort = function (event, reSort) {
         var $this = event.type === "keypress" ? $(event.currentTarget) : $(event.currentTarget).parent(),
             $this_ = this.$header.find('th').eq($this.index());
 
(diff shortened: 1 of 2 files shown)


djhvscf:
Do you have any example with this fix?

jvdmr:
Sure: In the below jsfiddles, click the first filter field, then type 'An<tab>' quickly.
Broken - https://jsfiddle.net/L31grfa4/1/
Fixed - https://jsfiddle.net/z3xhenj9/1/

I've also added a Fixed example on #3063

jvdmr:
Closing pull request as my branch structure was messing it up; will open a new clean PR

PR #3074, opened 2017-04-04
Filter and sort together

### LINKED ISSUE #3073: After applying filter, sort is undone but still indicated
Using filter-control, when adding a filter, then sorting the column, sort is applied as expected.
After this, editing filter (or removing altogether) shows new filtered (or unfiltered) data, unsorted, although column header still shows sorting caret.

See jsfiddle: https://jsfiddle.net/ojhcmzLy/1/
Enter 'for' in first column filter -> only 2 entries
Sort ascending -> first entry "Redmine ...", second entry "Show/hide ..."
Make first column filter empty -> all data shown, unsorted, but column header still shows ascending caret

jvdmr:
Fixes #3073 
Fixed jsfiddle: https://jsfiddle.net/ojhcmzLy/3/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1004,22 +1004,28 @@
         }
     };
 
-    BootstrapTable.prototype.onSort = function (event) {
+    BootstrapTable.prototype.onSort = function (event, reSort) {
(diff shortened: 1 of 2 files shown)


djhvscf:
This needs to fix the conflicts. Please open a new PR

PR #3076, opened 2017-04-04
Fix #3063 - create id on datepicker fields, use it to set value

### LINKED ISSUE #3063: Using datepicker filter-control on 2 identical tables causes both tables' datepicker field to be set

jvdmr:
Fixes #3063

Broken: https://jsfiddle.net/0yjj5r91/
Fixed: https://jsfiddle.net/eo9tdbgg/3/


wenzhixin:
Thanks for your contribution!

PR #3077, opened 2017-04-04
keep timeout per column, not globally

jvdmr:
Fixes issue where leaving a filter field too quickly (by using tab to move to the next field) would ignore the value just entered.

In the below jsfiddles, click the first filter field, then type 'An<tab>' quickly.
Broken - https://jsfiddle.net/L31grfa4/1/
Fixed - https://jsfiddle.net/z3xhenj9/2/


wenzhixin:
Thanks!

PR #3085, opened 2017-04-17
UpdateRow/updateByUniqueId extend or replace data.

andig89:
UpdateRow/updateByUniqueId use $.extend which only extend data (not remove attributes), so I add "replace" option.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2542,7 +2542,11 @@
             if (rowId === -1) {
                 return;
             }
-            $.extend(that.options.data[rowId], params.row);
+            if (params.hasOwnProperty('replace') && params.replace) {
+                that.options.data[rowId] = params.row;
+            } else {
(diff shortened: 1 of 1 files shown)


wenzhixin:
Can you help to add documentation?

andig89:
Hi,
yes but i don't have much time in this month, so i will try write new in next on my version. I see a lot of function to optimize/rewrite/modify so maybe i will be able to do that on writing doc 😀.

---
Regards Dariusz Gardyński

thornomad:
Hi @wenzhixin - I can use this update as well !  I just ran into this problem myself when a value returned by the server was changed to `undefined` -- and I couldn't figure out why `updateRow` wasn't updating the row!  That's, of course, because of `$.extend()`. 

Anyhow - what do you need to be able to merge this pull request I can update the documentation.

djhvscf:
This needs to fix the conflicts. Please open a new PR

PR #3120, opened 2017-05-03
Issue with reorder and master detail 

rpannell:
I found an issue with an undefined column when trying to reorder when the detail-view is enabled.

I had to remove undefined column when getting the list of field names from the th html records


djhvscf:
Please reopen against develop branch

PR #3152, opened 2017-05-24
repair bug

ackamen:
here this.options.customSearch  is string not a function , can't call apply


chrissharp80:
window[this.options.customSearch] is undefined. Can't call apply. My entire app is initialized, the table works, the details rows work, formatter functions work, and the basic search works. However, when I set `'customSearch: 'customSearch'` I hit the line of code altered in this issue and I get the error mentioned above. `this.options.customSearch` does contain my function, it's just not attached to the window object.

Issue #2582, opened 2016-08-19
When using the cookie plugin, the first ajax call isn't fired


erichelgeson:
When using the `cookie` extension it looks like the overridden `_initServer` in the extension returns too early and never tries to call the url here https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/cookie/bootstrap-table-cookie.js#L302-L305 I'm not sure what the logic is doing here, hence no patch :)

Example:
https://jsfiddle.net/spfgx18g/1/

When you load this you see the url is never called. If you comment out the cooke and re-run it will call the URL.

This also presents itself with serverside pagination, the first call is not sent but sorting will then send the ajax call.

May be related to https://github.com/wenzhixin/bootstrap-table/issues/2483 as well

RadekVala:
I had the same issue. But the example [here](http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/cookie.html) with similar scenario work well. I try to update the bootstrap-table plugin to version 1.11.0, it did not helped. Then I realized that after node update, the extension was still in v1.2.1, but examples are using v1.2.2. After manual update of bootstrap-table-cookie.js to v1.2.2. it is working also with server data.

erichelgeson:
It looks like 1.2.2 works, thanks. This is not currently in the release, but on the development branch.

vingorius:
Tks @RadekVala & @erichelgeson .

vingorius:
@wenzhixin , Could you update this module.

pandomic:
It seems that it is js's typecast issue.
In Cookie.js we have:
```js
cookiesPresent = function() {
    return bootstrapTable.options.cookie && bootstrapTable.getCookies(bootstrapTable);
};
// ...
if (cookiesPresent() && selectsWithoutDefaults.length === 0) {
    return;
}
```
If you try in console `false && []` you will got expected `false`, but unfortunately if you try `true && []` you'll got `Array` instead of `Boolean`, that;s why `return;` will be executed even if cookies are not presented.

So solution is pretty simple, we need just to change
```js
return bootstrapTable.options.cookie && bootstrapTable.getCookies(bootstrapTable);
```
to 
```js
return bootstrapTable.options.cookie && bootstrapTable.getCookies(bootstrapTable).length;
```

pasztorpisti:
### Reproducing the bug

http://jsfiddle.net/pasztorpisti/yvr6u93s/

[...]

PR #3184, opened 2017-06-09
Fix missing first ajax request with filter cookies

pasztorpisti:
This is a fix to #2582. The description and the repro of the bug: https://github.com/wenzhixin/bootstrap-table/issues/2582#issuecomment-307264950

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -170,7 +170,6 @@
             var parsedCookieFilters = JSON.parse(getCookie(bootstrapTable, bootstrapTable.options.cookieIdTable, cookieIds.filterControl));
 
             if (!bootstrapTable.options.filterControlValuesLoaded && parsedCookieFilters) {
-                bootstrapTable.options.filterControlValuesLoaded = true;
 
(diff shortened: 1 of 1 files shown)


djhvscf:
This PR has conflict.. please resolve them

src/extensions/cookie/bootstrap-table-cookie.js
```diff
@@ -193,7 +192,8 @@
                 });
 
                 bootstrapTable.initColumnSearch(cachedFilters);
-                bootstrapTable.refresh();
+                bootstrapTable.options.filterControlValuesLoaded = true;
+                bootstrapTable.initServer();
```
pasztorpisti: I've kept my implementation which calls the internal `initServer` instead of the public `refresh` (which triggers a 'refresh' event). The reason is that this part of the code is simply a "delayed init" and not really a refresh. At the init stage in which this code is executed the bootstrap table would normally execute `initServer` without this cookie plugin.

Note that the user experience could be improved by cutting down the length of the init delay from 250ms (see below) to a much smaller number but I don't know why it is such a large number. (Perhaps to avoid interference with other plugins that are also using delays for their init?)

PR #3195, opened 2017-06-12
when use bootstrap extensions group, the head width is wrong!

happyshi0402:
when use bootstrap extensions group, the head width is wrong!


djhvscf:
Can you share us an example with the issue and another with the fix?

wenzhixin:
Fixed by https://github.com/wenzhixin/bootstrap-table/commit/2abf4bc61b0f8309be7ed4919a83083efb52b03e
ref #3472

PR #3202, opened 2017-06-14
Provide typings - reopened after adding missing file

jregistr:
Provide simple typings to that this library out of the box is ready for typescript.
There is though a typings out there for this library under the definitelyTyped umbrella,
but it is extremely outdated and isn't correct. For instance, ts compiler will give an error for the line
$('#someId').bootstrapTable('method', {});
This pull requests adds a simple but ready type definition.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,5 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
-  "types": "./index.d.ts",
-  "dependencies": {
-    "@types/jquery": "^2.0.46"
-  }
+  "types": "./index.d.ts"
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,7 @@
+interface JQuery {
+
+    bootstrapTable(options: object): JQuery;
+
+    bootstrapTable(method: string, ...parameters: any[]): JQuery | any;
+
+}


djhvscf:
Not sure about this **PR** can you provide examples? or something where we can test it?

jregistr:
Is it okay if I make a project in another repository and link it here as an example?

djhvscf:
But what is the purpose of it?

jregistr:
It's so the typescript compiler will let you compile your code. The definition file for Jquery doesn't have a `.bootstrapTable` function by default. So when you do something like
`$('#id').bootstrapTable(...)`, it fails to compile. This adds a default definition that tells the ts compiler that such a function does exist.
As stated above, there is currently a types def provided here
https://www.npmjs.com/package/@types/bootstrap-table
but it's not up to date because ts compiler gave me an error for using a method. The line
`$('#id').bootstrapTable('append', {})` produced an error.
I made a custom definition file so that I could finish my project, figured I'd add it to this repository for other users.

jregistr:
Check out this repository for an example.
[https://github.com/jregistr/bootstrap-table-typings-examples](https://github.com/jregistr/bootstrap-table-typings-examples)
The success folder contains a project where I added the declaration file to define the bootstrap-table function call.

jregistr:
Here's an example library that provides its own declarations in the moment.d.ts file.
https://github.com/moment/moment

djhvscf:
@wenzhixin

package.json
```diff
@@ -40,5 +40,9 @@
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
+  },
+  "types": "./index.d.ts",
+  "dependencies": {
+    "@types/jquery": "^2.0.46"
```
wenzhixin: @jregistr I want to ask if this `dependencies` is necessary? I don't use typescript and use `npm` to install `bootstrap-table` will download this dependencies every time?

wenzhixin:
I think you can add the dependencies in your own project.

Issue #142, opened 2014-10-11
data-formatter search


shadow431:
After using "data-formatter" on a column to change the timezone of the Date/Time displayed, searching doesn't use new values, it uses original data.

Ex:
original: 2014-09-04 22:29:56
formated: 2014-09-04 03:29:56 pm

When searching "22" the row shows, but when searching "03", the visible hour, the table returns blank

mgitis:
I have been looking to see where it was possible to search formatted data columns and I found this post. I see where it is possible to search a column where the data-field is set to a known field from the binding json data source, hence if there is a field called "field1" in the binding json data. then once that field is set when creating the table header it can be searched. Now suppose your binding json data was complex and all data would have to be formatted as there would be the need to have multiple custom fields; that is, the table will consist of data that is a combination of data within the binding json data forming logically a new field. For example, in the json binding object you have "firstname" and "lastname", however you want to concatenate both fields and make a new field called "name". From this example "name" does not exist in the binding json data and hence the property data-field cannot be set to "name". How do you go about setting this field so that it could be searched by the built-in search utility or the filter plugin / extension? Please note that in this case I require a custom "field" as I am currently using "operate" as data-field value and it appears that the field ("operate") cannot be searched. Also, note that the custom fields can have more complex combinations and formatting than outlined in this simple example.

JSON data: {"firstname":"john", "lastname:"doe"} 

Column Header:
<th data-field="operate" data-formatter="opNameFormatter" data-sortable="false">Name</th>

JavaScript:

function opNameFormatter(value, row, index) {
   return row.firstname + '  ' + row.lastname 
}
          
Also, note that I tried setting the data-filed to one of the know fields from the binding json object and I was able to do a search on the field. However, this will not work especially if the json object and the combinations become more complex.   Cannot use the same field name for the data-filed properties especially if within the json there is a object with multiple properties (e.g return row.personalinfo.gender and return row.personalinfo.maritalstatus i n tow separate data-formatter functions). Data-field will not recognize "gender" or "maritalstatus", but will "personalinfo" which cannot be used multiple times as data-field  appears to be unique  
                         
Thanks in advance

PR #3260, opened 2017-07-12
Html strip on search

Taewa:
Sometimes I have to use `formatter` in order to create buttons or links in the table.
And when I search, the keyword searches even html tags.

For example, there is a button created by `formatter` and it will be
`<button type="button" data-index="122">Hello</button>`.
In this case, if you search `122`, you will be able to search which I don't want.
I want to search only by `Hello` because users will see only `Hello`.

So I added condition in `BootstrapTable.prototype.initSearch` with a boolean option called `searchHtmlStrip`

Here is jsFiddle
1. [Current bootstrap-table](https://jsfiddle.net/hadeath03/zggoj3ka/1/)
2. [Forked version bootstrap-table](https://jsfiddle.net/hadeath03/gLztzxLq/1/)

(This is my first time to do pull request, let me know if I miss something.)

### DIFF
--- a/dist/bootstrap-table.js
+++ b/dist/bootstrap-table.js
@@ -1277,6 +1277,9 @@
                         value = item[key];
                     }
 
+                    value = value.replace(/<[^>]*>/gi, '');
+                    console.log(value);
+
                     if (typeof value === 'string' || typeof value === 'number') {
                         if (that.options.strictSearch) {
                             if ((value + '').toLowerCase() === s) {
--- a/dist/bootstrap-table.js
+++ b/dist/bootstrap-table.js
@@ -363,6 +363,7 @@
             detailOpen: 'glyphicon-plus icon-plus',
             detailClose: 'glyphicon-minus icon-minus'
         },
+        searchHtmlStrip: false,
 
         customSearch: $.noop,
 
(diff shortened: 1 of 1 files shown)


krassowski:
@djhvscf if you have 3 minutes spare, please look at this PR. It looks like a valuable addition to already great bootstrap-table!

@Taewa I guess you will have higher chances to get this PR merged if you could add documentation entry for the newly introduced option. I think that [table-options.md](https://github.com/wenzhixin/bootstrap-table/blob/develop/docs/_i18n/en/documentation/table-options.md) is the document to edit.

Regarding the most important line:
```
value = value.replace(/<[^>]*>/gi, '');
```

It could be also written as:
```
value = $(value).text()
``` 
I did not tested that throughly but I bet it will be more secure as the hard work of parsing HTML is on jQuery side then.

wenzhixin:
I think this is the same feature with #142.
You can set `searchFormatter: false` column option to solve your problem.
https://jsfiddle.net/wenyi/zggoj3ka/2/
So close this PR, thanks anyway.

wenzhixin:
If `searchFormatter` cannot solve your problem, reopen this PR and update the documentation, thanks.

PR #3279, opened 2017-07-19
Fix cookie-expiration feature

vladimirgrezin:
Unfortuanally in commit [dce77b0aabfbe2894c8533f529e7d8a043ab77fb](https://github.com/wenzhixin/bootstrap-table/commit/dce77b0aabfbe2894c8533f529e7d8a043ab77fb#diff-9dc33f6bff50c3fdd9646828c7d37dd4) was broke cookie-expiration feature. This led to fact that only session sort cookies creates with cookie extension.
Thus, `cookieExpire` option is not working at all.
This pull request fixes that bug. `expires` prefered to `max-age` because some old browsers can only work with `expires`.

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -68,19 +68,18 @@
         switch(that.options.cookieStorage) {
             case 'cookieStorage':
                 document.cookie = [
-                        cookieName, '=', cookieValue,
-                        '; expires=' + that.options.cookieExpire,
-                        that.options.cookiePath ? '; path=' + that.options.cookiePath : '',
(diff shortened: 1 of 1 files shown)


djhvscf:
Do you have any example of this?

vladimirgrezin:
Here 2 fiddles.
[Using the current version of cookie extension](http://jsfiddle.net/pv07fyLe/3/)
[Using the fixed version](http://jsfiddle.net/jLk0snvm/2/)
To reproduce the error sort table by some field in each fiddle. Then close browser (to remove all session cookies) and go to fiddles again. As you can see, current verssion escapes sort because of session cookies.

wenzhixin:

wenzhixin:
Thanks for your contribution!

PR #3282, opened 2017-07-21
allow sorting by data-sort-value for fromHtml tables

kiafaldorius:
* Set data-sort-value on the <td> to use that value to sort instead of html()

This is for use cases where the value to be sorted is different from the display html. For example, rankings:

1st, 2nd, 3rd [...]10th, 11th, 12th, ordered by text becomes shows: 10th, 11th, 12th, 1st, 2nd, 3rd [...] which is unexpected and not how it should be.

This solves it by setting a data-sort-value attribute on the column (td) in the html, in this case, the data-sort-value will just be a number for the ranking without the ordinal text.

Thank you!

PS. I did not commit build artifacts because apparently, the build regenerates a lot of dist files, and generates some new ones...

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -203,6 +203,19 @@
         }
         return escape ? escapeHTML(value) : value;
     };
+    
+    var getItemSortValue = function (item, field, escape) {
+        var data;
+
+        if (item.hasOwnProperty('_' + field + '_data')) {
+            data = item['_' + field + '_data'];
+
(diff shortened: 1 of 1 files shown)


wenzhixin:
Please provide a [jsFiddle](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports) to show the problem, thanks!

djhvscf:

PR #3095, opened 2017-04-20
fix bug in pagination

wygdove:
bug detail:
when click active page number, it jump to document.baseURI+"#"
fixed:
change active page number's href to javascript:void(0);

PR #3292, opened 2017-07-27
Delete 'href="#' from <a href='#'></a>.This will cause a problem  in single page web application.And The href='#' is not useful for anything

xlloveyun:
Delete the href = '#' of the <a> tag. This will cause a single page application to click on the current page button jumps to an incorrect address. And <a> tag will not cause any problems when we delete href="#" .After the use of href="#"  the address bar will change the address.

### DIFF
--- a/dist/bootstrap-table.js
+++ b/dist/bootstrap-table.js
@@ -1392,7 +1392,7 @@
                     } else {
                         active = page === that.options.pageSize ? ' class="active"' : '';
                     }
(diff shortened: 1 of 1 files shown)


xlloveyun:
change to english

wenzhixin:
Thanks for your contribution!
#2646 #2645 #3095 
Fixed by https://github.com/wenzhixin/bootstrap-table/commit/3fd238ce8f5bde00af24289e554fe390d10f631f

PR #3329, opened 2017-08-24
Footer cant adjust itself on fixed table layout.

wwdd1:
Footer cant adjust itself when table layout is fixed and columns are customly sized;
Removing user-defined classes or all classes of table inside fixed footer container solved it.


wenzhixin:
@wwdd1 
Can you provide a [jsFiddle](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports) to show your problem you fixed? thanks!

djhvscf:

PR #3345, opened 2017-09-04
updateByUniqueId plus add

luhawk803:
during updateByUniqueId, when data have some new entries,  in order to avoid multiple init steps, just go ahead append the data.


wenzhixin:
I don't think this is a good idea, append row data in the update method can cause ambiguity. 
Thanks anyway.

PR #3352, opened 2017-09-08
fixed filter

mwatz122:


mwatz122:
that was an accident, sorry :/

PR #3466, opened 2017-11-25
Update bootstrap-table.css

onurkose:
You may see a missing left-border at the header of grouped tabled titles. This fix solve the problem by adding the original left-border the first child of second row of the grouped headers.


djhvscf:
The change should be applied under [develop/src/bootstrap-table.css](https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.css)

Please open a new PR with this change

PR #3472, opened 2017-11-29
Update bootstrap-table.css

onurkose:
fixes the missing left border of grouped header's second row's first td element. it was (probably) an issue when those headers created dynamically.

example: http://issues.wenzhixin.net.cn/bootstrap-table/welcome.html

### DIFF
--- a/src/bootstrap-table.css
+++ b/src/bootstrap-table.css
@@ -124,6 +124,10 @@
     border-left: none;
 }
 
+.fixed-table-container thead tr:nth-child(2) th:first-child {
+    border-left: 1px solid #ddd;
+}
+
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks for your contributor, but it will cause this bug in this case: http://jsfiddle.net/wenyi/e3nk137y/17836/

wenzhixin:
Fixed by https://github.com/wenzhixin/bootstrap-table/commit/2abf4bc61b0f8309be7ed4919a83083efb52b03e.
example: http://issues.wenzhixin.net.cn/bootstrap-table/

onurkose:
awesome!
thanks.

Issue #2508, opened 2016-07-27
Is that a Bug? Server pagination mode cannot export all data to excel


sugerlemon:
Hi guys,

 Newbie's here , I like your wonderful project!

Now i got some trouble,when I set server pagination mode,I try to export all data to excel, but it doesnt work, just got current page ,I have searched all the issues on the site,but can not find out solution. please help~

here is the code:
`<script src="/table/js/bootstrap-table.min.js"></script>`
`<script src="/table/js/bootstrap-table-zh-CN.min.js"></script>`
`<script src="/table/js/jquery.base64.js"></script>`
`<script src="/table/js/bootstrap-table-export.js"></script>`
`<script src="/table/js/tableExport.js"></script>`
`<script type="text/javascript" src="/table/js/FileSaver.min.js"></script>`
`<table id="mytab" class="table table-hover" ></table>`

 `$(function() {
            $('#mytab').bootstrapTable({
                url:'test.php',
                dataField: 'rows',
                striped:true,
                pagination: true,
    pageSize: 20,
     pageList: [5, 10, 20, 50],
    sidePagination: 'server',
    contentType: 'application/x-www-form-urlencoded',
    dataType: 'json',
    method: 'post',
     queryParamsType: 'limit',//查询参数组织方式
       queryParams: function getParams(params) {
 params.other = 'otherInfo';
      return params;
                },
                searchOnEnterKey: false,
                showRefresh: true,
                showColumns: true,
                buttonsAlign: 'left',
                toolbar: '#toolbar',
                toolbarAlign: 'right',
                showFooter:true,
                showExport:true,
                exportDataType: 'all',`
    `columns: [
                    {
                        title: '全选',
                        field: 'select',
                        checkbox: true,
                        width: 20,//宽度
                        align: 'center',//水平
                        valign: 'middle'//垂直
                    },
                    {
                        title: 'ID',//标题
                        field: 'id',//键名
                        sortable: true,//是否可排序
                        order: 'desc'//默认排序方式
                    },
                    {
                        field: 'name',
                        title: 'NAME',
                        sortable: true,
                        titleTooltip: 'this is name'
                    },
                    {
                        field: 'age',
                        title: 'AGE',
                        sortable: true,
                        footerFormatter:'sumFormatter'
                    },
                    {
                        field: 'info',
                        title: 'INFO[using-formatter]',
                        formatter: 'infoFormatter'//对本列数据做格式化
                    }
                ]
                });
        })`

[...]

PR #3477, opened 2017-12-02
Fix server pagination mode cannot export all data to excel #2508

daN4cat:
Applied the fix as per @maguitoon's comment. Thank you for the hint.


wenzhixin:
Thanks!

PR #3481, opened 2017-12-06
Cookie on filter-control with strict search enabled

Valery1991:
When using cookie in combination with filter-control and strict search enabled on a column, selecting the empty value would store it as an empty value in cookie. When returning to the page and thus getting the cookie, it would result in no results in bootstrap-table because it would literally search for ''.

This fix will allow for usage of strict search with filter-control and cookie enabled, and will show default values as expected when resetting filter.


wenzhixin:
Thanks

Issue #3486, opened 2017-12-08
The last visible column will use the change events listeners for the last column in the configuration


benny-innofaith:
Only tested and observed in firefox/windows64.

Reproducible steps:
1. Create columns with events listeners.
2. Hide the last few columns.
3. Changes in the last visible column will trigger event listeners for the last hidden column.

Go to http://jsfiddle.net/u73rc2ot/1/ and change some text in the `stargazers` column. The event listener for the `description` column will trigger and print something to console.

wenzhixin:
https://github.com/wenzhixin/bootstrap-table/commit/78f1c3278d3a5ff72ef58b42c8f6614015253694 http://jsfiddle.net/wenyi/u73rc2ot/4/

PR #3487, opened 2017-12-08
Do not attach event listeners for hidden columns

benny-innofaith:
This is a possible solution for #3486


wenzhixin:
Thianks, fixed in https://github.com/wenzhixin/bootstrap-table/commit/78f1c3278d3a5ff72ef58b42c8f6614015253694

PR #3490, opened 2017-12-09
Don't select the row when <a> tags or <button> tags are clicked.

### LINKED ISSUE #3457: Clicking links with event handlers while using clickToSelect option
See #2154 for the initial discussion. Eventually, we experienced an issue with the workaround presented there. It works fine for a normal link, but turns out problematic when you want to add some event handlers to it. For example, if you'd like to show an alert once the link gets clicked, the workaround breaks it by stopping propagation of the event. A simple and reasonable solution would involve extending this condition https://github.com/wenzhixin/bootstrap-table/blob/21ae6acc49fe0229597fa33793cb09e4753fbd2a/src/bootstrap-table.js#L1869-L1875
to check if the event target is neither link nor button. This way a checkbox gets selected when the corresponding row is clicked, unless the clicked element is a link/button. Thoughts?

(I'm creating the issue as our comments on #2154 are left without any response.)

jfly:
This fixes #3457.

### DIFF
--- a/docs/_i18n/en/documentation/table-options.md
+++ b/docs/_i18n/en/documentation/table-options.md
@@ -525,6 +525,17 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
         <td>True to select checkbox or radiobox when clicking rows.</td>
     </tr>
     <tr>
+        <td>ignoreClickToSelectOn</td>
+        <td>data-ignore-click-to-select-on</td>
+        <td>Function</td>
+        <td><code>{ return $.inArray(element.tagName, ['A', 'BUTTON']); }</code></td>
+        <td>
+        Takes one parameters:<br>
+        element: the element clicked on.<br>
+        Return true if the click should be ignored, false if the click should cause the row to be selected. This option is only relevant if clickToSelect is true.
+        </td>
+    </tr>
+    <tr>
         <td>singleSelect</td>
         <td>data-single-select</td>
         <td>Boolean</td>
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -394,6 +394,10 @@
 
         customSort: $.noop,
 
+        ignoreClickToSelectOn: function (element) {
+            return $.inArray(element.tagName, ['A', 'BUTTON']);
+        },
+
         rowStyle: function (row, index) {
             return {};
         },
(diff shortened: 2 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1911,7 +1911,8 @@
             that.trigger(e.type === 'click' ? 'click-row' : 'dbl-click-row', item, $tr, field);
 
             // if click to select - then trigger the checkbox/radio click
-            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect) {
+            var ignoreTags = [ "A", "BUTTON" ];
+            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect && !ignoreTags.includes(e.target.tagName)) {
```
wenzhixin: Lower version browser does not support `includes`, use `$.inArray` method instead.

src/bootstrap-table.js
```diff
@@ -1911,7 +1911,8 @@
             that.trigger(e.type === 'click' ? 'click-row' : 'dbl-click-row', item, $tr, field);
 
             // if click to select - then trigger the checkbox/radio click
-            if (e.type === 'click' && that.options.clickToSelect && column.clickToSelect) {
+            var ignoreTags = [ "A", "BUTTON" ];
```
wenzhixin: I think this can be consider as table option.

src/bootstrap-table.js
jfly: Sure thing.

src/bootstrap-table.js
jfly: Ok! I implemented this with an optional callback `ignoreClickToSelectOn` which takes an element and returns true or false. If you like this idea, how should I update the docs? I see `docs/_i18n/en/documentation/table-options.md`, but there are also Spanish and Chinese translations I don't feel comfortable trying to update =)

src/bootstrap-table.js
```diff
@@ -394,6 +394,10 @@
 
         customSort: $.noop,
 
+        ignoreClickToSelectOn: function(element) {
+            return $.inArray(element.tagName, [ "A", "BUTTON" ];
+        },
```
wenzhixin:
```
ignoreClickToSelectOn: function (element) {
    return $.inArray(element.tagName, ['A', 'BUTTON']);
},
```

jfly: Whoops, sorry for the typo, I didn't actually test it out. I've updated the PR (along with the documentation), and actually tested it out this time. Things should be working now!

docs/_i18n/en/documentation/table-options.md
```diff
@@ -524,6 +524,17 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
         <td>false</td>
         <td>True to select checkbox or radiobox when clicking rows.</td>
     </tr>
+    <tr>
+        <td>ignoreClickToSelectOn</td>
+        <td>data-ignore-click-to-select-on</td>
+        <td>Function</td>
+        <td>{}</td>
```
wenzhixin: The default value is `<code>{ return $.inArray(element.tagName, ['A', 'BUTTON']); }</code>`

jfly: Ahh, that's what the `{}` is, got it! I've updated the docs accordingly.

wenzhixin:
Thanks @jfly! @djhvscf can you help to update the es docs?

PR #3499, opened 2017-12-12
Fix #3150, Fix #3310

### LINKED ISSUE #3150: Fullscreen mode [Feature request]
Are there any plans to implement a full screen toolbar option and functionality? Something like summernote does.

### LINKED ISSUE #3310: Using searchText causes two requests to be sent.
Using searchText to initialize default value of search box causes two requests to be sent.

screenshot from 2017-08-13 13-09-08

As an effect, when loading larger data, user first sees "No matching records found" which is later replaced with the proper result from the second query. Detrimental to UX.

Here is the faulty code:

```
$('#table').bootstrapTable({searchText: '20'});
```

Minimal JSFiddle: http://jsfiddle.net/e3nk137y/15381/

I managed to prevent this double-request behavior, simply modifying query params on first request.
Here is my workaround: http://jsfiddle.net/e3nk137y/15364/ (also [here, on SO](https://stackoverflow.com/a/45651882/6646912)).

Is this behavior intended? Do I do anything wrong? Do you have ideas how we can fix it?

Edit: Tested on Chrome 60.0.3112.90 and Firefox 54, Ubuntu.

djhvscf:
Examples:

http://jsfiddle.net/e3nk137y/17866/ , http://jsfiddle.net/e3nk137y/17867/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -25,7 +25,8 @@
                 toggle: 'glyphicon-list-alt icon-list-alt',
                 columns: 'glyphicon-th icon-th',
                 detailOpen: 'glyphicon-plus icon-plus',
-                detailClose: 'glyphicon-minus icon-minus'
+                detailClose: 'glyphicon-minus icon-minus',
+                fullscreen: 'glyphicon-fullscreen'
             },
             pullClass: 'pull',
             toobarDropdowHtml: ['<ul class="dropdown-menu" role="menu">', '</ul>'],
@@ -43,7 +44,8 @@
                 toggle: 'fa-toggle-on',
                 columns: 'fa-th-list',
                 detailOpen: 'fa-plus',
-                detailClose: 'fa-minus'
+                detailClose: 'fa-minus',
+                fullscreen: 'fa-arrows-alt'
             },
             pullClass: 'float',
             toobarDropdowHtml: ['<div class="dropdown-menu dropdown-menu-right">', '</div>'],
@@ -366,6 +368,7 @@
         showPaginationSwitch: false,
         showRefresh: false,
         showToggle: false,
+        showFullscreen: false,
         buttonsAlign: 'right',
         smartDisplay: true,
         escape: false,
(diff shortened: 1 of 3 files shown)


src/bootstrap-table.js
```diff
@@ -360,6 +368,7 @@
         showPaginationSwitch: false,
         showRefresh: false,
         showToggle: false,
+        showFullscreen: false,
```
wenzhixin: Please help to add the documentation, thanks

src/bootstrap-table.js
```diff
@@ -516,6 +525,9 @@
         formatToggle: function () {
             return 'Toggle';
         },
+        formatFullscreen: function () {
+            return 'Fullscreen';
+        },
```
wenzhixin: need to add documentation

src/bootstrap-table.js
```diff
@@ -1265,7 +1292,13 @@
 
         this.options.pageNumber = 1;
         this.initSearch();
-        this.updatePagination();
+        if (event.firedByInitSearchText) {
+            if (this.options.sidePagination === "client") {
```
wenzhixin: if (this.options.sidePagination === 'client') {

src/bootstrap-table.js
```diff
@@ -2914,6 +2947,11 @@
         this.updatePagination();
     };
 
+    BootstrapTable.prototype.toggleFullscreen = function () {
+        var bootstraptablediv = $(".bootstrap-table");
+        bootstraptablediv.toggleClass("fullscreen");
```
wenzhixin: I think here can use `this.$el.closest('.bootstrap-table').toggleClass('fullscreen');` instead, because `$(".bootstrap-table")` will return all tables.

wenzhixin:
Thanks @djhvscf, can you rebase the develop because there are unnecessary commits.
before:
image

after rebase:

djhvscf:
Thanks. Working on this

PR #3509, opened 2017-12-19
Fix #3150, Fix #3310

### LINKED ISSUE #3150: Fullscreen mode [Feature request]

### LINKED ISSUE #3310: Using searchText causes two requests to be sent.

djhvscf:
Examples:

http://jsfiddle.net/e3nk137y/17866/ , http://jsfiddle.net/e3nk137y/17867/ https://jsfiddle.net/6te5qxy1/6/

### DIFF
--- a/docs/_i18n/en/documentation/localizations.md
+++ b/docs/_i18n/en/documentation/localizations.md
@@ -66,6 +66,11 @@
         <td>-</td>
         <td>'All'</td>
     </tr>
+    <tr>
+        <td>formatFullscreen</td>
+        <td>-</td>
+        <td>'Fullscreen'</td>
+    </tr>
 	</tbody>
 </table>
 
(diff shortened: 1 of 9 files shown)


wenzhixin:
Great! Thanks @djhvscf

Issue #3050, opened 2017-03-29
cookie 和 选择每页条数 同时使用


hidcc:
当我同时使用`data-side-pagination="server" data-cookie="true" data-page-list="[10, 25, 50, 100, ALL]"` 时，修改pagesize后刷新并不会重置offset， 导致数据偏移不正确， onPageChange事件无法直接区分是哪个改动，不知有什么较好的解决方案？

djhvscf:
Can you translate this issue to english, please?

hidsvip:
table attribute: `data-side-pagination="server" data-cookie="true" data-page-list="[10, 25, 50, 100, ALL]"` 
When I modified the pagesize and refresh the page, the param 'offset' is not be reset.

such as:
1. show page,  request: `order=asc&offset=0&limit=10`
2. next page, request:`order=asc&offset=10&limit=10`
3. change pagesize: request:`order=asc&offset=0&limit=25`
4. refresh page, request:`order=asc&offset=25&limit=25`

If the total is 11, the page is empty after refresh the page, and I can't choose the page number.

guomaoqiu:
table attribute: data-side-pagination="server" data-cookie="true" data-page-list="[10, 25, 50, 100, ALL]"
When I modified the pagesize and refresh the page, the param 'offset' is not be reset.

such as:

show page, request: order=asc&offset=0&limit=10
next page, request:order=asc&offset=10&limit=10
change pagesize: request:order=asc&offset=0&limit=25
refresh page, request:order=asc&offset=25&limit=25
If the total is 11, the page is empty after refresh the page, and I can't choose the page number.

hidsvip:
@djhvscf @wenzhixin These two plugins can't work well together

djhvscf:
Please provide a [jsFiddle](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports) to show your problem, thanks!

hidsvip:
@djhvscf  http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/cookie.html
this demo has the problem too, 
You can see it in the following ways：
1. change page to 80
2. change pagesize to 100
3. refresh page

hidsvip:
@djhvscf @wenzhixin  http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/cookie.html
this demo has the problem too, 
You can see it in the following ways：
1. change page to 80
2. change pagesize to 100
[...]

PR #3512, opened 2017-12-23
Fix for #3050, Set page number to 1 after filter

petar313:
Fix for #3050, Set page number to 1 after filter, Set page number to 1 after search


wenzhixin:
Thanks for your contribution!

PR #3520, opened 2017-12-28
StickyHeader Bootstrap 4 support

Kroneaux:
A small edit for StickyHeader that checks for the Bootstrap version and uses the "hidden" or "d-none" class to hide the floating header.


wenzhixin:
Thanks for your contribution!

PR #3539, opened 2018-01-10
Update bootstrap-table-print.js

riggzh:
add print extension support multi columns


riggzh:
Imgur
Before
Imgur
After
Imgur

riggzh:
I will append some data to test

Issue #3546, opened 2018-01-15
Printing extension prints columns that are not visible


dskilken:
When using the print extension to print the table, it grabs the entire table, not just the visible columns.  If you have used the methods .bootstrapTable('hideColumn', thecolumnindex);  and .bootstrapTable('showColumn', thecolumnindex);  to add and remove columns from view, and then go to print the table, the entire table prints.   Is there a way to have it only print the visible columns.  I am not sure if dropping or adding columns is considered "filtering."

riggzh:
It's use printIgnore in column options

riggzh:
Maybe you can edit buildTable method judge showColumn to hide columns
`function buildTable(data, columnsArray) {
                        var out = "<table><thead>";
                        for(var k = 0; k < columnsArray.length; k++) {
                            var columns = columnsArray[k];
                            out += "<tr>";
                            for(var h = 0; h < columns.length; h++) {
                                if (!columns[h].printIgnore) {
                                    out += "<th ";
                                    if (columns[h].rowspan) {
                                        out += (" rowspan='" + columns[h].rowspan + "'");
                                    }
                                    if (columns[h].colspan) {
                                        out += (" colspan='" + columns[h].colspan + "'");
                                    }
                                    out += (">" + columns[h].title + "</>");
                                }
                            }
                            out += "</tr>";
                        }
                        out += "</thead><tbody>";
                        for(var i = 0; i < data.length; i++) {
                            out += "<tr>";
                            for(var l = 0; l < columnsArray.length; l++) {
                                var columns = columnsArray[l];
                                for(var j = 0; j < columns.length; j++) {
                                    if (!columns[j].printIgnore && columns[j].field) {
                                        out += ("<td>" + formatValue(data[i], i, columns[j]) + "</td>");
                                    }
                                }
                            }
                            out += "</tr>";
                        }
                        out += "</tbody></table>";
                        return out;
                    }`

riggzh:
https://jsfiddle.net/RiggZh/yxyLurwL/2/

https://github.com/riggzh/bootstrap-table/tree/patch-1/src/extensions/print

You can see this js history to edit

wenzhixin:
Merged.

riggzh:
@wenzhixin 我没有讲这个合并进去，也许应该合？

dskilken:
[...]

PR #3548, opened 2018-01-16
 if value is 0 , it will print '0' not '-'

riggzh:
https://jsfiddle.net/RiggZh/yxyLurwL/4/

please ignore extra submission, it's solution #3546


wenzhixin:
Thanks

Issue #3558, opened 2018-01-23
Header width does not match column width


Frebla:
I have a problem on my 15,6 inches screen.
Example: http://issues.wenzhixin.net.cn/bootstrap-table/index.html

![2018-01-23_10-31-22](https://user-images.githubusercontent.com/32418668/35263385-98742178-0029-11e8-8beb-cd503b5d9e1f.png)

hensing:
@frebla did you use the latest dev-version or a release?
This bug should be fixed in the dev-branch

Frebla:
Yes, I did. That problem exist not only on my computer, just go to 
http://issues.wenzhixin.net.cn/bootstrap-table/index.html
and try to scale it. You will see.

PR #3569, opened 2018-01-26
Fix-bug-#3558

Frebla:
I think I have found solution for the problem with scale of the table.


djhvscf:
Please resolve the conflicts and provide a example of this working

djhvscf:

PR #3607, opened 2018-02-16
Added the possibility to use rowStyles when using tree grid.

pmorelli92:
Bootstrap-table has the rowStyle property in order for us to set classes for an entire row.
For example:

```
rowStyle: (row, index) => {
	if (row.foo != undefined) {
		return { classes: 'success' };
	} else {
		return { classes: 'danger' };
	}
}
```

When using the plugin tree-grid we can see that it overrides our customized rowStyle function:

```
that.options.rowStyle = function (item, idx) {
	var x = item[that.options.idField] ? item[that.options.idField] : 0;
	return {
		classes: 'treegrid-' + x
	};
};
```
So after the tree is initialized, we lost our rowStyle. In order to workaround this I added a function named: `afterTreeRowStyle` that applies our customizations after the tree initialization.

### DIFF
--- a/src/extensions/treegrid/bootstrap-table-treegrid.js
+++ b/src/extensions/treegrid/bootstrap-table-treegrid.js
@@ -1,6 +1,6 @@
 /**
  * @author: YL
- * @version: v1.0.0
+ * @version: v1.1.0
  */
 !function ($) {
     'use strict';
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks for your PR, I think we can extend `rowStyle` instead of adding a new option.

PR #3611, opened 2018-02-20
Update bootstrap-table-de-DE.js

qwe900:
Added pagination , export, clear filters, removed dots, added plural.


wenzhixin:
Thanks

Issue #3653, opened 2018-03-11
getHiddenRows function does not show hidden rows when passed "true"


chrissharp80:
According to the doc, if `getHiddenRows` is called and passed `true` then it will also show the rows which were hidden. This is desired because there is a performance issue when showing the rows with a loop.

If rows are hidden in the table and the `getHiddenRows` function is invoked with the `show` argument passed as true the function ignores the boolean. 

1. Hide some rows
2. Call `getHiddenRows` as follows `$table.bootstrapTable('getHiddenRows', true);`
3. Observe that the rows have not re-appeared.

There is no need to show a jsfiddle of this bug. Just looking at the function in `bootstrap-table.js` confirms that the `show` argument is ignored.

chrissharp80:
I have a fix for this and will be opening a PR shortly.

wenzhixin:
Added an example to develop branch: http://dev.wenzhixin.net.cn/bootstrap-table-examples/#methods/getHiddenRows.html

PR #3655, opened 2018-03-11
Fix 'getHiddenRows()' function does not show hidden rows when passed 'true'

chrissharp80:
Addresses https://github.com/wenzhixin/bootstrap-table/issues/3653


wenzhixin:
Thanks!

PR #3674, opened 2018-03-15
Remove `fixed` option from resizable extension

pmichna:
According to [colResizable documentation](http://www.bacubacu.com/colresizable/#attributes) the `fixed` option is deprecated. `resizeMode` is the one that should be used.


wenzhixin:
Sorry, we need to merge this to develop branch: https://github.com/wenzhixin/bootstrap-table/pull/3690

PR #3682, opened 2018-03-17
Update bootstrap-table-export.js

mohshraim:
solving problem of multiple export button will be displayed when using export extension with advancedSearch (bootstrap-table-toolbar) extension...

this show the bug.
screen shot 2018-03-17 at 11 47 58 pm

solve it by adding the export button to index 0 of btnGroup
$btnGroup[0] instead of $btnGroup

### DIFF
--- a/src/extensions/export/bootstrap-table-export.js
+++ b/src/extensions/export/bootstrap-table-export.js
@@ -67,7 +67,7 @@
                         '</button>',
                         '<ul class="dropdown-menu" role="menu">',
                         '</ul>',
-                    '</div>'].join('')).appendTo($btnGroup);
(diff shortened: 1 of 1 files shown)


wenzhixin:
Please provide a fiddle to solve the problem, I think it's better to update `$btnGroup = this.$toolbar.find('>.btn-group'),` to `this.$toolbar.find('> .columns.btn-group'), `?

BTW: please update the PR because we need to merge to develop branch.

mohshraim:
@wenzhixin 
your suggestion didn't work..

 i think its better to be solved on advanced-toolbar extension as its adding a new btn-group element without check if one already exist

wenzhixin:
Fix the bug in #3720

PR #3685, opened 2018-03-19
Adding support for cookies and fixing an existing issue with filter control

### LINKED ISSUE #1769: check/uncheck methods does not handle singleSelect option
Looking at the code, `checkAll` and `uncheckAll` methods are equivalent as clicking on the "select all" checkbox:

``` javascript
this.$container.off('click', '[name="btSelectAll"]')
.on('click', '[name="btSelectAll"]', function () {
    var checked = $(this).prop('checked');
    that[checked ? 'checkAll' : 'uncheckAll']();
    that.updateSelected();
});
```

This is nice.

The problem is the following: there is one code for `check` / `uncheck` methods and one **different** code for the click handler on rows checkboxes:

``` javascript
// check & uncheck methods

BootstrapTable.prototype.check = function (index) { this.check_(true, index); };

BootstrapTable.prototype.uncheck = function (index) { this.check_(false, index); };

BootstrapTable.prototype.check_ = function (checked, index) {
    var $el = this.$selectItem.filter(sprintf('[data-index="%s"]', index)).prop('checked', checked);
    this.data[index][this.header.stateField] = checked;
    this.updateSelected();
    this.trigger(checked ? 'check' : 'uncheck', this.data[index], $el);
};
```

``` javascript
// click handler on checkboxes

this.$selectItem = this.$body.find(sprintf('[name="%s"]', this.options.selectItemName));
this.$selectItem.off('click').on('click', function (event) {
    event.stopImmediatePropagation();

    var $this = $(this),
        checked = $this.prop('checked'),
        row = that.data[$this.data('index')];

    if (that.options.maintainSelected && $(this).is(':radio')) {
        $.each(that.options.data, function (i, row) {
            row[that.header.stateField] = false;
        });
    }

    row[that.header.stateField] = checked;

    if (that.options.singleSelect) {
        that.$selectItem.not(this).each(function () {
            that.data[$(this).data('index')][that.header.stateField] = false;
        });
        that.$selectItem.filter(':checked').not(this).prop('checked', false);
    }

    that.updateSelected();
    that.trigger(checked ? 'check' : 'uncheck', row, $this);
});
```

Looking at this, we can see that `check` / `uncheck` methods doesn't handle `maintainSelected` and `singleSelect` options, which can lead to have multiple rows selected even if `singleSelect` is set to true `true`.

Would be nice to have one single function for both (event handler and public method), what do you think ?

### LINKED ISSUE #1983: Just one cookie at a time?
Been testing this plugin, which is great, but there's this issue:

When I use both the search filter cookie (tableId.bs.table.pageNumber) and the page navigation cookie (tableId.bs.table.searchText), only the latter is applied to the bootstrap table.

Otherwise both of them work if I use them independently (if I filter on first page, or navigate to another page without any filter).

### LINKED ISSUE #3474: Filter Control Multiple choice bug
like title

**Bug1**
you can try 1 $0

[http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/filter-control.html](http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/filter-control.html)

**Bug2**
th data-field="name" data-filter-control="input">Item Name</th>//export(json,xls or..) ok
th data-field="price" data-filter-control="select">Item Price</th>  //export(json.xls or..) not ok

djhvscf:
Fix #1983 https://jsfiddle.net/djhvscf/b7518dkf/15/
Fix #1769 https://jsfiddle.net/djhvscf/mgg544eL/9/
Fix #3474


djhvscf:
Hi @wenzhixin  I need then to rewrite these changes. Closing it since it is invalid at this point.

wenzhixin:
OK, @djhvscf

PR #3718, opened 2018-03-30
feature/es6

wenzhixin:
Use es6 to rewrite bootstrap-table.js and export extensions

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,7 @@
+module.exports = {
+    "extends": "standard",
+    "globals": {
+      "jQuery": true,
+      "$": true
+    }
+};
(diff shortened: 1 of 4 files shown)


cabeth:
Hi Wenzhixin, 
I really like your table, and i'm trying to get it to work in my project. However, i'm not able to integrate your module using es6 import statement "import 'bootstrap-table'" (i also updated the project to the latest beta build). I get an error: jQuery is not defined. For my build i use Browseify with Babelify (with es2015 preset).  It does work when i'm using require('bootstrap-table') and setting the window.jQuery to jquery installed using an es6 import, but that doesn't feel right ;). 
Hope you can help me!

PR #3737, opened 2018-04-11
Limit fullscreen CSS rule scope

lucaswerkmeister:
Only apply the fullscreen rule to bootstrap-table elements, not any other elements that also happen to have a fullscreen class.


djhvscf:
Thanks. You're right

lucaswerkmeister:
OK, thanks for merging :)

PR #3787, opened 2018-05-10
Changed pull- to float- because pull was removed from Bootstrap 4

NewLunarFire:
`pull-left` and `pull-right` have been removed from Bootstrap 4 in favor of `float-left` and `float-right`. This requires a minor change to css and js source files.

[https://getbootstrap.com/docs/4.0/migration/#utilities](https://getbootstrap.com/docs/4.0/migration/#utilities)

> Added .float-{sm,md,lg,xl}-{left,right,none} classes for responsive floats and removed .pull-left and .pull-right since they’re redundant to .float-left and .float-right.

### DIFF
--- a/src/bootstrap-table.css
+++ b/src/bootstrap-table.css
@@ -280,7 +280,7 @@
     padding: 0 !important;
 }
 
-.bootstrap-table .pull-right .dropdown-menu {
+.bootstrap-table .float-right .dropdown-menu {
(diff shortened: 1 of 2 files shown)


NewLunarFire:
Nevermind, you have to load Bootstrap before bootstrap-table now to make sure the library uses the configuration for the correct version of Bootstrap

PR #3821, opened 2018-05-27
fix the group-by-v2 the header would be show only one column.

chinarma:
fix the bug when use group-by-v2 the header would only show one column name.


djhvscf:
This needs to resolve the conflicts. Please open a new PR

PR #3849, opened 2018-06-17
fix german translation

swiffer:


sebastianjohngvl:
Hi,
When will this fix be released?

PR #3854, opened 2018-06-19
Replace opencollective with opencollective-postinstall

sterngeist:
https://github.com/opencollective/opencollective-cli/issues/3#issuecomment-378039208

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "directories": {
     "doc": "docs"
   },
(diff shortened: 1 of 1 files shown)


sterngeist:
Simply updated to latest Node.js v10.4.1, used "npm install -g opencollective" first, afterwards "npm install bootstrap-table" and install process worked like a charm. So pull request is obsolete.

PR #3859, opened 2018-06-21
Fix plural commited in #3611

danielh84:
actually there is "Zeilenn" if there are more than 1 rows... it has to be "Zeilen"


danielh84:
duplicate: #3849

PR #3872, opened 2018-06-30
Fix #2545: customSearch not triggering custom function

### LINKED ISSUE #2545: customSearch not triggering custom function
I am using bootstrap-table on osx. Trying to implement customSearch. However, when I set data-custom-search to `data-custom-search="customSearchFunction"` in the table attributes in the HTML, **_customSearchFunction_** does not get called. Instead I get this error:

```
bootstrap-table.js:1229 Uncaught TypeError: this.options.customSearch.apply is not a function
```

That's because the function name is being treated as a string. If I set a breakpoint before that line and enter this command in the console `this.options.customSearch = customSearchFunction`, then things work properly (temporarily).

How do I tell bootstrap-table to call my customSearchFunction (the right way)?

amiart:
This fixes a regression introducted by Pull request #3152

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1254,7 +1254,11 @@
     initSearch () {
       if (this.options.sidePagination !== 'server') {
         if (this.options.customSearch !== $.noop) {
-          window[this.options.customSearch].apply(this, [this.searchText])
+          if (typeof this.options.customSearch === 'string') {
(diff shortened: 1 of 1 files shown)


amiart:
Here is [jsFiddle](http://jsfiddle.net/amiart/e3nk137y/34935/)

You will get the following error:

> TypeError: window[this.options.customSearch] is undefined

when you start typing in Search field and the patch is not appied.

PR #3903, opened 2018-07-24
Fix the bug touching the possibility to filter several columns and cl…

RRENOU:
…ear filters via according button.


djhvscf:
Hi @RRENOU this PR is including a change under dist files. Please only modify src files.

Also provide an example of this PR working.

PR #3920, opened 2018-07-30
Fix iterating for calculate nested function

savin-alexey:


savin-alexey:
@djhvscf  examples in this case is redundant:
This case, if `name = 'obj.func1.func2'`
Expected result - `func = window.obj.func1.func2`
Result with iteration over name is - `window.o.b... etc`
We must compose a function using words and not letters
```JS
  if (typeof name === 'string') {
    // support obj.func1.func2
    const names = name.split('.') // names = ['obj', 'func1', 'func2']

    if (names.length > 1) {
      func = window
      for (const f of names) { // if using `name` variable, result would be - window.o.b.j....... 
        func = func[f]
      }
}
...
```


P.S.
[The last working code was here](https://github.com/wenzhixin/bootstrap-table/commit/57915afd9a740721e5414c8fb0c561902d4f9650#diff-95c58f988cb3038d8e3b2a5176112424L166)

PR #3940, opened 2018-08-08
Allow different x-editable configuration per table row

kkozlik:
In my use case I need different x-editable elements for each row in one column in dependency on datatype on that row. 

This pull request allow `editable` option to be a function that can generate different configuration for x-editable.

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -111,6 +111,20 @@
                 return;
             }
 
+            var data = that.getData();
+
+            that.$body.find('a[data-name="' + column.field + '"]').each(function(i, element){
+                var $element = $(element);
+                var $tr = $element.closest('tr');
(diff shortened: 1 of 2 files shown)


src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -111,6 +111,20 @@
                 return;
             }
 
+            var data = that.getData();
+
+            that.$body.find('a[data-name="' + column.field + '"]').each(function(i, element){
+                var $element = $(element);
+                var $tr = $element.closest('tr');
+                var index = $tr.data('index');
+                var row = data[index];
```
kkozlik: I am not sure if I can safely use:
```
var row = data[i];
```
here and get rid of the two lines above...

djhvscf:

kkozlik:
Here is the fiddle: http://jsfiddle.net/karelkozlik/vtya5cf0/25/

PR #3975, opened 2018-09-12
fix "check" function using wrong variable name

Bigpet:
$this is undefined here, $el should be semantically the right thing here.


djhvscf:

Bigpet:
any radio box example is broken, but here's a simple one:

broken:
http://jsfiddle.net/1Ldw6ynj/

working: 
http://jsfiddle.net/5bhymsfu/

PR #4077, opened 2018-12-14
icon bug

zhangshitang:
in bootstrap4 using font-awesone icon ,we can't find class fa-toggle-up and fa-toggle-down in new font-awesone-4.7.0  I changed


wenzhixin:
Just only need to change src/bootstrap-table.js.

PR #4089, opened 2018-12-30
Fix #4008

### LINKED ISSUE #4008: Select all rows and "keyup" EventListener empty the table when using Filter Control extension
When using Filter Control extension, once clicking on the select all rows in the header and after that, pressing on any other key in the keyboard, all of the table rows are removed.

here is a link to a fiddle:

https://jsfiddle.net/eitanmg/e3nk137y/42640/

to restore the issue, click on the select all rows checkbox in the column header, and then press on any key on your keyboard, and you will see that all the rows disappeared.

how can i overcome this issue?

djhvscf:
Fix #4008
Rewrite filter-control to ES6


wenzhixin:
Merged in #4090, thanks @djhvscf

djhvscf:
@wenzhixin It is my pleasure and it is good to see you back 👍

PR #4094, opened 2019-01-01
Filter control cookies improvements

djhvscf:
feat(cookie extension): rewrite boostrap-table-cookie to ES6
feat(cookie extension): Save filterBy method
feat(filter-control extension): Add placeholder as a empty option to the select controls
feat(filter-control extension): Add clearFilterControl method in order to clear all filter controls
feat(js): Improve pagination options

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -307,7 +307,7 @@
     paginationDetailHAlign: 'left', // right, left
     paginationPreText: '&lsaquo;',
     paginationNextText: '&rsaquo;',
-    paginationSuccessivelySize: 3, // Maximum successively number of pages in a row
+    paginationSuccessivelySize: 5, // Maximum successively number of pages in a row
     paginationPagesBySide: 1, // Number of pages on each side (right, left) of the current page.
     paginationUseIntermediate: false, // Calculate intermediate pages for quick access
     search: false,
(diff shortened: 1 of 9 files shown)


src/bootstrap-table.js
```diff
@@ -307,6 +307,9 @@
     paginationDetailHAlign: 'left', // right, left
     paginationPreText: '&lsaquo;',
     paginationNextText: '&rsaquo;',
+    paginationSuccessivelySize: 3, // Maximum successively number of pages in a row
```
wenzhixin: I think here is better to be `5` by default, which is consistent with the previous usage habits.

src/bootstrap-table.js
```diff
@@ -1440,83 +1443,96 @@
...
         }
 
-        if (this.totalPages >= 6) {
-          if (this.options.pageNumber >= 3) {
-            html.push(
-              Utils.sprintf('<li class="page-item page-first%s">',
-                this.options.pageNumber === 1 ? ' active' : ''),
-              '<a class="page-link" href="#">', 1, '</a>',
-              '</li>')
+        if (this.options.pageNumber < (this.options.paginationSuccessivelySize - 2)) {
```
wenzhixin: Should be: `if (this.options.pageNumber < (this.options.paginationSuccessivelySize - 1)) {`

Because when `this.options.pageNumber = 3`, the pagination shows:
`1 2 3 4 ...`

wenzhixin:
Thanks, @djhvscf.

PR #4104, opened 2019-01-09
encode need for no latin symbol

alsovdev:
browser crash to 500 then tried ti read cookie without encodeURIComponent value


wenzhixin:
Fixed https://github.com/wenzhixin/bootstrap-table/commit/4403520e29201edaa9ee07c62e717bce22b3de71, thanks!

Issue #3394, opened 2017-10-02
groupBy formatter


leipzig:
I have most of my content uri-encoded. I use a formatter for these columns
```
    function decodeURIFormatter(value, row, index) {
        return decodeURIComponent(value);
    } 
```

For the groupBy rows there is no formatter, so I end up doing something like:
```
        $table.on('all.bs.table', function (e: Event) {
            $('.info td:nth-child(3)').each( function( index ){
                $(this).html(decodeURIComponent($(this).html()));
            });
        });
```

Is there a more elegant solution to applying a groupBy formatter?

Falseee:
https://bootstrap-table.com/docs/extensions/group-by-v2/#groupbyformatter
#4109

PR #4109, opened 2019-01-13
 bootstrap-table-group-by header formatter

Hoppelite:
An additional option for bootstrap-table-group-by to allow formatting of group headers.
Option is in the form of a function with the signature groupByFormatter(value, idx, data) where:
value: is the original header text
idx: is the index of the group
data: is an array of the row objects for this group
return: a string to be used as the group header html

This is useful when you require the group header to be formatted differently from the field value.


djhvscf:

wenzhixin:
Need to update the docs: https://github.com/wenzhixin/bootstrap-table/blob/develop/site/docs/extensions/group-by-v2.md

Hoppelite:
Possible resolution of #3394

PR #4114, opened 2019-01-15
Bootstrap-Table Pipeline extension feature request

doug-the-guy:
Bootstrap-Table Pipeline is an extension that allows a hybrid approach
to paging. A pipeline "cache" size can be set to determine how many
rows are returned by the server. The client side will page through
this data set until a new request is required (sorting, searching,
paging out of current cache or page size change).

This will allow for a balance of performance between all of the data
at once (client-side paging) and a new request for each page
(server-side paging).


djhvscf:
This extension looks cool but we will need an example

PR #4123, opened 2019-01-20
Adding pageNumber to the server side pagination

### LINKED ISSUE #3974: Set a page other than 1 when using searchText
When initializing the bootstrap-table, if I set { searchText: "some value", pageNumber: 2 } using serverSide pagination, I always get page 1 back.

I am also using queryParamsType: custom.

Looking at the headers going to the server, all my other values that come from the params object in the queryParams function call have the initial values I set like page size, sorting, the searchText, etc. but the pageNumber is always 1.

If I do not set the searchText. I will get page 2 data back.

I am trying to return the user to a previous state of the page where they searched for a value, moved to page 2 and then drilled into some data. When they click back to this page, I need it to reload to page 2 of the searched data.

djhvscf:
Fix #3974

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@ ChangeLog
 
 ### 1.13.3
 - **New(js):** Added `pageNumber` parameter to the server side pagination.
-- **New(group-by-v2 extension): ** New `groupByFormatter` option.
-- **New(pipeline extension): ** New pipeline extension `bootstrap-table-pipeline`.
+- **New(group-by-v2 extension):** New `groupByFormatter` option.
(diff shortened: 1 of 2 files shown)


wenzhixin:
Thanks!

wenzhixin:
Sorry @djhvscf, I think that issue is not a bug, because when we use `queryParamsType = 'limit'`, the `pageNumber` will send to the server as `offset` param: http://jsfiddle.net/wenyi/5wsfmr6k/.

> Request URL: https://examples.wenzhixin.net.cn/examples/bootstrap_table/data?search=Item+21&order=asc&offset=10&limit=10

I had removed the code in PR #4131.

Issue #3780, opened 2018-05-08
Card-view click events issue


Miguel-Rey:
I need to handle some 'click' events in my table. There is no problem in standar view, but when I toggle it to Card View it doesnt work. It seems that two things happens:

a) If there is any content which handles a click event, Card View prevents that from triggering. The only work with this is with the custom event "click-cell"
b) When you toggle to Card View, all the content gets wrapped in the first cell, so It doesn't matter where you click. The event triggered is always 'click-cell' of the first td.

Here is a fiddle. As you can see, there is no way to click on the "description".

http://jsfiddle.net/gakjn61m/

jimgit:
Try using the row variable:

See this[ JSFIDDLE](http://jsfiddle.net/jimkennelly/57yu4zqx/)


![image](https://user-images.githubusercontent.com/1921941/50910415-5b39b500-13f3-11e9-913f-24a3d477b5c9.png)

bob526:
Thank you for your quick reply.

Yes, for people who only want the data in row variable that will work.
But, in my real use case, I need the correct field variable because I will do different jobs according to the clicked field.

I updated the jsFiddle to reduce the ambiguity.
[updated jsFiddle](http://jsfiddle.net/jm02oaLk/13/)

wenzhixin:
Merged.

PR #4124, opened 2019-01-20
Fix cardview click event

bob526:
This fix the #3780 

``` javascript
const $cardviewArr = $(target).parents('.card-views').children()
const $cardviewTarget = $(target).parents('.card-view')
const index = this.options.cardView ? $cardviewArr.index($cardviewTarget) : $td[0].cellIndex
```
I only modified these lines.

A [not working jsFiddle](http://jsfiddle.net/jm02oaLk/13/)
1. Click cardview toggle
2. Click hyperlink
3. Read the **field = xyz** in `#fake_console`, you can see the wrong field string.

[My patch is working jsFiddle](http://jsfiddle.net/onb487u3/)
Read the field = xyz part, this time, it shows the correct field string


wenzhixin:
Thanks for your contribution!

PR #4148, opened 2019-01-31
theadClasses is not set by default

TantumLabs:
If a thead exists and $.BootstrapTable.DEFAULTS.theadClasses is set the class is never set on init.


wenzhixin:
Updated the code format, thanks!

Issue #4161, opened 2019-02-12
showFullscreen button click does nothing


earlseay:
showFullscreen button click does nothing

Any browser. The click event is not firing for the showFullscreen button effectively doing nothing. The button is there, but there is no event attached.

1. set showFullscreen: true
2. render the table
3. click the ful screen button
4. nothing happens

I believe this is because the fullscreen button click event was attempted to be attached to a button that wasn't created yet. You just need to move the event handler attachment down about 20-40 lines where the rest of them are.

earlseay:
Pull request open for fix: #4162

wenzhixin:
Merged into develop branch :)

PR #4162, opened 2019-02-12
fix showFullscreen onclick misplacement

earlseay:
move the onclick event handler down to after the button has actually been created


earlseay:
This pull request will fix #4161

wenzhixin:
Thanks for your contribution!

Issue #3874, opened 2018-07-01
Clear filterControl with Cookie


manukieli:
I use filterControl and Cookie that store to localstring.

When I filter some column data with input filter and reload the page, the talbe still filtered because the cookie is active. It's OK.
But if I click to "Clear Filter" button and after I reload the page, the filter still aply.
To Chrome console, when I click to "Clear Filter", I see bs.table.filterControl is deleted and  is recreated.
To resolve it, to <b>BootstrapTable.prototype.clearFilterControl</b> function, I've add this line `            $.each(that.options.filterControls, function (i, item) {
                item.text = '';
            });`
after this line:<br>`$.each(that.options.valuesFilterControl, function (i, item) {
                item.value = '';
            });`

I'dont' know if this lines I've added could create another problem, but it resolve my problem to clear filterControl.

I'm using filterControl v2.1.3 and cookie v.1.2.3

djhvscf:
This is an issue in Cookie extension. Working in progress

djhvscf:
Fixed in https://github.com/djhvscf/bootstrap-table/commit/fc5b29fec3ea7ae7b89226c81e8bdbb9eb02e6df

djhvscf:
Fixed here https://github.com/wenzhixin/bootstrap-table/pull/3968

manukieli:
The problem back again with the version v.13.3
You can see the issue here:
https://jsfiddle.net/2ot4d5r3/11/

manukieli:
I've the same problem with the version v.1.13.4
I'm using localstorage to store filter data.
I've add this line into the file bootstrap-table-filter-control.js
After this lines (line 829):
`
            $.each(that.options.valuesFilterControl, function (i, item) {
              hasValues = hasValues ? true : item.value !== '';
              item.value = '';
            });
`
I've add this line:
`
            $.each(that.options.filterControls, function (i, item) {
              item.text = ''
            });
`
And the issue is fixed.

manukieli:
I've the same problem with the version v.1.13.5
I'm using localstorage to store filter data.
I think it's because when your script invoke **function clearFilterControl()**, this function create the variable cookies with the function **UtilsFilterControl.collectBootstrapCookies()**.
And this function find only cookie and not into the localStorage.

ktecho:
@djhvscf It's happening to me with 1.14.1 using default storage (cookieStorage).

[...]

PR #4181, opened 2019-02-25
Fix issue #3874

manukieli:
I'm using localstorage to strore cookies.
When I want to clear it your actual code read only cookies, with my update, if the cookie array is 0 they read localstorage.
And my update clear that.options.filterControls and I put the resetSearch before if the code is exit with the controls.length condition.


wenzhixin:
Please update the src file instread of dist file.

manukieli:
> Please update the src file instread of dist file.

Yes, I've done it later.

Issue #3042, opened 2017-03-25
filterBy function with <, <=, > and >= than


ManicPumpkin:
Hey guys,

is there a way to filter values inside a column of the BS table with a filter greater or smaller than a given value? 

For example filtering the temperature, which is < 20°C in a column which has given values like [34, 20, 8, -6, 56, -24]. Currently I'm only able to filter for a temperature equal a given value.

wenzhixin:
Merged in the develop branch.

vossivossi:
> Merged in the develop branch.

Where can I download the develop branch?

wenzhixin:
This feature has been released in the latest version.

vossivossi:
Ok, thank you very much fpr the fast reply. I will give it a try!

farayman3:
I believe this feature does not work for comparing negative values (E.g. "< -10" or "-10 >") or am I not doing this right?

UtechtDustin:
@farayman3 your right, i created a PR which fix that issue!

PR #4218, opened 2019-03-20
Impoved Search - Added check for comperations (<,>,<=,=<,>=,=>)

Falseee:
I impoved the search checks so comperations search (<,>,<=,=<,>=,=>) are possible now.
This feature was requested by @ManicPumpkin on Issue #3042.

The following comperations are possible:
```
> value
< value
<= value
=< value
>= value
=> value
value <
value >
value <=
value =<
value >=
value =>
```

JsFiddle Demo: https://jsfiddle.net/s2gpvoqu/

### DIFF
--- a/site/docs/api/table-options.md
+++ b/site/docs/api/table-options.md
@@ -705,6 +705,14 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
 - **Detail:**
 
   Enable the search input.
+  
+  There are 3 ways to search:
+  - The value contains the search query (Default).   
+    Example: Github contains git.
+  - The value must be identical to the search query.  
+    Example: Github (value) and Github (search query).
+  - Comparsions (<, >, <=, =<, >=, =>)   
+    Example: 4 is larger than 3.
 
 - **Default:** `false`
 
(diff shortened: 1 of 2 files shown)


wenzhixin:
Thanks for your contribution!

It seems that there are problems in [this case](https://jsfiddle.net/wenyi/5847mypf/4/), let's see how to improve it.

Falseee:
I fixed the bug ([working example](https://jsfiddle.net/of1cg4tp/1/)), also i improved the documentation about searching (description of the 3 search types (contains, strict and comparison)).

wenzhixin:
Thanks! Fixed the lint errors: https://github.com/wenzhixin/bootstrap-table/commit/d6b1c78812ae36abf8d78d6d98c29a09e7a41487.

Falseee:
@wenzhixin 
My fault i already fixt hat lint errors, but forgott to push it...sorry!
Can you explain me why all `;` will removed, is there a good reason for that or just to save some memory ?

wenzhixin:
Because other ES6 projects remove `;`, so use the same rules, each line can be one less character :)
Also, refer to some of the more popular projects, they also do not use `;`. 
For examples:
https://github.com/twbs/bootstrap/blob/master/js/src/alert.js
https://github.com/vuejs/vue/blob/dev/src/core/index.js

Issue #4207, opened 2019-03-14
multiple-sort not working with data-query-params="functionName"


twu2:
If we assign queryParams with data-query-params="functionName", then the queryParams is a string, not a function.
It will make multiple-sort not working.

This patch will fix this issue.

[multiple-sort-data-query-params.zip](https://github.com/wenzhixin/bootstrap-table/files/2964652/multiple-sort-data-query-params.zip)

twu2:
example for this bug:
https://jsfiddle.net/c9bfay0g/1/

wenzhixin:
Merged in the develop branch.

PR #4222, opened 2019-03-21
check if queryParams is a function to prevent critical issues

Falseee:
Just the PR for the fix by @twu2 in #4207 in a minimal version. 

Working Jsfiddle: https://jsfiddle.net/mLx0ztwy/1/


wenzhixin:
It's better to use `return $.fn.bootstrapTable.utils.calculateObjectValue(that.options, t, [params])` instead. 

https://jsfiddle.net/wenyi/qzo3f7kt/4/

Falseee:
Good to know that this function exists, im using it now ! :)

wenzhixin:
Thanks!

Issue #4227, opened 2019-03-22
totalNumber of rows always 0 after inserting new row


basementmedia2:
I want a new row to be inserted at the end of the table.
So my plan was to get the total number of rows and then insert the new row at index = totalNumRows+1.

But i always get total number of rows = 0...

Here's a jsfiddle
https://jsfiddle.net/7v13agyc/

wenzhixin:
It seems that your fiddle cannot show the problem.

basementmedia2:
Hi,

shouldn't the event  

```
$("#test_table").on('post-body.bs.table', function () {
  console.log($("#test_table").bootstrapTable('getOptions').totalRows);
})
```

which is in the fiddle log the number of total rows in the console?

Falseee:
I agree that `.bootstrapTable('getOptions').totalRows` should return the count of all row (same count as shown in pagination).
As workaround you can use `$("#test_table").bootstrapTable('getData').length`.
Example for the workaroung: https://jsfiddle.net/btLoaj1p/

wenzhixin:
Merged in the develop branch.

PR #4228, opened 2019-03-22
Also set the totalRows option if pagination is not active

Falseee:
Currently the totalRows option is only set if pagination is true (option is set in initPagination).
`$("table").bootstrapTable('getOptions').totalRows` will always return 0 if pagination is set to false (more informations in issue #4227)

Issue found by @basementmedia2 in issue #4227

Not working jsfiddle: https://jsfiddle.net/g71pdr42/2/
Working jsfiddle: https://jsfiddle.net/g71pdr42/3/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2077,6 +2077,10 @@
       this.updateSelected()
       this.resetView()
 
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks for your contribution!

Issue #2392, opened 2016-06-17
updateCellByUniqueId


Dimigergo:
Hello!

I use the x-editable plugin to make changes in a table. When editable-save.bs.table event is raised, there is no row index, but I have the newValue.UId. When the user change a field I would like to update only one cell in this row. 
If I call updateByUniqueId the whole row will be updated, and the modification highlight (bold text) remove from the editable cell.

Is it possibly to add a method updateCellByUniqueId like updateByUniqueId (row)?

Or it is possible to get the row index from row UniqueId?

Thanks.

wenzhixin:
After the user change a field, the row will be updated automatically, you do not have to manually update it.

christlee1989:
@wenzhixin  I have the same issue , i have one custom hidden cell(visible:false) ,  when editable-save.bs.table triggered,i want to change this cell's value. I need row index too.. or 'updateByUniqueId'

wenzhixin:
Merged.

PR #4229, opened 2019-03-22
Added row index as parameter for the onEditableSave event

Falseee:
As feature request from @Dimigergo (issue #2392) i added the row index as parameter to the onEditableSave Event.

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -11,7 +11,7 @@
     onEditableInit () {
       return false
     },
-    onEditableSave (field, row, index, oldValue, $el) {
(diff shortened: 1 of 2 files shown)


src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -11,7 +11,7 @@
     onEditableInit () {
       return false
     },
-    onEditableSave (field, row, oldValue, $el) {
+    onEditableSave (field, row, index, oldValue, $el) {
```
wenzhixin: It's better to use `rowIndex` to be same with documentation.

djhvscf: Agree

Falseee: done

Issue #3295, opened 2017-07-29
filterby for multiple columns with "OR" condition


thomas-nguyen-3:
I am looking at the filterby example here: http://issues.wenzhixin.net.cn/bootstrap-table/#methods/filterBy.html

and wonder if i can apply the filterby code for multiple columns in this way:

example: from the above filterby link. I would like to find the (item Name = Item 1) or (Item Price = $1 or $2)

I understand that for the "And" condition: (item Name = Item 1) AND (Item Price = $1 or $2)

I will use 
$table.bootstrapTable('filterBy', { name: "Item 1", price: [$1, $2] });

But is there a way to do the above with "OR" condition:
 (item Name = Item 1) OR (Item Price = $1 or $2)

mlbaquerizo:
I would like this as well. I can only think of:
```
//get array of all table data
var fullTableData = $table.bootstrapTable('getData');

//filter using javascript built in array filter function
var filters = fullTableData.filter(function(item){
    return (item.name == "Item 1" || item.price == 1 || item.price == 2)
});

// reload bootstrapTable with filtered array
$table.bootstrapTable('load', filters);
```
Then to unfilter you would need to reload with `fullTableData`:
```
$table.bootstrapTable('load', fulltableData);
```
Kind of a hack since you would need to store the complete data somewhere and then refilter any time you want to do another OR filter...

Falseee:
Its already implemented/merged the reopening was a mistake i guess ? @djhvscf

PR #4230, opened 2019-03-23
Improved the filterBy function with OR condition and custom filter algorythm

Falseee:
This PR improves the filterBy function, its basicly adds a or condition check(requested by @thomas-nguyen-3  in #3295) and additionally i added a option to use a custom filter algorithm.

The new functionalities are documented.

Jsfiddle with all 3 filter algorithm: https://jsfiddle.net/j1o6p4s8/
In the Jsfiddle the custom algorithm will remove all rows which match with at least one filter!

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -384,8 +384,8 @@
     showFullscreen: false,
     smartDisplay: true,
     escape: false,
-    filterOptions  : {
-      'filterAlgorithm': 'and' //and means all given filter must match, or means one of the given filter must match
+    filterOptions: {
+      'filterAlgorithm': 'and' // and means all given filter must match, or means one of the given filter must match
     },
     idField: undefined,
     selectItemName: 'btSelectItem',
@@ -1358,15 +1358,15 @@
         // Check filter
         if (typeof this.filterOptions.filterAlgorithm === 'function') {
           this.data = this.options.data.filter((item, i) => {
-            return this.filterOptions.filterAlgorithm.apply(null, [item, f]);
-          });
+            return this.filterOptions.filterAlgorithm.apply(null, [item, f])
+          })
         } else if (typeof this.filterOptions.filterAlgorithm === 'string') {
           this.data = f ? this.options.data.filter((item, i) => {
-            let filterAlgorithm = this.filterOptions.filterAlgorithm;
-            if(filterAlgorithm === 'and') {
+            const filterAlgorithm = this.filterOptions.filterAlgorithm
+            if (filterAlgorithm === 'and') {
               for (const key in f) {
                 if (
-                    (Array.isArray(f[key]) &&
+                  (Array.isArray(f[key]) &&
                         !f[key].includes(item[key])) ||
                     (!Array.isArray(f[key]) &&
                         item[key] !== f[key])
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1353,19 +1356,43 @@
...
-              (Array.isArray(f[key]) &&
-              !f[key].includes(item[key])) ||
-              (!Array.isArray(f[key]) &&
-              item[key] !== f[key])
-            ) {
-              return false
+        if (typeof this.filterOptions.filterAlgorithm === 'function') {
+          this.data = this.options.data.filter((item, i) => {
+            return this.filterOptions.filterAlgorithm.apply(null, [item, f]);
+          });
```
wenzhixin: need to fix the eslint problem.

src/bootstrap-table.js
```diff
@@ -1569,10 +1596,6 @@
           to = o.paginationSuccessivelySize
         }
 
-        if (to > this.totalPages) {
-          to = this.totalPages
-        }
-
```
wenzhixin: deleting these columns?

djhvscf:
Any update @Falseee

src/bootstrap-table.js
Falseee: Yes!
Check https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L1584-L1586 and https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L1596-L1598 code duplication.

Falseee:
@djhvscf i will update it today.

src/bootstrap-table.js
Falseee: done

charliehewis:
This has been massively helpful in working out how to use the custom filter algorithm, thankyou

UtechtDustin:
@charliehewis It is doumentated (https://bootstrap-table.com/docs/api/methods/#filterby), in my post above is als a jsfiddle which contains examples.
You just have to use/set the `filterAlgorithm` option.

charliehewis:
@UtechtDustin The documentation is not very helpful with respect to the custom filterAlgorithm and the example given there doesn't use it. I've spent a lot of time over the past few weeks with filters trying to get it to work and the example you gave here was the key to understanding it. I've also struggled with the persistence of the filter, clearing it and a few other things so I'd be happy to write some extra docs or examples that would help other people in my position

UtechtDustin:
@charliehewis Okay, could you please explain what do you try to achieve ?
Also please provide us an example (using our [editor](https://live.bootstrap-table.com/)) what do you already tried.

charliehewis:
@UtechtDustin I used the editor to make this example but I don't know where the save button on the editor is actually saving it to, or how to give you a link to what I wrote. I also can't get the right settings on jsfiddle. I've put it in a text file for now (sorry).

In my project I have Bootstrap tables on a series of webpages. The project is in typescript, using ReactJS but not react-bootstrap. I have found that when setting the filter on a table on one page, that same custom filter is used for a different table on a different page as if it is persisting even when it doesn't make sense to do so (because the data-fields have changed and the id of the new table is different). In order to fix this, I have had to include upon page load a method which updates the filter to clear it by explicitly returning to default $('#table').bootstrapTable('filterBy', {}, { 'filterAlgorithm': 'and'})

[JQuery_Bootstrap_Table_Filter.txt](https://github.com/wenzhixin/bootstrap-table/files/4687853/JQuery_Bootstrap_Table_Filter.txt)

UtechtDustin:
@charliehewis on the top is a save button, after you saved the url has be changed to your example url.

charliehewis:
@UtechtDustin Ah yes I see now, thanks. This is my example which demonstrates how you can use the custom filter with a key that isn't one of the data-fields, but it could be tidied up by having the 'choice' value be passed in as some property of the button clicked so that only one function is needed:
https://live.bootstrap-table.com/code/charliehewis/3169

Issue #3774, opened 2018-05-05
Column formatter  field parameter


elixom:
In the documentation for column options the formatter function lists 4 parameters. The fourth is "field" in the code version 1.12.1 only the first 3 are actually passed. The documentation also says it has 3 parameters even though the syntax lists 4.

For reference the "field" can be determined in the function by `this.field` but either the documentation or the code need to be updated.

Issue exists in all browsers on all environment.

wenzhixin:
Fixed

PR #4252, opened 2019-03-26
Added column field to formatters function

Falseee:
Added missing column field to formater function call (see #3774).


wenzhixin:
Thanks

PR #4257, opened 2019-03-28
add title attribute

2fps:
When the table has many columns and the displayed content is truncated, the detailed data can be displayed by hovering the mouse over the corresponding column。

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1935,6 +1935,8 @@
         }
         if (item[`_${field}_title`]) {
           title_ = Utils.sprintf(' title="%s"', item[`_${field}_title`])
+        } else {
+          title_ = Utils.sprintf(' title="%s"', item[ field ])
         }
         cellStyle = Utils.calculateObjectValue(this.header,
           this.header.cellStyles[j], [value_, item, i, field], cellStyle)


Falseee:
Can you provide us an jsfiddle ?

wenzhixin:
I think if all the items are added to the title, it may increase the interference: http://jsfiddle.net/wenyi/e3nk137y/44352/.

In general, if some columns are too long, you can use the formatter to solve the problem: http://jsfiddle.net/wenyi/e3nk137y/44356/.

What do you think? @djhvscf @Falseee 

Kapture 2019-03-29 at 8 35 40

djhvscf:
What about if this is added only if the cell has ellipsis?

djhvscf:
https://stackoverflow.com/a/10017343

Falseee:
I also think we should check if a ellipsis is active, but also this should be an optional option.

wenzhixin:
Consider implementing this feature as an extension.

Falseee:
https://github.com/wenzhixin/bootstrap-table/blob/develop/src/bootstrap-table.js#L1939-L1941
Isnt it already implemented by simply add a normal title attribute to the columns ?
@wenzhixin @djhvscf

wenzhixin:
@Falseee only transfer from html: https://examples.bootstrap-table.com/#welcomes/from-html.html

Falseee:
Sure, but we could check for an title key for json loaded tables.

djhvscf:
What about this column option https://bootstrap-table.com/docs/api/column-options/#titletooltip ?

djhvscf:
@2fps please create this as an extension

Issue #2628, opened 2016-09-08
Is there a way to use detail method 'expandRow/collapseRow' without icon 'detail-icon' shown?


gjzhappy:
My issue is as mentioned above ,I want to expand row in my java script manually,and I don't want to see the 'detail-icon' column,is there a solution?

stevespringett:
@gjzhappy did you ever find an answer to this? I want to do the same thing.

stevespringett:
@gjzhappy found an example here:
https://jsfiddle.net/dabros/6ony6cs2/1/

referenced in #964

tianyaxue:
@stevespringett this example i get it ,but it has problem when you detail view is a table,it will display none first td in detail table!

Falseee:
Could be a good feature, what do you think @wenzhixin @djhvscf ?

djhvscf:
Yes it is.

PR #4259, opened 2019-03-28
Added option to hide the show/hide icons for the detail view

Falseee:
Feature requested by @gjzhappy in issue #2628.

Before I could implement this feature, i had to change the way how we expand/collapse the detail view.
Before we used the find function to check if there is an icon and passed the icon to all functions (expandRow/collapseRow/expandAllRows/collapseAllRows), now i add a data attribute which allow us to check if the row has a detail view (rowIndex is now passed insteaf of the element of the icon).

I also documentated the toggleDetailView function

Demo: http://jsfiddle.net/20qho9dw/3/

### DIFF
--- a/site/docs/api/methods.md
+++ b/site/docs/api/methods.md
@@ -475,6 +475,14 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
   Toggle the card/table view.
 
+## toggleDetailView
+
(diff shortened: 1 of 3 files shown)


wenzhixin:
Updated `showDetailView` to `detailViewIcon`.

wenzhixin:
Thanks for your contribution! @Falseee

PR #4273, opened 2019-04-05
Fix/4097

Falseee:
This PR will change the parameters of the (un)checkAll events:

-It will pass as first parameter for the uncheckall event an array of the rows AFTER the event was triggered (requested by qigechuizi in issue #4097).
- Also i added for both events as second parameter the rows before the event was triggered.

Demo: http://jsfiddle.net/v1do4a3t/4/


wenzhixin:
Thanks for your contribution!

PR #4288, opened 2019-04-11
Fix issue 4072

### LINKED ISSUE #4072: Checkbox and radio auto-checked
When we use data ajax-requested, checkboxes and radios work well, following the official jsfiddle example.
But when we write the table directly in html, all the checkboxes / radios are checked.
I use bootstrap-table 1.12.2 on Chrome, OS X El Capitan.

[jsfiddle](http://jsfiddle.net/dvkefsxm/2/)

UtechtDustin:
Checks if the formatted value is not false, to fix #4072

Fix jsfiddle of the issue: http://jsfiddle.net/t9xmksfL/2/

### DIFF
--- a/site/docs/api/column-options.md
+++ b/site/docs/api/column-options.md
@@ -180,6 +180,9 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 - **Detail:**
 
   Set `true` to show a radio. The radio column has fixed width.
+  
+  If a value is given the Checkbox is automatically checked.  
+    Its also possible to check/uncheck the radio by use an formatter (return `true` to check, return `false` to uncheck).
 
 - **Default:** `false`
 
@@ -192,6 +195,9 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 - **Detail:**
 
   Set `true` to show a checkbox. The checkbox column has fixed width.
+  
+  If a value is given the Checkbox is automatically checked.  
+  Its also possible to check/uncheck the checkbox by use an formatter (return `true` to check, return `false` to uncheck). 
 
 - **Default:** `false`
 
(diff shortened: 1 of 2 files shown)


vincentwinkel:
Thanks for taking this issue in consideration, but not sure to understand the new format.
Before, `formatter` was used to format a cell (if we want to add icons, etc) AND to check a cell.
Because now, it only returns true/false to (un)check a cell, how can we format a cell?
I also think the word `formatter` is not appopriate for its goal and should be probably replaced by `selected`.

UtechtDustin:
The formatter works like before, but if you use a formatter on a checkbox/radiobutton you can check/uncheck it (return true/false).

vincentwinkel:
Except for the layout (no CSS added), I don't understand between the file `https://rawgit.com/wenzhixin/bootstrap-table/8be580deb3024dc93d2fa2989e940501f764d018/src/bootstrap-table.js` that you use in your fiddle and `https://unpkg.com/bootstrap-table@1.14.2/dist/bootstrap-table.min.js` that I take from the examples page.
They are both v1.14.2 but the first one works, not the second one.
Here is the HTML generated by my framework: http://jsfiddle.net/y7tcajux/ (you can (toggle comments at the top to switch files)

Screen Shot 2019-04-23 at 19 19 34

Screen Shot 2019-04-23 at 19 19 56

UtechtDustin:
@vincentwinkel 
The first one was from my commit/Pull Request, its merged in the develop branch.
So its not 1.14.2 its after that, in the next version its fixed :)

vincentwinkel:
Alright. So, according to the changelog, lets wait the version 2...
Weird to wait for a big update since it's a bug and not a request. Hopefully I don't use (yet) bootstrap-table on production :/

wenzhixin:
Consider picking bug commits and release a new minor version.

Issue #4033, opened 2018-11-01
Cookie + Filter Control lost select option on refresh browser


manukieli:
I'm using:
bootstrap-table.js v1.12.1
bootstrap-table-filter-control.js v2.1.3
bootstrap-table-cookie v.1.2.3

I've the cookie active for the filter control and some field are using data-filter-control="select"
When I filter with the select, the filter is working well, but when I refresh the browser, the filter is applied correctly with the cookie data. But the select field is not showing data.

![select lost](https://user-images.githubusercontent.com/14294794/47853192-3535da00-ddde-11e8-8a8e-ac7273c1c663.gif)

To resolve this issue, to the file bootstrap-table-cookie.js, I've replace the line 185-187 with this code
`                            if (cookie.text !== '') {
                                $(element).val(cookie.text);
                                cachedFilters[cookie.field] = cookie.text;
                            }`

to this lines:
`                            if (cookie.text !== '' && $(element).is('input')) {
                                $(element).val(cookie.text);
                                cachedFilters[cookie.field] = cookie.text;
                            }else if(cookie.text !== '' && $(element).is('select')){
                                $(element).append($('<option>', {value: cookie.text, text: cookie.text}));
                                $(element).val(cookie.text);
                                cachedFilters[cookie.field] = cookie.text;
                            }`

And now is working well.
View the result:
![select not lost](https://user-images.githubusercontent.com/14294794/47853306-9b226180-ddde-11e8-9572-1775ae319830.gif)

To this if: `if(cookie.text !== '' && $(element).is('select'))`
If I apply this `$(element).val(cookie.text);` and return it to console.log, the value is null.

jimgit:
What tool are you using for the screen video?  Very helpful.

manukieli:
I'm using Camtasia and save to Gif.

manukieli:
This issue persist with bootstrap-table version 1.13.5
And I resolve it with this same solution describe in the first post.

wenzhixin:
PR is welcome.

UtechtDustin:
#2624

steinbachio:
This is neither duplicate nor is it resolved.
The Solution of manukieli is working.
I am working with `data-filter-data="url:/foo/bar/values"`

Issue #2624, opened 2016-09-07
cookie extension does not work with filterControl extension properly


hrigu:
I have a combination of the filterControl and the cookie extension. I think this combination does not work together:
1) When reloading the page, the filter string is correctly shown in the filter input field, but the rows aren't filtered accordingly.
The pagination seems to work fine with the cookie extension, as well sorting the columns, 

2) when I have the pagination set in my cookie, and I make a filter, then the filter is displayed and the pagination is lost: correkt.
After reloading the page, the old page is displayed, the filter input as well, but the rows aren't filtered.

See http://jsfiddle.net/e3nk137y/8489/ for a demonstration

Christian

Falseee:
Working demo: http://jsfiddle.net/vxjac65z/

UtechtDustin:
Works fine in the latest version, please update !
Demo: https://live.bootstrap-table.com/code/UtechtDustin/43

PR #4294, opened 2019-04-13
Fix issue #4033

manukieli:
This path resolve the issue when you refresh browser and use Cookie + Filter Control extension and have a filter with select element.

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -173,10 +173,17 @@
 
           const applyCookieFilters = (element, filteredCookies) => {
             $(filteredCookies).each((i, cookie) => {
-              if (cookie.text !== '') {
-                $(element).val(cookie.text)
-                cachedFilters[cookie.field] = cookie.text
-              }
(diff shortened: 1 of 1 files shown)


djhvscf:
Do you have an example?

manukieli:
I've update to the version 1.14.2 and it seems now it work.

wenzhixin:
This branch has conflicts

manukieli:
I've create an example to Fiddle and I get another problem with the verison 1.14.2
I'm using FilterControl and Cookie.
I've set the CookieStorage like "localStorage".
When I filter the column Item Price, the filter work well.
But when I refresh the page, I get all data into table, but the filter field, appear with the last filter data before refresh and the filter is not apply.
You can see an example here: https://jsfiddle.net/pj8cktse/4/

UtechtDustin:
Your description sounds like #2624

djhvscf:
It is the same description @UtechtDustin

djhvscf:
After a quick review, I'm closing this pull request since it won't solve the issue between filter control and cookie extensions.

manukieli:
@djhvscf, can you resolve the issues???

PR #4296, opened 2019-04-14
Adding filterOrderBy column option

### LINKED ISSUE #4149: bootstrap-table-filter-control should not order alphabetically by default
I'm using 1.13.3 (`Bootstrap-Table` and `bootstrap-table-filter-control` extension) taking the select values from a variable:

```
var viviendas = {
     "31":"Bbbbbb","32":"Aaaaa"
};
```

The column in the table is like this:

```
data-filter-control="select" data-filter-data="var:viviendas"
```

The bootstrap-table-filter-control extension is ordering the select values, showing `Aaaaa` first and then `Bbbbbb`. The extension should have a parameter to set this up and not order alphabetically by default.

djhvscf:
Fix #4149 

jsFiddle: http://jsfiddle.net/djhvscf/94ozkfhs/3/

### DIFF
--- a/site/docs/extensions/filter-control.md
+++ b/site/docs/extensions/filter-control.md
@@ -142,7 +142,7 @@ Dependence if you use the datepicker option: [bootstrap-datepicker](https://gith
 
 - **Detail:**
 
-   Set this in order to order the options in a select control whether ascending or descending.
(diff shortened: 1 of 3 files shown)


djhvscf:
Closing and added in a future PR

PR #4309, opened 2019-04-20
Resolve the issue Cookie + Filter Control lost select option on refresh browser #4033

manukieli:
They resolve this issue


UtechtDustin:
A Demo of the fix would be great (jsfiddle), thank you for very much the contribution.

UtechtDustin:
Tested local, don't work for me, it just add a new option with the last selected item (duplicate of the item) and select it.
But the Table don't filter.

Please update your PR.

manukieli:
For me is working well.

djhvscf:
Testing it

src/extensions/cookie/bootstrap-table-cookie.js
```diff
@@ -172,10 +172,17 @@ const UtilsCookie = {
 
         const applyCookieFilters = (element, filteredCookies) => {
           $(filteredCookies).each((i, cookie) => {
-            if (cookie.text !== '') {
-              $(element).val(cookie.text)
-              cachedFilters[cookie.field] = cookie.text
-            }
+              if (cookie.text !== '' && $(element).is('input')) {
```
djhvscf: Save the `$(element)` in a const variable, so you can use it in following lines

djhvscf:
Please consider to retest this solution with the latest version of filter-control extension

PR #4311, opened 2019-04-22
prevent overwriting the searchText

### LINKED ISSUE #2776: Filter Control overwrites "search" parameter in API call

UtechtDustin:
Fixes #2776

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -773,16 +773,9 @@ $.BootstrapTable = class extends $.BootstrapTable {
       delete this.filterColumnsPartial[$field]
     }
 
-    // if the searchText is the same as the previously selected column value,
-    // bootstrapTable will not try searching again (even though the selected column
-    // may be different from the previous search).  As a work around
-    // we're manually appending some text to bootrap's searchText field
-    // to guarantee that it will perform a search again when we call this.onSearch(event)
-    this.searchText += 'randomText'
-
     this.options.pageNumber = 1
     this.EnableControls(false)
-    this.onSearch(event)
+    this.onSearch()
     this.trigger('column-search', $field, text)
   }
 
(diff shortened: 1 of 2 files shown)


src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -773,16 +773,9 @@ $.BootstrapTable = class extends $.BootstrapTable {
...
-    // if the searchText is the same as the previously selected column value,
-    // bootstrapTable will not try searching again (even though the selected column
-    // may be different from the previous search).  As a work around
-    // we're manually appending some text to bootrap's searchText field
-    // to guarantee that it will perform a search again when we call this.onSearch(event)
-    this.searchText += 'randomText'
-
     this.options.pageNumber = 1
     this.EnableControls(false)
-    this.onSearch(event)
```
djhvscf: This [PR](https://github.com/wenzhixin/bootstrap-table/pull/4309) is not working since this the event parameter was deleted. @UtechtDustin could you, please, review this and let us know why are you removing this parameter?

UtechtDustin: Checkout the changes in `src/bootstrap-table.js`, i added a check which only overwrite the search text if a event (currentTarget) is given.

The overwrite makes only sense for the "main" search input, for other filter we have the `this.filterColumnsPartial` variable.

UtechtDustin: Anything else will cause the issue #2776

djhvscf: Got it

Issue #4334, opened 2019-04-29
Implement a version number


tangram67:
Hi to all.

Is it possible to implement a version number like this:

```
--- bootstrap-table.js	2019-03-18 17:13:02.000000000 +0100
+++ bootstrap-table-versioned.js	2019-04-29 20:47:31.865557373 +0200
@@ -498,6 +498,7 @@
 
     // BOOTSTRAP TABLE CLASS DEFINITION
     // ======================
+    var VERSION = '1.14.2';
 
     var DEFAULTS = {
       height: undefined,
@@ -3709,6 +3710,7 @@
     BootstrapTable.LOCALES = LOCALES;
     BootstrapTable.COLUMN_DEFAULTS = COLUMN_DEFAULTS;
     BootstrapTable.EVENTS = EVENTS;
+    BootstrapTable.VERSION = VERSION;
 
     // BOOTSTRAP TABLE PLUGIN DEFINITION
     // =======================
@@ -3757,6 +3759,7 @@
     $.fn.bootstrapTable.defaults = BootstrapTable.DEFAULTS;
     $.fn.bootstrapTable.columnDefaults = BootstrapTable.COLUMN_DEFAULTS;
     $.fn.bootstrapTable.locales = BootstrapTable.LOCALES;
+    $.fn.bootstrapTable.version = BootstrapTable.VERSION;
     $.fn.bootstrapTable.methods = allowedMethods;
     $.fn.bootstrapTable.utils = Utils;
 
@@ -3767,4 +3770,4 @@
       $('[data-toggle="table"]').bootstrapTable();
     });
   })(jQuery);
-});
\ Kein Zeilenumbruch am Dateiende.
+});
```
So one can use the following code to get the version string:

```
function getBootstrapTableVersion() {
	return $.fn.bootstrapTable.version;
};
```
Thanks in advance,
Dirk.

UtechtDustin:
Sure its possible, could you explain for what you need the version ?

Möglich ist es natürlich, kannst du uns erklären wofür genau diese Funktion brauchst ?

UtechtDustin:
Makes sense i guess, also other populare Libs has the version as constant/getter.
Pull Requests are welcome :-)
****
Macht Sinn, sollte implementiert werden. Haben auch einige andere beliebte Libs.
Kannst gerne ein Pull Request stellen.

tangram67:
[...]

PR #4336, opened 2019-04-30
Implement a version number

tangram67:
As mentioned in 

#4334 

I setup a pull request for the version number.


src/constants/index.js
```diff
@@ -1,3 +1,5 @@
+const VERSION = '1.14.2';
```
wenzhixin: Do not need `;`

PR #4380, opened 2019-06-04
Added screen reader support for pagination

### LINKED ISSUE #4362: Screenreader support pagination
Hi,
I use the table plugin together with Bootstrap 4 and the results are really impressive :)
But, websites have to be barrier-free these days and I have some stomach ache here when paginating tables.

The button to choose how many rows to show doesn't seem to be a real drop-down. At least Jaws (popular screenreader) can't explain the handling.

But more important is the control to jump between the paginated pages. A title attribute or aria-label would have to be added.
Currently Jaws reads the following: Link: 1, link: 2, Link: 3...
Much more helpful would be: Link: to page 1, Link: to page 2, Link: to page 3...

The whole thing is certainly only a small thing, and for someone who knows how to implement it easily.
I've already tried to adapt the code accordingly, but honestly I can't get an overview.

It would be cool if someone could help.

UtechtDustin:
fix #4362

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -918,7 +918,7 @@ class BootstrapTable {
 
       html.push(`<div class="${this.constants.classes.pull}-${o.paginationHAlign} pagination">`,
         Utils.sprintf(this.constants.html.pagination[0], Utils.sprintf(' pagination-%s', o.iconSize)),
-        Utils.sprintf(this.constants.html.paginationItem, ' page-pre', o.paginationPreText))
+        Utils.sprintf(this.constants.html.paginationItem, ' page-pre', o.formatSRPaginationPreText(false, 'zzzz'), o.paginationPreText))
 
       if (this.totalPages < o.paginationSuccessivelySize) {
         from = 1
@@ -946,7 +946,7 @@ class BootstrapTable {
 
       const middleSize = Math.round(o.paginationPagesBySide / 2)
       const pageItem = (i, classes = '') => Utils.sprintf(this.constants.html.paginationItem,
-        classes + (i === o.pageNumber ? ` ${this.constants.classes.paginationActive}` : ''), i)
+        classes + (i === o.pageNumber ? ` ${this.constants.classes.paginationActive}` : ''), o.formatSRPaginationPageText(i), i)
 
       if (from > 1) {
         let max = o.paginationPagesBySide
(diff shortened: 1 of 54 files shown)


djhvscf:
@UtechtDustin do you have any example of this pull request?

UtechtDustin:
If you need a example i can create one, but its only add the `aria-label` with the text `to page $page` (based on the selected trans).. same for the next and back button.

KaptnDieter:
Thanks a lot dude! @UtechtDustin 🙏

UtechtDustin:
force pushed same commit for trigger the ci

src/bootstrap-table.js
```diff
@@ -919,7 +919,7 @@ class BootstrapTable {
 
       html.push(`<div class="${this.constants.classes.pull}-${o.paginationHAlign} pagination">`,
         Utils.sprintf(this.constants.html.pagination[0], Utils.sprintf(' pagination-%s', o.iconSize)),
-        Utils.sprintf(this.constants.html.paginationItem, ' page-pre', o.paginationPreText))
+        Utils.sprintf(this.constants.html.paginationItem, ' page-pre', o.formatSRPaginationPreText(false, 'zzzz'), o.paginationPreText))
```
wenzhixin: What is the purpose of `false, 'zzzz'` parameters?

src/constants/index.js
```diff
@@ -50,7 +50,7 @@ const CONSTANTS = {
       pageDropdownItem: '<li role="menuitem" class="%s"><a href="#">%s</a></li>',
       dropdownCaret: '<span class="caret"></span>',
       pagination: ['<ul class="pagination%s">', '</ul>'],
-      paginationItem: '<li class="page-item%s"><a class="page-link" href="#">%s</a></li>',
+      paginationItem: '<li class="page-item%s"><a class="page-link" aria-label="%s" href="javascript:void(0)">%s</a></li>',
```
wenzhixin: @UtechtDustin Please help to update other themes, thanks.

src/bootstrap-table.js
UtechtDustin: Im not sure right now (i dont have the IDE open), i will check it later.

wenzhixin:
Updated the zh-CN and zh-TW locales.
`git rebase develop` and push force.
You need to `git pull -rebase` @UtechtDustin

src/constants/index.js
UtechtDustin: added

src/bootstrap-table.js
UtechtDustin: i think it was an rebase issue...
i removed it!

PR #4383, opened 2019-06-04
fixed on click event for paginationVAlign 'both'

### LINKED ISSUE #4352: showJumpto failing for bottom controls with paginationVAlign:'both'
With paginationVAlign set to 'both', two sets of textbox + button are added within a 'li' with class 'jumpto' on the top and bottom. However, the bottom set doesn't work. I've found the reason that it doesn't work and implemented a temporary fix with minimal changes to the page-jump-to library for my own purposes but would love if a permanent solution can be found.

Issue lies here: 
Lines 38-40 of bootstrap-table-page-jump-to.js

```javascript
$jumpTo.find('button').click(() => {
    this.selectPage(+$jumpTo.find('input').val())
})
```

Now `$jumpTo.find('input')` always refers to the input on the top no matter which button I click. The problem is they both don't have any specific selectors to select their 'companion' input so the one that ends up getting picked by `find().val()` is always the text in the top input. So nothing happens on clicking the bottom 'GO' button since `.find().val()` returns an empty string.

My temporary fix involves using some pure javascript instead of jquery so I don't like it very much:

```javascript
$jumpTo.find('button').click(() => {
    this.selectPage(+($($jumpTo.find('input')[0]).val() || $($jumpTo.find('input')[1]).val()))
})
```

I don't think a fiddle link is required here except to demonstrate that it does not work as the issue is pretty obvious(I'm a jquery noob, but still found it pretty quickly) but here's one anyway: [jsFiddle](https://jsfiddle.net/om8qzk72/)

Edit:
Also, my quick fix has its own quirks. If I enter the page number in the top input and then (without hitting 'GO') go to the bottom and enter the page number in the bottom input (maybe because I changed my mind?!), then even though I clicked the bottom 'GO' button, I'll be navigated to the page number entered in the top input. I guess it is pretty obvious why I don't like my fix.

UtechtDustin:
Fix #4352

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -4,7 +4,7 @@ let bootstrapVersion = 4
 try {
   const rawVersion = $.fn.dropdown.Constructor.VERSION
 
-  // Only try to parse VERSION if is is defined.
+  // Only try to parse VERSION if it is defined.
   // It is undefined in older versions of Bootstrap (tested with 3.1.1).
   if (rawVersion !== undefined) {
     bootstrapVersion = parseInt(rawVersion, 10)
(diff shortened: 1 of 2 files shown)


wenzhixin:
Thanks!

PR #4387, opened 2019-06-06
fixed scrolling in ie11

### LINKED ISSUE #4167: showColumns cannot scroll down the sidebar in IE11
By creating a bootstrap table with `showColumns: true` table option, user cannot scroll down the sidebar in Internet explorer 11 browser.

Try to scroll by dragging the bar, it won't work. Only by clicking the arrows or using mouse wheel.

UtechtDustin:
Fix #4167

### DIFF
--- a/src/themes/theme.scss
+++ b/src/themes/theme.scss
@@ -37,7 +37,7 @@
       .dropdown-menu {
         text-align: left;
         max-height: 300px;
-        overflow: auto;
+        overflow-y: scroll;
         z-index: 1001;
       }
 
--- a/src/themes/theme.scss
+++ b/src/themes/theme.scss
@@ -38,7 +38,7 @@
         text-align: left;
         max-height: 300px;
         overflow: auto;
-        -ms-overflow-style:scrollbar;
+        -ms-overflow-style: scrollbar;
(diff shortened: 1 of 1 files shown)


wenzhixin:
I think it's not a good way to solve the problem because it will always display the scrollbar (for example in Chrome):
image

Maybe we can solve it with this answer:
https://stackoverflow.com/questions/45890061/overflow-y-scroll-not-working-in-ie-11

UtechtDustin:
If the scrollbar is visible in chrome we have a overflow issue or not ?
I will check it later!

UtechtDustin:
reverted the commit and used the bugfix from stackoverflow.
Tested in ie11, firefox and chrome.

src/themes/theme.scss
```diff
@@ -38,6 +38,7 @@
         text-align: left;
         max-height: 300px;
         overflow: auto;
+        -ms-overflow-style:scrollbar;
```
wenzhixin: Add an space ` ` after `:`
```
-ms-overflow-style: scrollbar;
```

UtechtDustin: done

PR #4388, opened 2019-06-06
Added options to show/hide icons/text on buttons

### LINKED ISSUE #4324: The Columns button tag has the wrong title text
### The Columns button tag has the wrong title text

A summary of the issue and the browser/OS environment in which it occurs. If suitable, include the steps required to reproduce the bug.

All browsers. View source on the columns and see that the columns button (inside the div for the dropdown) has the title text as "Fullscreen". This should say "Columns". It looks like a copypasta mistake where you just included the wrong text formatter method in the title text of the button.

Bonus feature:

It would also be nice to have an option to show the button text on the toolbar buttons. Our UI/UX designer says it's always helpful to have text with an icon as icons aren't always easy to deduce their meaning.

UtechtDustin:
fix #4324

I added two options (`showButtonIcons` and `showButtonText`) to show/hide icons/text on Buttons.

Demo:
Text: http://jsfiddle.net/hsab569k/3/
Icons: http://jsfiddle.net/hsab569k/4/

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -54,8 +54,8 @@ const CONSTANTS = {
       icon: '<i class="%s %s"></i>',
       inputGroup: '<div class="input-group">%s<span class="input-group-btn">%s</span></div>',
       searchInput: '<input class="%s%s" type="text" placeholder="%s">',
-      searchButton: '<button class="btn btn-default" type="button" name="search" title="%s"><i class="%s %s"></i></button>',
-      searchClearButton: '<button class="btn btn-default" type="button" name="clearSearch" title="%s"><i class="%s %s"></i></button>'
+      searchButton: '<button class="btn btn-default" type="button" name="search" title="%s">%s %s</button>',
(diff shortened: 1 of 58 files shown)


src/constants/index.js
```diff
@@ -54,8 +54,8 @@ const CONSTANTS = {
       icon: '<i class="%s %s"></i>',
       inputGroup: '<div class="input-group">%s<span class="input-group-btn">%s</span></div>',
       searchInput: '<input class="%s%s" type="text" placeholder="%s">',
-      searchButton: '<button class="btn btn-default" type="button" name="search" title="%s"><i class="%s %s"></i></button>',
-      searchClearButton: '<button class="btn btn-default" type="button" name="clearSearch" title="%s"><i class="%s %s"></i></button>'
+      searchButton: '<button class="btn btn-default" type="button" name="search" title="%s">%s %s</button>',
+      searchClearButton: '<button class="btn btn-default" type="button" name="clearSearch" title="%s">%s %s</button>'
```
wenzhixin: Need to update other themes.

UtechtDustin: added

PR #4390, opened 2019-06-07
Added stylelint

wenzhixin:
Added stylelint and fixed the css error.


wenzhixin:
Updated the branch name: #4391

PR #4391, opened 2019-06-07
Added stylelint

wenzhixin:
Added stylelint and fixed the css error.


package.json
```diff
@@ -24,10 +24,14 @@
...
     "sass": "^1.16.1",
+    "stylelint": "^10.1.0",
+    "stylelint-config-standard": "^18.3.0",
     "vue-template-compiler": "^2.6.10"
   },
   "scripts": {
-    "lint": "eslint src",
+    "lint:js": "eslint src",
+    "lint:css": "stylelint src/**/*.scss",
+    "lint": "run-s lint:*",
```
UtechtDustin: Looks good at all, but could you also add a script which execute `lint:*` and our check tools ?

UtechtDustin:
Added Scripts to check tools and a script which execute the linters and the tools.

PR #4392, opened 2019-06-07
fixed bug with json object

### LINKED ISSUE #4375: filtering does not work json sub-object
Hi,

If my `data-field` is a top level item e.g. `<th data-field='x'>` the filtering works. If my item happens to be a sub-object (for example) `<th data-field='x.y'>` filtering does **not** work.

I have tried it by loading by Ajax and using `append` and also using `data-url`.

UtechtDustin:
fix #4375

This will fix the filtering with the `filter-control` extension with json objects, and also fix the `uniqueId` table option with json objects.

Demo: https://jsfiddle.net/b4yLa6hk/4/


wenzhixin:
Thanks!

PR #4394, opened 2019-06-08
WIP:test

UtechtDustin:


UtechtDustin:
@wenzhixin 
Something is wrong with the ci checks, deleted the whole documentation of the column-options and the check is passed.
You can also see the errors in the ci report, i think you have to use the exit code 1 if errors are found.

PR #4397, opened 2019-06-08
Use the sync function instead of the async function

UtechtDustin:


tools/check-api.js
```diff
@@ -53,8 +53,7 @@ class API {
         console.log(error)
       })
     }
-
-    fs.writeFile(file, outLines.join('## '), () => {})
+    fs.writeFileSync(file, outLines.join('## '), () => {})
```
wenzhixin: We can remove the callback: `fs.writeFileSync(file, outLines.join('## '))`

UtechtDustin: done

PR #4399, opened 2019-06-09
go through the filtered data not through the whole data

### LINKED ISSUE #4002: Preventing hidden selections from counting as selections
I am trying to return all visible selections. If something is "checked" but is not visible because a search filter has been applied, it will still return the hidden selection.

This functionality happens with both the below functions. From reading the description, I understood it should only happen with 'getAllSelections.'  

http://issues.wenzhixin.net.cn/bootstrap-table/#methods/getSelections.html
http://issues.wenzhixin.net.cn/bootstrap-table/#methods/getAllSelections.html

Perhaps that is not the case. Even so, I tried to bypass that by using this:

`$('#data-table').bootstrapTable({
        onSearch: function () {
            $('#data-table').bootstrapTable('uncheckAll');
        }
    });`

But it doesn't work most of the time. I think what is happening is it will only uncheck the rowif the rowis visible. So if you apply a filter, then the hidden row is still selected.

I cannot tell if these interactions are intended or bugs. The description of the functions has me a little confused, sorry.

UtechtDustin:
fix #4002

In my opinion we should add this as option and also add the options from the `getData` function (only currentPage, include hidden rows).
What do you think ?

Demo: http://jsfiddle.net/nxjbmu23/5/
Check the first checkbox, now the method (button) will return one row.
Search for 13 and check the method again, it will show you 0 rows because the other data is filtered out.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1952,12 +1952,11 @@ class BootstrapTable {
 
   getSelections () {
     // fix #2424: from html with checkbox
-    return this.options.data.filter(row =>
-      row[this.header.stateField] === true)
+    return this.data.filter(row => row[this.header.stateField] === true)
   }
 
   getAllSelections () {
-    return this.options.data.filter(row => row[this.header.stateField])
+    return this.data.filter(row => row[this.header.stateField])
   }
 
   load (_data) {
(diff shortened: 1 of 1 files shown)


djhvscf:
This is ok for me. but @wenzhixin please review this PR

wenzhixin:
Updated and merged.

PR #4417, opened 2019-06-14
Fix search text overwritting

### LINKED ISSUE #2203: Filter-control and server side pagination
A have server side pagination grid with search control and I wan't to add filter control working on client-side (filter only what is visible). It's possible?
Now when add some text to input filter column, there is server side refresh and entered text is sended same as search text.

UtechtDustin:
Only set the search text if it was triggered by the original search input to prevent overwriting the search text (e.g. for the filter-control extension).

fix #2203 
Demo: http://jsfiddle.net/6nt0wxpy/4/

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -631,7 +631,7 @@ class BootstrapTable {
         </div>
       `,
       Utils.sprintf(this.constants.html.inputGroup,
-        `<input class="${this.constants.classes.input}${Utils.sprintf(' input-%s', o.iconSize)}" type="text" placeholder="${o.formatSearch()}">`,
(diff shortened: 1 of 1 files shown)


src/bootstrap-table.js
```diff
@@ -685,8 +685,10 @@ class BootstrapTable {
         return
       }
 
-      this.searchText = text
-      this.options.searchText = text
+      if ($(currentTarget).hasClass('search-input')) {
```
djhvscf: this.options.search is not an option here?

UtechtDustin: No is not an option, because i have to check if the input which triggered the search is an input for the filter-control extension or the normal search input.
With `this.options.search` i only know if the search is enabled or not.

PR #4439, opened 2019-06-22
Added virtual scroll options and set false by default

wenzhixin:
* Added `virtualScroll` option: https://live.bootstrap-table.com/code/wenzhixin/58

* Added `virtualScrollItemHeight` option:
Before: https://live.bootstrap-table.com/code/wenzhixin/59
After: https://live.bootstrap-table.com/code/wenzhixin/60


UtechtDustin:
Maybe i'm blind, but i cant see a difference on the demos..
Could you explain the difference ?

wenzhixin:
You can scroll to the bottom.

UtechtDustin:
Oh okay, i just scrolled some pixel up and down.

PR #4441, opened 2019-06-22
init the toolbar to refresh the buttons e.g. showColumns

### LINKED ISSUE #2820: Reordering columns and hide/showing through the data-columns=true hides the wrong table heading
Reordering columns and hide/showing through the data-columns=true hides the wrong table heading

**This can be reproduced on your demo:** 
http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/reorder-columns.html
(Any browser, but I tested on chrome)

**Setup on the code side: data-show-columns="true", data-reorderable-columns="true"**

Step 1 - Drag/Reorder 'Item Name' column one space to the left. The order will now be 'Item Name', 'ID', 'Item Price'.

Step 2 - Click on the dropdown-toggle and uncheck 'Item Name', to HIDE this column. The result will be the heading 'ID' will disappear, but the 'Item Name' heading will still display. Than the columns that belonged to Item Name in the table-body will disappear.

Step 3 - Click on the dropdown-toggle and uncheck 'Item Name', to SHOW this column. The result will be the first column displayed will have the heading 'Item Name', but the data will be the data for ID. The second column will have the heading 'ID', but have the data for Items Name.

Following these steps the expected result would be 'Item Name' heading and data to be in the first column, than 'ID' heading and data in the next column.

(https://github.com/wenzhixin/bootstrap-table/issues/1765)).

I noticed this bug on my companies application in all 3 major browsers. When I was able to reproduce on the demo you guys provide, I realized the bug must either be in the reorder-columns extension or the part of the base code that defines logic for data-columns=true.

Thank you guys for investigating this issue.

UtechtDustin:
fix #2820 

Before: https://live.bootstrap-table.com/example/extensions/reorder-columns.html
After: https://live.bootstrap-table.com/code/UtechtDustin/65

### DIFF
--- a/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
+++ b/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
@@ -166,6 +166,7 @@ BootstrapTable.prototype.makeRowsReorderable = function () {
       that.header.fields = ths
       that.header.formatters = formatters
       that.initHeader()
+      that.initToolbar()
       that.initBody()
       that.resetView()
       that.trigger('reorder-column', ths)


wenzhixin:
Can we solve the problem?

djhvscf:
We can extract the creation of showColumns

PR #4456, opened 2019-06-26
Fixed colspan group header display bug

### LINKED ISSUE #2051: Apparent bug using colspan to form column groups
Hi guys, excellent library...however, I'm having an issue making column groups with `colspan` attribute

Here is a fiddle, http://jsfiddle.net/tdow66ef/1/

If you see the code, the second column of `group a`, named `a-two`, is hidden. But the first column of group b is taking it's place. If you toggle `a-two`'s visibility, it returns to it's normal, desired position.

### LINKED ISSUE #3183: The colspan should be update and the 'thead' should be able to be hidden
https://live.bootstrap-table.com/code/wenzhixin/102

1.hide column "stars"
2.hide column "forkes"
3.You can find the problem

I tried set visible for "thead" to fix this problem, but not work, because "thead" not support hidden.
Function "initHeader" adding "thead" to HTML will never change，"BootstrapTable.prototype.initHeader" the code reads as follows:
```JavaScript
                //should  added ->  if(column.visible){
                    html.push('<th' + sprintf(' title="%s"', column.titleTooltip),
                        column.checkbox || column.radio ?
                            sprintf(' class="bs-checkbox %s"', column['class'] || '') :
                            class_,
                        sprintf(' style="%s"', halign + style),
                        sprintf(' rowspan="%s"', column.rowspan),
                        sprintf(' colspan="%s"', column.colspan),
                        sprintf(' data-field="%s"', column.field),
                        "tabindex='0'",
                        '>');
               // }
```
The same issues as
https://github.com/wenzhixin/bootstrap-table/issues/2637
https://github.com/wenzhixin/bootstrap-table/issues/3054

wenzhixin:
* **Fix #3183**

Before: https://live.bootstrap-table.com/code/wenzhixin/102
After: https://live.bootstrap-table.com/code/wenzhixin/105

Steps:  hide 'Stars' and 'Forkes' columns

* **Fix #2051**

Before: https://live.bootstrap-table.com/code/wenzhixin/106
After: https://live.bootstrap-table.com/code/wenzhixin/107

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -54,12 +54,16 @@ export default {
         const colspan = r.colspan || 1
         const index = flag[i].indexOf(false)
 
+        r.colspanIndex = index
+
         if (colspan === 1) {
           r.fieldIndex = index
           // when field is undefined, use index instead
           if (typeof r.field === 'undefined') {
             r.field = index
           }
+        } else {
+          r.colspanGroup = r.colspan
         }
 
         for (let k = 0; k < rowspan; k++) {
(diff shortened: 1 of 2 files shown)


src/utils/index.js
```diff
@@ -72,6 +76,26 @@ export default {
     }
   },
 
+  updateFieldGroup (columns) {
+    const allColumns = [].concat(...columns)
```
UtechtDustin: conact a empty array ?

wenzhixin: Because the columns is a two-dimensional array, so here need to convert it to one dimension, and then we can use the `find` method.
We can use `[].concat` method to do this (ES6), for example:
```
const arr = [[1], [3]]
[].concat(...arr)
// result: [1, 3]
```

wenzhixin: I found that we can use `flat` function instead of `contact`, and it's better to understand.

PR #4522, opened 2019-07-31
colgroup

AmbroseCdMeng:
sometimes, the width attributes needs to use with table-layout:fixed
but, if the table has 'colspan' or 'rowspan', The width will fail.
so we can set the width by 'colgroup',
like this
<colgroup width="100vw" ></colgroup>


wenzhixin:
Thanks for your contribution!
Don't change the code style, and provide an [Online Example](https://live.bootstrap-table.com) to show your problem, thanks!

wenzhixin:
Please create a new PR, thanks!

PR #4557, opened 2019-08-26
Set a simple button (no dropdown) when there is only one export option defined

rafaeljusi:
Set a simple button (no dropdown) when there is only one export option defined


wenzhixin:
Example: https://live.bootstrap-table.com/code/wenzhixin/511
There is still a small problem on the button display.

wenzhixin:
Improved this feature based on your commits in #4563, thanks!

Issue #1899, opened 2016-01-17
Do not collapse a detail view after table initialization or refresh (Feature request)


monchoz:
Please include a feature kind of like a param in `refresh` method that prevents closing or collapsing rows that are expanded. 

I use this workaround https://github.com/wenzhixin/bootstrap-table/issues/1603 and it's perfect in some cases like when you need to update a specific row and the action is triggered by a button inside the detail view (for instance).

Regards!

dabros:
similar to #1498, and seems an interesting feature to me as poor performance in some circumstances if full refresh for single row update and ect.

wenzhixin:
I think they are not the same problem.

wenzhixin:
ref #1593

monchoz:
Thanks! that is a good approach that also works. Personally there are some issues I still dislike, in a context, I use detail rows with tab contents in a very CSS styled way on every row and whenever a row gets updated it does not remain in the same same tab, the whole detail content is re-constructed, I can live with that but one thing I don't is the page scrolling to top every time `initBody` gets called, do you know how can I prevent that?

My thoughts about that were on line 1573 but got no luck

`if (!fixedScroll) {
   this.scrollTo(0);
}`

Thanks!

dabros:
@monchoz - a passing thought on scroll issue is that maybe its the browser scrolling to top because there is no content below.

Try using fixed height for table and or its container, height options exist but just as easy to use min-height on your main content containner (assuming nav,container,footer type approach).

`initBody` recreates the contents, so for x milliseconds there is no content there and the whole page gets shorter.

But thats assuming you mean "page" when you say "page", instead of when internal scrollbar in tbody appears.

If you did mean internal scrollbar, then for exact same reason it would scroll up but you could do something like storing current scroll position during `onPreBody` then restoring in `onPostBody`, or equatable approach using other callbacks.

monchoz:
@dabros Thanks! I tried using a fixed height and got the same behaviour. I think caching the current scroll position could be an approach but not sure if still will move from position for a glyph, it's kind of annoying, have to try.

manukieli:
If a table have detailView = true and some row are expanded.
If I update the data of this row, the detailView is collapsed.
How can I keep the detailView open???

UtechtDustin:
@manukieli 
As you maybe see this issue is marked as "feature", its currently not possible.

wenzhixin:
[...]

Issue #1603, opened 2015-10-26
Expanded row is collapsing after row update


Avivbh:
When using updateByUniqueId (and maybe also when using updateRow, didn't check it) the row that is being updated is collapsing.
I don't know if you consider this an unwanted behavior, but in my case it isn't.
To locally fix it I added expandRow_ call in the end of updateByUniqueId method (marked row in picture).
![capture](https://cloud.githubusercontent.com/assets/15223524/10736136/0a4668a2-7c14-11e5-837d-5d81c8c7e965.PNG)

Don't know if you consider this to be a bug/issue, just putting it up there for you to know :)

djhvscf:
Can you provide a fiddle?? @Avivbh

wenzhixin:
#1498

PR #4571, opened 2019-08-31
keep opened details on update/insert

aureliendavid:
Hi, 

[ I don't know if this is merge-able as a full-fledged feature, but it works well for my use case, and since I saw a few people having similar issues (in #1899, #1498, #1603) I'm sharing it in case it is of some interest ]

The problem is that when we call `insertRow()` or `updateByUniqueId()` on a table which have expanded details panes, the whole `tbody` is refreshed and the details panes are closed (and changes made in them are lost). 

The idea of this PR is that when the body is being rebuilt, to check if the row had an expanded details pane, and if it did, to copy it to the "new" row being built. 

This has a few caveats: 

* after an `insertRow()`, the row indices change, so we need to link the old row position and the new one ; to do this I use the `uniqueId` which works well, but it means that this feature will only work on tables that have a `uniqueId`

* when using `updateRowByUniqueId()`, if the row that is being updated had an expanded pane, we don't want to just copy its content, since the update might have changed what it needs to display ; so in this case, we re-expand the details after the update for the updated row (so we'll lose possible changes made to it but I think it makes more sense - this could be a setting)

This probably needs more work and tests but it's been useful in my case. Thanks for this great project btw.


djhvscf:
Please provide an [Online Example](https://live.bootstrap-table.com/) to show this PR working, thanks!

djhvscf:
Hi! There is not any reply, in that case we are going to close this PR. Remember to add an [example](https://github.com/wenzhixin/bootstrap-table/blob/develop/CONTRIBUTING.md#bug-reports). Thanks!

PR #4614, opened 2019-09-29
Fix for #4577

### LINKED ISSUE #4577: Filter Control select control does not work correctly when used in more than one table on the same HTML page
On a single page with two tables, if both tables utilize the filter control, then only one of the table's filter select dropdowns will be populated, and the other table will have empty select dropdowns. 

Example:
https://live.bootstrap-table.com/code/dougauerbach/524

UtechtDustin:
Fixes #4577

Before: https://live.bootstrap-table.com/code/dougauerbach/524
After: https://live.bootstrap-table.com/code/UtechtDustin/775


UtechtDustin:
@wenzhixin 
Our editor has some issues, as example i used the example from the User and tried to change from CDN to this branch. This worked but the changes was not loaded or in the source code (that is the reason why i include the javascript file myself).

Also i tried to remove the extension from the example (unchecked the checkbox), which dont worked for me... after some check/uncheck it worked.. idk

wenzhixin:
@UtechtDustin https://live.bootstrap-table.com/code/wenzhixin/777 
The filter-control url is correct, but I think there is cache when you first open this url: https://cdn.jsdelivr.net/gh/wenzhixin/bootstrap-table@fix/4577/src/extensions/filter-control/bootstrap-table-filter-control.js, you can find that `bootstrap-table-filter-control` is the old code. [rawgit.com] works correct, but it will close.

PR #4633, opened 2019-10-11
Serbian RS translations

kanazir:
Serbian RS cyrillic and latin translations added.


UtechtDustin:
Thanks for your contribution.

PR #4665, opened 2019-10-31
Removed getAllSelections method

wenzhixin:
ref #776

Because in the latest version, the result of `getAllSelections` is the same as `getSelections`.

Now we can use `maintainMetaData` option to get all selected rows when enable paging and/or have filters.

Example: https://live.bootstrap-table.com/code/wenzhixin/993


UtechtDustin:
There is a different.

check the row with of `Item 1` and then search for `randomString`, now use the `getSelections`. You will get no results, but if you do the same thing with `getAllSelections` you will still get the checked data.

https://live.bootstrap-table.com/code/UtechtDustin/997

wenzhixin:
OK, so we still need this method, it that right?

UtechtDustin:
I think so.

PR #4669, opened 2019-10-31
Dont change the view if document was resized by keyboard

### LINKED ISSUE #4627: Android Input lost focus on bootstrap table (cardview)
Hello
in android devices in cardview mode if I try to insert text into the input the keyboard disappears and the control loses focus.
If I don't load the js bootstrap-table-mobile.js and remove the mobileResponsive property I don't have the problem.

jsfiddle: https://jsfiddle.net/Martin90/9p8zxvob/26/
Could be a issue?
Thanks
Andrea

UtechtDustin:
fix #4627

The issue is that if you click on the input, the keyboard will be displayed and resize event will be triggered (lesser height). Because of the resetView the input will be recreated and the keyboard disappears.

### DIFF
--- a/src/extensions/mobile/bootstrap-table-mobile.js
+++ b/src/extensions/mobile/bootstrap-table-mobile.js
@@ -47,9 +47,9 @@ $.BootstrapTable = class extends $.BootstrapTable {
       // reset view if height has only changed by at least the threshold.
       const width = $(window).width()
       const height = $(window).height()
-      const activeElement = $(document.activeElement)
+      const $activeElement = $(document.activeElement)
 
-      if (activeElement && $.inArray(activeElement.prop('nodeName'), ['input', 'select', 'textarea'])) {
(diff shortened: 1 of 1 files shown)


src/extensions/mobile/bootstrap-table-mobile.js
```diff
@@ -47,6 +47,11 @@ $.BootstrapTable = class extends $.BootstrapTable {
       // reset view if height has only changed by at least the threshold.
       const width = $(window).width()
       const height = $(window).height()
+      const activeElement = $(document.activeElement)
+
+      if (activeElement && $.inArray(activeElement.prop('nodeName'), ['input', 'select', 'textarea'])) {
```
wenzhixin:
```
const $activeElement = $(document.activeElement)

if ($activeElement.length && ['INPUT', 'SELECT', 'TEXTAREA'].includes($activeElement.prop('nodeName'))) {
```

UtechtDustin: done

PR #4675, opened 2019-11-01
Added support to use the custom sort attribute

### LINKED ISSUE #2148: Group by - order of group
I'm using the group-by-v2 extension. I've created a fiddle to demonstrate my issue:

http://jsfiddle.net/et9z85zv/3/

I want the order of groups to be decided based on the values of the Id field. In the example, I want 'groupC' to be the first group (which has rows with Id 1, 3 and 4), then 'groupB' and then 'groupA'. But by default, the order is being set by the alphabetical sorting of the group names. 

The default behavior doesn't work for me because the group names have some significance in my application and I need the groups to be in a particular order. Think of it like a calendar, where group names represent month names. Showing the months in alphabetical order April, August, December .. won't make any sense. I need a way to specify the order as January, February, March ..  

Is there a way I can specify the order of groups or have it pick the order based on the values of a particular field ('Id' field in my example)?

Thanks to the bootstrap-table team for creating a wonderful product.

UtechtDustin:
fix #2148 

I think the title says enough :-)

### DIFF
--- a/src/extensions/group-by-v2/bootstrap-table-group-by.js
+++ b/src/extensions/group-by-v2/bootstrap-table-group-by.js
@@ -41,6 +41,7 @@ $.extend($.fn.bootstrapTable.defaults, {
   groupByFormatter: undefined
 })
 
+const Utils = $.fn.bootstrapTable.utils
 const BootstrapTable = $.fn.bootstrapTable.Constructor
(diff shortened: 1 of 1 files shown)


wenzhixin:
Thanks!

PR #4680, opened 2019-11-02
Reorder columns programatically 

### LINKED ISSUE #2578: Reorder columns programatically
Hello. I've set up my table and have been using the column reordering extension so users can move columns by dragging and dropping. What I would like to do is reorder columns programmatically. I'm not seeing any way to do this in the documentation so has anyone had this use case and come up with a solution? I've been trying to use jquery to do it but haven't found a reliable way to consistently and continuously manipulate the column orders. Would I need to rerender the table each time to do this? Thanks.

UtechtDustin:
fix #2578 

Example: https://live.bootstrap-table.com/code/UtechtDustin/1026

### DIFF
--- a/site/docs/extensions/reorder-columns.md
+++ b/site/docs/extensions/reorder-columns.md
@@ -60,11 +60,11 @@ Dependence:
 
 ### orderColumns
 
-- **parameters:** `object` e.g. `{"name": 0, "price": 1}`
+- **parameters:** `object` e.g. `{name: 0, price: 1}`
 
 - **Detail:**
 
-   Reorders the the columns by the given object.   
+   Reorders the the columns by the given object.
    The Object key has to be the [field](https://bootstrap-table.com/docs/api/column-options/#field) and the value is the column index (starts by 0).
 
 ## Events
--- a/site/docs/extensions/reorder-columns.md
+++ b/site/docs/extensions/reorder-columns.md
@@ -56,6 +56,17 @@ Dependence:
 
 - **Default:** `null`
 
+## Methods
+
+### orderColumns
+
+- **parameters:** `object`
+
+- **Detail:**
+
+   Reorders the the columns by the given object.   
+   The Object key has to be the [field](https://bootstrap-table.com/docs/api/column-options/#field) and the value is the column index (starts by 0).
+
 ## Events
 
 ### onReorderColumn(reorder-column.bs.table)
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -283,6 +283,7 @@ class BootstrapTable {
           Utils.sprintf(' rowspan="%s"', column.rowspan),
           Utils.sprintf(' colspan="%s"', column.colspan),
           Utils.sprintf(' data-field="%s"', column.field),
+          Utils.sprintf(' id="%s"', column.field),
```
wenzhixin: Is the `id` attribute required? If there are multiple tables at the same time, the`id` may be the same.
> [The id global attribute defines an identifier (ID) which must be unique in the whole document.](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

UtechtDustin: Yes the id is required for the sorting, but your right.
Should we move the[cookieidtable](https://bootstrap-table.com/docs/extensions/cookie/#cookieidtable) to the core (without cookie prefix), so we can always determinate the table.
In that case i would use the idtable-field

wenzhixin:
Updated: https://live.bootstrap-table.com/code/wenzhixin/1039
https://github.com/wenzhixin/dragtable/commit/15718063ca6b39d40526fe6e7bbf0bcd7e18fdae
Use `data-field` instead of `id`, so we don't need to update the code of `bootstrap-table.js`.

UtechtDustin:
> Updated: https://live.bootstrap-table.com/code/wenzhixin/1039
> [wenzhixin/dragtable@1571806](https://github.com/wenzhixin/dragtable/commit/15718063ca6b39d40526fe6e7bbf0bcd7e18fdae)
> Use `data-field` instead of `id`, so we don't need to update the code of `bootstrap-table.js`.

I dont think this is a good idea, that means that everybody have to use the dragtable from that repository.

- That means nobody can add it with npm, yarn, etc.
- The Link of our documentation points to the original repository.
- Anybody who already used the extension before and try to use this method will fail because they use the original dragtable.

wenzhixin:
But the original repository is no longer updated from Sep 2, 2016.

wenzhixin:
@UtechtDustin Found a solution that doesn't need to change the original repository.

UtechtDustin:
Sure the repository of the dragtable was not updated a long time, but there are a lot of websites which already use the old repository via npm, yarn etc. So we should always prevent modify 3th party libs.

But the new solutions looks good 👍

PR #4687, opened 2019-11-07
Removed resetWidth method

### LINKED ISSUE #4671: the bottom border bug with height option is still there
fix method: padding += this.$header.outerHeight(true) + 1
but in my project ,this problem is still there,when the option contains height,the bottom border donot show

```
var columnArray = [{
					checkbox: true,
					radio:false,
					formatter:function actionFormatter(value, row, index) {
			                }
				},
				{
					field: 'id',
					title: 'id',
					align: 'center', // 横向居中
					valign: 'middle', // 纵向居中
					visible: false,
					sortable: false,
					searchable: false,
					events: {}
				},
				{
					field: 'text',
					title: "文本",
					align: 'center', // 横向居中
					valign: 'middle', // 纵向居中
					visible: true,
					sortable: false,
					events: {}
				}];
$(unit_table).bootstrapTable({
				clickToSelect: true,
				singleSelect: false,
				striped: true,
				columns: columnArray,
				height: 500,
				search: true,
				searchText: "",
				formatSearch: function() {
					return "输入搜索";
				},
				searchAlign: "left",
				onDblClickRow: function (item) {
		                }
			});
```

and the fix problem example link, the before link can show the table,but the after link the table do not show

https://live.bootstrap-table.com/code/wenzhixin/880

wenzhixin:
Because `resetView` method contains the function of `resetWidth`, now the `reset` event of window uses the method of `resetwidth`, which is not comprehensive. So I think we can remove the `resetwidth` method, and use `resetView` instead.

Will fix #4671, you can resize the `result` window to show the problem:

image

Before: https://live.bootstrap-table.com/code/UtechtDustin/1048
After: https://live.bootstrap-table.com/code/wenzhixin/1079

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ChangeLog
 ---------
 
+### 1.15.6
+
+- **Update:** Removed `resetWidth` method and use `resetView` instead.
+
 ### 1.15.5
 
 - **New:** Added `jqXHR` for `responseHandler` option and `onLoadSuccess` event.
--- a/site/docs/api/methods.md
+++ b/site/docs/api/methods.md
@@ -426,14 +426,6 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
 - **Example:** [Reset View](https://examples.bootstrap-table.com/#methods/reset-view.html)
 
-## resetWidth
-
-- **Parameter:** `undefined`
-
-- **Detail:**
-
-  Resizes header and footer to fit current columns width.
-
 ## showLoading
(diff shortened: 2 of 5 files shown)


aammjian:
使用treegrid的时候onresetview的change事件中resetwidth被remove后就会显示错误，需要调整。

wenzhixin:
@aammjian 之前已经更新了的，谢谢提醒： https://github.com/wenzhixin/bootstrap-table-examples/commit/b8f0608dd705dd740b057bfc153ce6f84c036bc1

PR #4692, opened 2019-11-09
Removed natural-sorting extension

### LINKED ISSUE #4686: Extension natural-sorting not working when applied to bootstrap-table starter template
I was testing the natural-sorting extension with the bootstrap-table starter template, but it does not work.

https://live.bootstrap-table.com/code/GCru/1066

(In the editor it does not show the rows but when publishing the file they are shown.)

Thank you

wenzhixin:
Removed the natural-sorting extension

ref #4686


djhvscf:
I agree with @UtechtDustin .. this should be remove as extension and we should add it as example/resource.

Issue #4683, opened 2019-11-06
No data in cell, if data-field are not in the last line of the header


SEModer:
Example https://live.bootstrap-table.com/code/SEModer/1052

If swap line2 and line3, be good.
On show-columns, showed last line header cols....

How to fix this so that only columns with data-field are used? Or is it a bug in this case?

UtechtDustin:
This should work for you: https://live.bootstrap-table.com/code/UtechtDustin/1053

SEModer:
> This should work for you: https://live.bootstrap-table.com/code/UtechtDustin/1053

Yes, i'm know, but:
![image](https://user-images.githubusercontent.com/789171/68296494-9447c200-00b6-11ea-909e-7e12cb0b3fb9.png)

UtechtDustin:
Oh okay i understand...
Never worked with colspans, also i dont know how the code works for that..
So @wenzhixin could you help him ?

djhvscf:
For me this is a bug. @UtechtDustin @wenzhixin

akkivarma10:
Hi, I am looking into fixing this issue, any suggestions @djhvscf @wenzhixin @UtechtDustin

wenzhixin:
@akkivarma10 As currently designed, your case is not supported.

akkivarma10:
Hi @wenzhixin , I have fixed this issue on local, will be submitting a Pull request in some time.

wenzhixin:
I found a workaround for your case: https://live.bootstrap-table.com/code/wenzhixin/1364.

PR is welcome.

akkivarma10:
Hi @wenzhixin , I have submitted the Pull request. Please review it & let me know if any changes needs to be done. Thanks

wenzhixin:
ref #3213

wenzhixin:
I think it's hard to implement this feature. 
You can use the workaround.

Issue #3213, opened 2017-06-16
Table structure breaks with column visibility control enabled and multirow headers


johnnymetz:
Table structure breaks using the `data-show-columns="true"` attribute on a table equipped with a multirow header. Try hiding a column in the following example: https://live.bootstrap-table.com/code/wenzhixin/1409

It appears the cell in any header row above the lowest row aren't hidden.

SO question + more examples here: https://stackoverflow.com/questions/44104377/bootstrap-table-table-structure-breaks-when-hiding-colspan-columns

djhvscf:
Hi @jonathanmetz it is an issue. This is included to be fied in our next release 1.12.0

dskilken:
This is the same problem as: #3560.    If you use the methods to hide, it cannot handle multiple rows in the thead.  This has to do with the way the table is initialized and built.    I had to work around this by using divs in the header cells.  This is not a small issue because it affects other plug-ins like Sticky-Header and Fixed-Column.     As a work around to your example, you would have to create your own javascript to hide columns using jQuery.  If you use unique classes on each column, you can easily hide a class in jquery.  If you are using extensions like sticky header, you can't really do this, and are stuck working around the format.

dskilken:
If this is fixed in a future release, please make sure to test it with other extensions - thanks

mkyral:
Hi,
I run into this issue as well. I have a quite complex header with 4 rows (many colspan and rowspan). The user request is ability to collapse multiple columns within the colspan group to one column. It worked correctly when I hide several columns. But from some number of columns, the header breaks. Colspan was not counted correctly.

I've investigated this issue and find out, that function setFieldIndex() does not handle correctly the rowspan/colspan combination.

This code:
```js
for (let k = 0; k < rowspan; k++) {
  flag[i + k][index] = true
}
for (let k = 0; k < colspan; k++) {
  flag[i][index + k] = true
}
```
sets first rowspan and then colspan. It works if there is only rowspan or only colspan. But in case when both are set, it sets true only for first column of all other rows.

e.g. rowspan=3, colspan=3 it will generate:
```
true|true|true|
true|false|false|
true|false|false|
```
But should generate:

```
true|true|true|
true|true|true|
true|true|true|
```

So fix should be:
```js
for (let k = 0; k < rowspan; k++) {
  for (let j = 0; j < colspan; j++) {
    flag[i + k][index+j] = true
  }
}
```
… for each row in rowspan will set colspan number of columns to true.

[...]

PR #4743, opened 2019-12-10
Fix bug - No data in cell, if data-field are not in the last line of the header

akkivarma10:
In case of multiple header rows, the row with data-field was not being considered
unless it was the last row.


akkivarma10:
Fixes the issue #4683

wenzhixin:
There are still bugs when you switch fields.

image

https://live.bootstrap-table.com/code/wenzhixin/1368

akkivarma10:
Ohh, I see. I also checked the example from the issue link & found that this issue is there also. Will look into it.

wenzhixin:
ref #3213, but I think it's hard to implement.

djhvscf:
Any update here?

djhvscf:
Close this PR because there is no any response in 7 days, if the problem exists reopen it, thanks!

Issue #4746, opened 2019-12-11
Column checkbox not being disabled in showColumns dropdown in case of toggleAll


akkivarma10:
Hi, I think I have encountered a bug where the column checkboxes are not being disabled when the checked columns count is equal to the minimumCountColumns. Also it only occurs when the showColumnsToggleAll is set to true.

Steps to reproduce the bug:
1. Create a table with 3 columns
2. Set showColumns & showColumnsToggleAll to true
3. Set minimumCountColumns as 2
4. Uncheck any 1 column (now the remaining 2 columns should have been disabled, but it doesn't)

[Online Example](https://live.bootstrap-table.com/code/akkivarma10/1397)

I think I have identified the issue. The issue is in the toggleColumn function at line no 2345
`const $items = this.$toolbar.find('.keep-open input').prop('disabled', false)`
Changing the selector in find to exclude toggle-all input element as follows seems to fix the issue. 
`const $items = this.$toolbar.find('.keep-open input:not(".toggle-all")').prop('disabled', false)`
I would love to submit a PR for the same.

UtechtDustin:
fixed in #4747

PR #4747, opened 2019-12-11
Fix bug - 4746 - Column checkbox not being disabled in case of toggleAll

akkivarma10:
Fixes issue #4746
Where column checkbox is not being disabled when the checked column
count is equal to minimumCountColumns


UtechtDustin:
Thank you for your contribution !

akkivarma10:
Thanks @UtechtDustin for quick review.

PR #4748, opened 2019-12-11
Parameter not being passed to resetSearch function

TantumLabs:
When using $table.bootstrapTable('resetSearch', 'searchText') the parameter was not being passed to the function


UtechtDustin:
Thank you for your contribution!

Issue #4776, opened 2020-01-01
Toggle All bug in case of checkbox / radio column


akkivarma10:
Toggle All bug when checkbox / radio column is present

There is a bug where the "Toggle all" hides all the column when the checkbox / radio column is present. In toggle dropdown (for default minimumCountColumns), after toggling all i.e hiding all columns, one column shows as checked & disabled, but in actual all the columns are hidden (except for checkbox column). The bug is, currently while toggling, it also takes the count of checkbox/radio columns bcoz of which it hides all other columns.

Steps to reproduce the issue in the given example link:
1. Click on the columns dropdown
2. Uncheck "Toggle all" option
You can see that 1 column shows as checked & disabled, but is hidden in table

[Online Example](https://live.bootstrap-table.com/code/akkivarma10/1506)

Findings: I think we need to filter out checkbox & radio columns in the function getVisibleColumns at line 2366. Also there might be more than 1 place where we need to add this filter.

PR #4777, opened 2020-01-01
Fix bug - 4776 - Toggle All bug when checkbox / radio column is present

akkivarma10:
Fixes issue #4776
Added filter for selection control columns (radio & checkbox) to fix
toggle all bug when checkbox or radio column is present.

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2371,7 +2371,7 @@ class BootstrapTable {
   }
 
   isSelectionColumn (column) {
-    return (column.radio || column.checkbox)
+    return column.radio || column.checkbox
   }
 
   showAllColumns () {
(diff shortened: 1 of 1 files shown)


UtechtDustin:
@akkivarma10 
Thank you for the pullrequest, could you add two examples ? (before and after)

akkivarma10:
Hi @UtechtDustin , I tried creating an example on https://live.bootstrap-table.com/, but unfortunately it isn't working. Found that it is not able to find the variables.scss file which is now renamed to _variables.scss in the new commit. Can u fix this. I will create a working example once this is done. Thanks.

UtechtDustin:
@akkivarma10 
Good catch!
I dont have access to the editor. @wenzhixin please fix it :-)

wenzhixin:
Fixed the online editor bug: https://live.bootstrap-table.com/code/wenzhixin/1519
Thanks for your contribution!

PR #4807, opened 2020-01-26
Fixing strings comparation

### LINKED ISSUE #4788: Multiple Column sort doesn't work with booleans
If you include a boolean column in multiple sort condition, the column is basically ignored.

Live example: https://live.bootstrap-table.com/code/AlexSikilinda/1539

This happens because of this code:

     if (typeof aa !== 'string') {
         aa = aa.toString();
     }

from onMultipleSort:

    BootstrapTable.prototype.onMultipleSort = function() {
        var that = this;

        var cmp = function(x, y) {
            return x > y ? 1 : x < y ? -1 : 0;
        };

        var arrayCmp = function(a, b) {
            var arr1 = [],
                arr2 = [];

            for (var i = 0; i < that.options.sortPriority.length; i++) {
                var order = that.options.sortPriority[i].sortOrder === 'desc' ? -1 : 1,
                    aa = a[that.options.sortPriority[i].sortName],
                    bb = b[that.options.sortPriority[i].sortName];

                if (aa === undefined || aa === null) {
                    aa = '';
                }
                if (bb === undefined || bb === null) {
                    bb = '';
                }
                if ($.isNumeric(aa) && $.isNumeric(bb)) {
                    aa = parseFloat(aa);
                    bb = parseFloat(bb);
                }
                if (typeof aa !== 'string') {
                    aa = aa.toString();
                }

                arr1.push(
                    order * cmp(aa, bb));
                arr2.push(
                    order * cmp(bb, aa));
            }

            return cmp(arr1, arr2);
        };

        this.data.sort(function(a, b) {
            return arrayCmp(a, b);
        });

        this.initBody();
        this.assignSortableArrows();
        this.trigger('multiple-sort');
    };

So it converts one bool to a string, and then compare the string with other boolean which causes compare function to always return 0.

djhvscf:
Fix #4788

### DIFF
--- a/site/docs/extensions/multiple-sort.md
+++ b/site/docs/extensions/multiple-sort.md
@@ -38,6 +38,16 @@ toc: true
 
 - **Default:** `true`
 
+### multiSortStrictSearch
+
+- **type:** `Boolean`
+
+- **Detail:**
+
+   Set true to enable strict search. This means that strings will be compared using toLowerCase.
+
+- **Default:** `false`
+
 ### sortPriority
 
 - **type:** `Object`
--- a/site/docs/extensions/multiple-sort.md
+++ b/site/docs/extensions/multiple-sort.md
@@ -38,13 +38,13 @@ toc: true
 
 - **Default:** `true`
 
-### multiSortStrictSearch
+### multiSortStrictSort
 
 - **type:** `Boolean`
 
 - **Detail:**
 
-   Set true to enable strict search. This means that strings will be compared using toLowerCase.
+   Set true to enable strict sorting. This means that strings will be compared and ordered using toLowerCase.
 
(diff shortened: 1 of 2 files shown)


site/docs/extensions/multiple-sort.md
```diff
@@ -38,6 +38,16 @@ toc: true
 
 - **Default:** `true`
 
+### multiSortStrictSearch
```
wenzhixin: Is `multiSortStrict` or `multiSortStrictSort` better?

UtechtDustin: multiSortStrictSort i guess

djhvscf: Yes, you both are right

PR #4835, opened 2020-02-02
feature/4585

### LINKED ISSUE #4585: Bootstrap-table multiple-sort
Hello

bootstrap-table-multiple-sort extension.

Can you add method something like: 
doMultipleSort([{sortName: “name”,sortOrder:”desc”},{sortName:”last_name”,sortOrder:”desc”}]); //so people can make own popups and so on...

That extension cannot be used without bootstrap.js and bootstrap.css added (why? too bad solution). Can you make it more flexible, it should work without default bootstrap too...

UtechtDustin:
fix #4585 

Example: https://live.bootstrap-table.com/code/UtechtDustin/1654

### DIFF
--- a/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
+++ b/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
@@ -702,7 +702,6 @@ BootstrapTable.prototype.setButtonStates = function () {
 }
 
 BootstrapTable.prototype.multiSort = function (sortPriority) {
-  console.log(sortPriority)
   this.options.sortPriority = sortPriority
   this.options.sortName = ''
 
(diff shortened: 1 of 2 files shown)


src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
```diff
@@ -712,3 +700,21 @@ BootstrapTable.prototype.setButtonStates = function () {
     this.$sortModal.find('#delete').attr('disabled', 'disabled')
   }
 }
+
+BootstrapTable.prototype.multiSort = function (sortPriority) {
+  this.options.sortPriority = sortPriority
+  this.options.sortName = ''
+
+  if (this.options.sidePagination === 'server') {
+    const t = this.options.queryParams
```
djhvscf: Can this variable be named as queryParams?

UtechtDustin: This part of the code was just moved to a function (not written by me), also this extension shoud be rewritten sometime.
But i removed the variable and use the queryParams without a variable.

PR #4845, opened 2020-02-05
added a new option to get/set/delete the values by a custom function

### LINKED ISSUE #4840: Saving table save state in database (Add custom save/load handler)
i am trying to save the table save state in Database and use it again after next login.
any example

UtechtDustin:
fix #4840


einfallstoll:
This isn't async right? Because it wouldn't be very useful, if you couldn't store/read it asynchronously (e.g. via API).

PR #4875, opened 2020-02-16
Updated the docs API sort by alphabetical

### LINKED ISSUE #4526: Suggestions to improve the documentation
Some suggestions to improve the documentation of bootstrapTable:

- column option "sorter": please include expected return values for custom sorter function (it was only through guessing that I found out it's actually -1, 0, 1)
- table options/column options/events/methods: please sort the options/methods in alphabetical order - also in the index in the right column (it's very tedious to scan for a certain option in random order)
- add to each option/method in which version it was added. Also if additional parameters were added in later versions please note accordingly (it has happened more than once that I tried for ages to use a certain option only to finally look into the bootstrapTable source code to find out that it is not yet included in the version we use)

Much appreciated!

wenzhixin:
Fix #4526

### DIFF
--- a/site/docs/api/column-options.md
+++ b/site/docs/api/column-options.md
@@ -8,118 +8,6 @@ toc: true
 
 The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 
-## field
-
-- **Attribute:** `data-field`
-
-- **Type:** `String`
-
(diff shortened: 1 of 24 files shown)


wenzhixin:
> Should be done for the extensions aswell.

Updated

PR #4891, opened 2020-02-20
fix/4873

### LINKED ISSUE #4873: Text/textarea field in editable extension doesn't handle quotation marks
I opened this Editable extension issue here (not x-editable repo) because their live demo works whereas bootstrap-table's demo doesn't. I am not familiar with Web development, maybe I missed something, please correct me.

**Context**

I was trying to use a table cell to store a paragraph (200 - 300 words) and the text contains single and double quotation marks.

**Environment**

- Chromium: 79.0.3945.130
- bootstrap 3
- bootstrap-table: latest
- bootstrap-table-editable: latest
- x-editable: 1.5.1

**Reproduce the Issue**

Test input: `This is a "quotation" mark`

1. Edit any text cell to test input in [demo from Bootstrap-Table](https://examples.bootstrap-table.com/index.html?bootstrap3#extensions/editable.html), it truncates the test input to `This is a`.
2. The textarea and test cell in [demo from x-editable](https://vitalets.github.io/x-editable/demo-bs3.html), preserves the whole test input.

**Potential Cause**

My guess is bootstrap-table (or some other library) tries to use `data-value` to store the text input, and html.dataset attribute doesn't handle quotation marks well, single or double quotation. And some code truncates the input to ensure that doesn't happen.

**My Hack**

The truncation seems to happen before x-editable submit() is called. So I escape the input before it happens.

From [x-editabel docs](https://vitalets.github.io/x-editable/docs.html#editable)
> Since 1.5.1 you can modify submitted value by returning object from validate: {newValue: '...'} or {newValue: '...', msg: '...'}

So, I added the followings to the x-editable's validate method.

1. Escape the `var escaped_value = value.replace(/"/g, "&quot;").replace(/'/g, "&apos;");`
2. Then `return {"newValue": escaped_value};`

UtechtDustin:
Fix #4873

Just edit the name column and some text with a `"` e.g. `Test"Test`.

Before: https://live.bootstrap-table.com/code/UtechtDustin/1893
After: https://live.bootstrap-table.com/code/UtechtDustin/1892

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -102,7 +102,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
         return
       }
 
-      const data = this.getData()
+      const data = this.getData(undefined, true)
       const $field = this.$body.find(`a[data-name="${column.field}"]`)
 
       $field.each((i, element) => {
@@ -131,6 +131,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
           }
         }
 
+        submitValue = Utils.escapeHTML(submitValue)
         $this.data('value', submitValue)
         row[column.field] = submitValue
         this.trigger('editable-save', column.field, row, rowIndex, oldValue, $this)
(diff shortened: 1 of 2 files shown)


src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -157,4 +158,23 @@ $.BootstrapTable = class extends $.BootstrapTable {
     })
     this.trigger('editable-init')
   }
+
+  getData (params, escaping) {
```
wenzhixin: I think it's better to use `params.escape` instead.

src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -157,4 +158,23 @@ $.BootstrapTable = class extends $.BootstrapTable {
...
+    if (escape) {
+      $.each(data, (i, object) => {
+        $.each(object, (objectIndex, value) => {
+          object[objectIndex] = Utils.unescapeHTML(value)
+        })
+      })
+    }
+
+    return data
+  }
```
wenzhixin:
```
getData (params, escaping) {
  const data = super.getData(params)

  if (params.escape) {
    for (const row of data) {
          row[key] = Utils.unescapeHTML(value)
        }
    }
  }
}
```

PR #4909, opened 2020-02-25
Added  field for remove method

### LINKED ISSUE #4907: remove: needs ability to be remove by rowIndex
```
	    key: "remove",
	    value: function remove(params) {
	      var len = this.options.data.length;
	      var i;
	      var row;

	      if (!params.hasOwnProperty('field') || !params.hasOwnProperty('values')) {
	        return;
	      }

		  **if (params.field.includes('rowIndex'))
			  params.values.forEach(rowIndex => this.options.data.splice(rowIndex, 1));**
		  else
	      for (i = len - 1; i >= 0; i--) {
	        row = this.options.data[i];

	        if (!row.hasOwnProperty(params.field)) {
	          continue;
	        }

	        if (params.values.includes(row[params.field])) {
	          this.options.data.splice(i, 1);

	          if (this.options.sidePagination === 'server') {
	            this.options.totalRows -= 1;
	          }
	        }
	      }

	      if (len === this.options.data.length) {
	        return;
	      }

	      this.initSearch();
	      this.initPagination();
	      this.initSort();
	      this.initBody(true);
	    }
```

wenzhixin:
Fix #4907, #564

Example: https://live.bootstrap-table.com/code/wenzhixin/1992

### DIFF
--- a/site/docs/api/methods.md
+++ b/site/docs/api/methods.md
@@ -381,8 +381,8 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
   Remove data from table, the params contain two properties:
 
-  * `field`: the field name of remove rows.
-  * `values`: the array of values for rows which should be removed.
+  * `field`: the field name of remove rows. If `$index` is not in your fields, you can use this special field `$index` to remove rows by row index.
+  * `values`: the array of values for rows which should be removed. If you use the special field `$index`, you can pass an array of indexes.
 
 - **Example:** [Remove](https://examples.bootstrap-table.com/#methods/remove.html)
 
@@ -648,4 +648,3 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
   * `replace` (optional): set to `true` to replace the row instead of extending.
 
 - **Example:** [Update Row](https://examples.bootstrap-table.com/#methods/update-row.html)
-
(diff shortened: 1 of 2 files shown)


site/docs/api/methods.md
```diff
@@ -381,7 +381,7 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
   Remove data from table, the params contain two properties:
 
-  * `field`: the field name of remove rows.
+  * `field`: the field name of remove rows. If `$index` dose not in your fields, you can use this special field `$index` to remove rows by row index.
```
djhvscf: Typo dose

djhvscf: Also, this should include that we can pass an array of indexes

site/docs/api/methods.md
```diff
@@ -381,8 +381,8 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
   Remove data from table, the params contain two properties:
 
-  * `field`: the field name of remove rows.
-  * `values`: the array of values for rows which should be removed.
+  * `field`: the field name of remove rows. If `$index` does not in your fields, you can use this special field `$index` to remove rows by row index.
```
djhvscf: Hi @wenzhixin I think the correct statement should be something like 

If `$index` **is not** in your fields, you can use this special field `$index` to remove rows by row index.

wenzhixin:
Updated

PR #4928, opened 2020-03-12
First time issue with cookie and filter-control

wollistik:
This fixes an issue where Cookie data is not present (e.g. first time visit to a page) and filter-control enabled (with a select input).

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -195,11 +195,10 @@ const UtilsCookie = {
   },
   initCookieFilters (bootstrapTable) {
     setTimeout(() => {
+      const cachedFilters = {}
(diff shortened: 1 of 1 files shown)


wenzhixin:
Please provide an [Online Example](https://live.bootstrap-table.com) to show your changes, thanks!

UtechtDustin:
@wollistik do you have any news for us ?

wollistik:
Hi, sorry for the long time with no response. I really tried to get a working example, but could not get some working to show you the issue.
I encountered a strange behavior, which might be the root cause.
When you have a look at the following example, you will encounter following behavior: 

* open the page
* table gets initialized
* cookies for filtering/pagination will be created short time (< 1 second) after the table is displayed, even if nothing is filtered for

https://live.bootstrap-table.com/code/wollistik/2343

The last point is where the issue - in my case the cookies do not get created in my site. I have looked through the code, but I couldn't figure out why and when this happens in the demo. It seems like this callback does not get executed when using with webpack - and therefore filtering hangs after the first time the cookie is created with filter values.

UtechtDustin:
So that means this PR will not fix a acutally issue with the bootstrap-table.
Lets talk about your issue in the issue (#4929).

PR #4940, opened 2020-03-21
Fix ui bug

### LINKED ISSUE #4915: 在使用1.16.0的版本中，使用bootstrap-table-treegrid报RangeError:MaxiMum call stack size exceeded错误
在使用1.16.0的版本中，使用bootstrap-table-treegrid报**RangeError:MaxiMum call stack size exceeded**错误，请问什么原因？

以下是bootstrap-table-treegrid.js中57行代码（就是这段代码在报错）：
      // jquery.treegrid.js
      this.options.rowStyle = (item, idx) => {
        const res = this._rowStyle(item, idx)
        const id = item[this.options.idField] ? item[this.options.idField] : 0
        const pid = item[this.options.parentIdField] ? item[this.options.parentIdField] : 0
        res.classes = [
          res.classes || '',
          `treegrid-${id}`,
          `treegrid-parent-${pid}`
        ].join(' ')
        return res
      }
      this.initTr(node, $.inArray(node, data), data, parentDom)
    }

### LINKED ISSUE #4919: Tables do not respect server sort anymore
Hi,

using 1.16.0, server-sort will be overridden by client-sort even if sidePagination is set to "server" and serverSort is set to true;

"initSort" is called even if both settings above are set and therefore the client will resort the responded server data.

### LINKED ISSUE #4920: Extension: Exporting with selected mode cannot export data rows exceed the number of page limit (Client side)
The correct number of rows cannot be successfully exported **when the total number of rows exceeds the page limit**. 

I have turned on the ``maintain-meta-data`` and ``client-side pagination``.

I have created an example [here](https://live.bootstrap-table.com/code/DingkunLiu/2077).  

For example, I can **check all items on the first page**, and **switch to the next page and check another three more**. Then when I export the table, **I can only get the three rows on the second page**, which is the current page I'm on.

This issue seems to be related to the issue #3377.

### LINKED ISSUE #4926: 当使用参数detailView: true,detailViewIcon: false时，表格列的按钮事件绑定失败
版本：v1.16.0
问题描述：
当使用一下表格参数时，表格列的events参数无法绑定事件
```
{
detailView: true,
detailViewIcon: false
}
```

简单排查了一下发现如下代码：
```javascript

key: "initBodyEvent",
	    value: function initBodyEvent() {
//..
//... 源码大概位于5540
	        if (typeof events === 'string') {
	          events = Utils.calculateObjectValue(null, events);
	        }

	        var field = _this9.header.fields[i];

	        var fieldIndex = _this9.getVisibleFields().indexOf(field);

	        if (fieldIndex === -1) {
	          return;
	        }

//问题出在这
			  if (_this9.options.detailView && !_this9.options.cardView) {
	          fieldIndex += 1;
	        }

	        var _loop = function _loop(key) {
	          if (!events.hasOwnProperty(key)) {
	            return "continue";
	          }

	          var event = events[key];

	          _this9.$body.find('>tr:not(.no-records-found)').each(function (i, tr) {
//...
```
如果detailViewIcon为false，则表格列已经少了detailViewIcon的列，但是并没有过滤此情况。导致绑定事件的列往后偏移一列

### LINKED ISSUE #4927: doesn't fire "on-all" event in vue component
Can't catch "on-all" event from vue component, but any other works fine: 
on-options-refresh, on-reset-view, on-dbl-click-row etc.

template:
`<bootstrap-table` ref="table" :columns="tblData.columns" :data="tblData.rows" :options="tableData.options" @on-all="UpdateColumnWidth" @on-dbl-click-row="onDblClickModalRow" @on-check="onCheckModalRow"></bootstrap-table>\'`

### LINKED ISSUE #4939: In Vue component - Cell Formatter function doesn't take "field" parameter. Only value, row and index.
Only in Vue, but in online editor - all works fine
So, it should be three or four?
image

wenzhixin:
Fix #4939

---

Fix #4926 
* Before: https://live.bootstrap-table.com/code/wenzhixin/2210
* After: https://live.bootstrap-table.com/code/wenzhixin/2211

---

Fix #4927

---

Fix #4915

* Before: https://live.bootstrap-table.com/code/wenzhixin/2214
* After: https://live.bootstrap-table.com/code/wenzhixin/2214

---

Fix #4919

* Before: https://live.bootstrap-table.com/code/drummer73/2221
* After: https://live.bootstrap-table.com/code/wenzhixin/2222

---

Fix #4920

* Before: https://live.bootstrap-table.com/code/DingkunLiu/2077
* After: https://live.bootstrap-table.com/code/wenzhixin/2223

### DIFF
--- a/site/docs/api/column-options.md
+++ b/site/docs/api/column-options.md
@@ -245,7 +245,7 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 
   The context (this) is the column Object.
 
-  The cell formatter function, take three parameters:
+  The cell formatter function, take four parameters:
 
   * `value`: the field value.
   * `row`: the row record data.
@@ -504,4 +504,3 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 - **Default:** `px`
 
 - **Example:** [Width Unit](https://examples.bootstrap-table.com/#column-options/width-unit.html)
-
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -225,7 +225,7 @@ class BootstrapTable {
     this.options.columns.forEach((columns, i) => {
       html.push('<tr>')
 
-      if (i === 0 && !this.options.cardView && this.options.detailView && this.options.detailViewIcon) {
+      if (i === 0 && Utils.hasDetailViewIcon(this.options)) {
         html.push(`<th class="detail" rowspan="${this.options.columns.length}">
           <div class="fht-cell"></div>
           </th>
@@ -1285,7 +1285,7 @@ class BootstrapTable {
       html.push(`<td colspan="${this.header.fields.length}"><div class="card-views">`)
     }
 
-    if (!this.options.cardView && this.options.detailView && this.options.detailViewIcon) {
+    if (Utils.hasDetailViewIcon(this.options)) {
       html.push('<td>')
 
       if (Utils.calculateObjectValue(null, this.options.detailFilter, [i, item])) {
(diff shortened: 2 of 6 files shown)


UtechtDustin:
Seems working, except the fix for #4915.
I dont know whta the fix ist, becasue its in chinese, but if i click on the refresh Button the is empty.

wenzhixin:
Sorry, the after link is error.

Fix #4915

Before: https://live.bootstrap-table.com/code/wenzhixin/2214
After: https://live.bootstrap-table.com/code/wenzhixin/2215

PR #4951, opened 2020-03-27
filterDatepickerOptions Json parse

zeydo:
filterDatepickerOptions comes as a string, and the datepicker plugin ignores it. I solved this problem by filtering the values of filterDatepickerOptions by json parse. Sorry for bad english.

### DIFF
--- a/dist/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/dist/extensions/filter-control/bootstrap-table-filter-control.js
@@ -2336,7 +2336,7 @@
 	        $.each(that.columns, function (i, _ref9) {
 	          var filterControl = _ref9.filterControl,
 	              field = _ref9.field,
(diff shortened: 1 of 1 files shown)


UtechtDustin:
@zeydo 
Thank you for the contribution.
Could you please provide us two example for that PR ?
One before the request and one after the fix ?

Please use our [editor](https://live.bootstrap-table.com/) for the examples.

zeydo:
@UtechtDustin 
Hi,
[https://live.bootstrap-table.com/code/zeydo/2277
](https://live.bootstrap-table.com/code/zeydo/2277
)As you can see, I made a description to select multiple dates for datepicker options and ignored it. It does not perceive other options. Because datepicker perceives options as string values, but these options must be objects.

UtechtDustin:
You edited the dist file instead of the source file, please fix that.
After that we can check it.

UtechtDustin:
#4953

PR #5018, opened 2020-05-06
fix onPostBody this context

Zhangjin1993:
When I use reorder-rows plugin, parameter `data` and `this` in onPostBody are lost. Then, I change the source code and everything seems right. So, I created this pr and I'm still not sure if I considered everything.
Before: https://live.bootstrap-table.com/code/Zhangjin1993/2786
After: https://live.bootstrap-table.com/code/Zhangjin1993/2787


wenzhixin:
Thanks for your contribution!

PR #5041, opened 2020-05-24
Feature/5005

### LINKED ISSUE #5005: Send searchable fields in ajax data
I'm writing dynamic php script to receive "bootstrap-table" parameters and generate json result.

The problem is that i receive "search" with searchString but not searchable fields to add in where query.

¿It's is possible to send in request? all elements with this attribut "data-searchable="true""???  or manualy adding in new data attribute???

Regards!

UtechtDustin:
Fix #5005 

Before: https://live.bootstrap-table.com/code/nicojmb/2694
After: https://live.bootstrap-table.com/code/UtechtDustin/3039

(Check the query params while searching)

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -952,7 +952,7 @@ class BootstrapTable {
     pageList = pageList.map(value => {
       if (typeof value === 'string') {
         return value.toLowerCase() === opts.formatAllRows().toLowerCase() ||
-        ['all', 'unlimited'].includes(value.toLowerCase())
+          ['all', 'unlimited'].includes(value.toLowerCase())
           ? opts.formatAllRows() : +value
       }
       return value
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -527,13 +527,13 @@ class BootstrapTable {
 
       toggle: `<button class="${this.constants.buttonsClass}" type="button" name="toggle"
         aria-label="Toggle" title="${opts.formatToggle()}">
-        ${opts.showButtonIcons ? Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, opts.icons.toggleOff) : '' }
+        ${opts.showButtonIcons ? Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, opts.icons.toggleOff) : ''}
         ${opts.showButtonText ? opts.formatToggleOn() : ''}
         </button>`,
 
       fullscreen: `<button class="${this.constants.buttonsClass}" type="button" name="fullscreen"
         aria-label="Fullscreen" title="${opts.formatFullscreen()}">
-        ${opts.showButtonIcons ? Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, opts.icons.fullscreen) : '' }
+        ${opts.showButtonIcons ? Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, opts.icons.fullscreen) : ''}
         ${opts.showButtonText ? opts.formatFullscreen() : ''}
         </button>`,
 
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1653,6 +1653,15 @@ class BootstrapTable {
...
     }
 
+    if (this.options.search && this.options.sidePagination === 'server') {
+      params.searchable = []
+      for (const column of this.columns) {
+        if (!column.checkbox && column.searchable && ((this.options.visibleSearch && column.visible) || !this.options.visibleSearch)) {
+          params.searchable.push(column.field)
+        }
+      }
+    }
```
wenzhixin, line 1663: https://live.bootstrap-table.com/code/wenzhixin/3069

I think that only sends the parameters to the server when sets a searchable `false` for certain columns.

```js
    if (this.options.search && this.options.sidePagination === 'server') {
      const searchable = []

      for (const column of this.columns) {
        if (!column.checkbox && column.searchable && ((this.options.visibleSearch && column.visible) || !this.options.visibleSearch)) {
         searchable.push(column.field)
        }
      }

      if (this.columns.filter(column => !column.searchable).length) {
        params.searchable = searchable
      }
    }
```

src/bootstrap-table.js
```diff
@@ -952,7 +952,7 @@ class BootstrapTable {
     pageList = pageList.map(value => {
       if (typeof value === 'string') {
         return value.toLowerCase() === opts.formatAllRows().toLowerCase() ||
-          ['all', 'unlimited'].includes(value.toLowerCase())
+        ['all', 'unlimited'].includes(value.toLowerCase())
```
wenzhixin, line 955: This line does not need to be modified

src/bootstrap-table.js
UtechtDustin: I checked my and your example again, it sends the field to the server if searchable is set to `true`.

src/bootstrap-table.js
UtechtDustin: auto formatting of phpstorm :-(
I reverted this :)

src/bootstrap-table.js
wenzhixin: Sorry for the late reply. I mean, if we don’t set searchable for all our columns, then we don’t send params to the server, which keeps the same behavior as before. What do you think?

wenzhixin: image

UtechtDustin: But isnt it correct that we dont send the paramteres to the server if we dont have searchable columns ?
So the server can check if there are searchable columns, if yes filter the data based on the columns and the search string. If not ignore the search and return all data.

wenzhixin: The server can check if there are searchable columns, if yes filter the data based on the columns and the search string. If not filter the data based on the search string, which keeps the same behavior as before.

UtechtDustin: No its not the same behavior as before.

Before the server only got the search string, which means the server didnt know for which columns he should search.

After the server get a list of columns which he should search for the given search string.
If the Server get no list of columns, the server can return the whole dataset.

@djhvscf could please also check tha ?

wenzhixin: image

In most cases, the searchable option of the column is not set, which means that all columns are searched. Without this parameter, it means **All**.

For example https://developer.github.com/v3/issues/#list-repository-issues.

> labels: A list of comma separated label names.

If there are 100 labels in our project, they are label1, label2, ... label100. When we want to search for issues with all labels, we don't need to write: labels=label1,label2...label100, we don't need to pass this parameter, it means all.

wenzhixin:
@UtechtDustin update this PR?

UtechtDustin:
@wenzhixin i will do, but im currently busy.
If you want you can add the last check (If all columns are searchable, dont send a additional parameter).

wenzhixin:
Updated

PR #5047, opened 2020-05-27
Removed return key validation for select

### LINKED ISSUE #5046: data-search-on-enter-key along with select not working
Created a sample.

https://live.bootstrap-table.com/code/sekarpdkt/3093

if you make   `data-filter-control="false"`, it works

Thanks
Regards
Sekar

sekarpdkt:
This shall fix the issue of select filter not working, if we enable filter on enter key option

Fixes #5046


UtechtDustin:
Fixes #5046

Issue #5061, opened 2020-06-05
Add toggleFilterControl method


marceloverdijk:
Similar as e.g. `togglePagination` and `toggleFullscreen` it would be nice to have a `toggleFilterControl` method.

Issue #5065, opened 2020-06-06
Add showFilterControlSwitch to show/hide filter control from toolbar


marceloverdijk:
It would be nice to have an option like `data-show-filter-control-switch`: `true` to show/hide the filter controls dynamically and have a button for it in the toolbar.

Similar as the show pagination switch.

marceloverdijk:
Maybe it could use the fa toggle-on/toggle-off icons?

marceloverdijk:
Related to #5061

marceloverdijk:
This is a working example: https://live.bootstrap-table.com/code/marceloverdijk/3277

Wdyt?

UtechtDustin:
Looks good, but i think we should first fix and merge #5061.
After that we can use the new method for the new toggle button.

marceloverdijk:
@UtechtDustin Yes I agree to fix #5061 first. I have looked at it but can't find it yet.
This example was just to get feedback, it should off course be integrated in the filter control extension.

Btw maybe the approach in #5061 by calling `refreshOptions` is maybe not the good way to do it?

```
toggleFilterControl () {
    this.refreshOptions({
      filterControl: !this.getOptions().filterControl
    })
  }
```

UtechtDustin:
I also think this not the best way to solve it, because its a method which is designed for external usage not for internal.

Im currently busy to check that, but as soon i have some freetime i will check that!

marceloverdijk:
@djhvscf Once #5061 is solved and merged I will create a PR asap for the adding a switch button in the toolbar.

Issue #4869, opened 2020-02-12
Column filters not showing when using table fixed height


EtienneTLM:
Hi, it seems there is a bug with the Filter Control extension when using the 'data-height' attribute on the table element. Tested on Chrome and Firefox.

You can see it in action by going on the online editor page for the Filter Control : https://live.bootstrap-table.com/example/extensions/filter-control.html and then adding a 'data-height' attribute to the table element (example: data-height="400")

It gives something like this :

![image](https://user-images.githubusercontent.com/18191493/74379077-1a119280-4db5-11ea-8299-2f2a7b26a9b6.png)

Thanks for your help!

fabiawrc:
Same problem

dougauerbach:
I concur.

gumacuk15:
Same bug. Please fix it.

gumacuk15:
Also in "**Full screen**" height is not working!

djhvscf:
We need to add fitHeader override to support this. Working on it

AntoineMam:
Hi There. Any update for this Issue? #4869 ? thanks

UtechtDustin:
@AntoineMam Currently not, but a PR is welcome (if you have a bugfix).

AntoineMam:
> @AntoineMam Currently not, but a PR is welcome (if you have a bugfix).

Thanks, I'have calling <script type="text/javascript" src="https://unpkg.com/bootstrap-table@1.15.5/dist/extensions/filter-control/bootstrap-table-filter-control.min.js"></script> and height from last version 1.6.0 and this CDN running good, I'm not a JS dev but You can do a dif between you're twice version. I think hat you've done this way but... If I can help.

PR #5069, opened 2020-06-07
Fix for #5061 adding toggleFilterControl () method

marceloverdijk:
This PR is for issue #5061 adding a `toggleFilterControl ()` method.


UtechtDustin:
@marceloverdijk 
Please provide us an example using our [editor](live.bootstrap-table.com/).

marceloverdijk:
Hi @UtechtDustin can the example use the code from the PR? Or do you expect something else?

UtechtDustin:
@marceloverdijk yes, its possible.
Go on the Libraries page, then select `From GitHub src` on the option `Bootstrap Table source`.
Now you can set your Branch (with the repository) `marceloverdijk:toggle_filter_control` on the `GitHub src branch`.
After that you can create your example, save it and post the url of that example here.

marceloverdijk:
Great! I wasn’t aware. Will try to do it this evening.

marceloverdijk:
@UtechtDustin here is the example: https://live.bootstrap-table.com/code/marceloverdijk/3261

UtechtDustin:
There is a issue while filtering and then toggle the filter control.
Peek 2020-06-07 22-47

marceloverdijk:
Interesting, it works in my local project but indeed I can reproduce it in the example.
The setup of my table in my local project is a bit different. Let me check...

marceloverdijk:
Oke, this one works https://live.bootstrap-table.com/code/marceloverdijk/3262 but has the data define using `<tbody>` which I also have in my local project.

The [first (failing) example](https://live.bootstrap-table.com/code/marceloverdijk/3261) uses `data-url="json/data1.json"`.
So the issue is related to that and at least explains why it was working in my local project.
I'm new to bootstrap-table and did not use the `data-url` yet, do you maybe have an idea what the issue could be?

UtechtDustin:
No sorry, i would also have to debug that.

UtechtDustin:
Could be solution but we also have to "reset" the filter-control functions.
Because if you would filter for something, and then toggle the filter-controle (hide the controls) the table would be breaken (The filtered data would be displayed).

djhvscf:
> Could be solution but we also have to "reset" the filter-control functions.
> Because if you would filter for something, and then toggle the filter-controle (hide the controls) the table would be breaken (The filtered data would be displayed).

Yes, I think we can "reset" the filter controls and then hide/show it

marceloverdijk:
Yes I certainly agree this is a better approach then refresh options.

But there is more than only clearing the filters and then hide/show them.
Some users want the filter controls to be disabled by default. 
With refresh options this is possible by setting `data-filter-control = false` (or off course not setting the property at all). The refresh option will draw the filter controls, but if the filter controls are not drawn `data-filter-control = false` we cannot show them as they don't exists.

Should we introduce a new `data-filter-control-hidden` property to be able to show/hide the filter controls on init?

`data-filter-control = true` ==> show the filter control
`data-filter-control = true && data-filter-control = true` ==> hide filter control but create elements

djhvscf:
> Yes I certainly agree this is a better approach then refresh options.
> 
> But there is more than only clearing the filters and then hide/show them.
> Some users want the filter controls to be disabled by default.
> With refresh options this is possible by setting `data-filter-control = false` (or off course not setting the property at all). The refresh option will draw the filter controls, but if the filter controls are not drawn `data-filter-control = false` we cannot show them as they don't exists.
> 
> Should we introduce a new `data-filter-control-hidden` property to be able to show/hide the filter controls on init?
> 
> `data-filter-control = true` ==> create and show the filter controls
> `data-filter-control = true && data-filter-control = true` ==> create filter controls but hide them

Well in that case for me the data-filter-control-hidden is an overkill since you can simulate that in column configurations (by not setting the filter control)

UtechtDustin:
> Well in that case for me the data-filter-control-hidden is an overkill since you can simulate that in column configurations (by not setting the filter control)

As he already explaint the filter-control extension will not create Controls, events etc.
Which means we cant toggle that (without recreating the whole table).

So we have 2 ways to go:
- Use a new option e.g. `data-filter-control-hidden` to init the filter-control extension but dont show it.
- Recreate the whole table and just toggle on/off the filter-control extension.

djhvscf:
> > Well in that case for me the data-filter-control-hidden is an overkill since you can simulate that in column configurations (by not setting the filter control)
> 
> As he already explaint the filter-control extension will not create Controls, events etc.
> Which means we cant toggle that (without recreating the whole table).
> 
> So we have 2 ways to go:
> 
> * Use a new option e.g. `data-filter-control-hidden` to init the filter-control extension but dont show it.
> * Recreate the whole table and just toggle on/off the filter-control extension.

In that case I would prefer use the new option 'data-filter-control-hidden'.

marceloverdijk:
I refactored `toggleFilterControl` to not call the refresh but show/hide the element.
Also the new `data-filter-control-hidden` attribute is added.

New example: https://live.bootstrap-table.com/code/marceloverdijk/3426 (using local data from the html document)

But also the initial [example](https://live.bootstrap-table.com/code/marceloverdijk/3261) is now working /cc @UtechtDustin 


Note while working on it I noticed:

https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L484-L486

`UtilsFilterControl.getControlContainer(that).find('.filterControl').hide()` 
shouldn't this be:
`UtilsFilterControl.getControlContainer(that).find('.filter-control').hide()`?

site/docs/extensions/filter-control.md
```diff
@@ -33,6 +33,19 @@ Dependence if you use the datepicker option: [bootstrap-datepicker](https://gith
 
 - **Default:** `false`
 
+### filterControlHidden
```
wenzhixin, line 36: I think `filterControlVisible` is better than `filterControlHidden`,  like our ohter options.

src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -958,4 +964,15 @@ $.BootstrapTable = class extends $.BootstrapTable {
       }
     }
   }
+
+  toggleFilterControl () {
+    this.options.filterControlHidden = !this.options.filterControlHidden
+    const filterControl = UtilsFilterControl.getControlContainer(this).find('.filter-control')
```
wenzhixin, line 970: `const $filterControls`

UtechtDustin:
Please also (re)set the correct table header height.
https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L84

site/docs/extensions/filter-control.md
marceloverdijk: OK, will change it later today.

marceloverdijk:
> Please also (re)set the correct table header height.
> https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L84

@UtechtDustin where should that be called? Currently I see only here: https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L668

marceloverdijk:
> > Please also (re)set the correct table header height.
> > https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L84
> 
> @UtechtDustin where should that be called? Currently I see only here: https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L668

Never mind I see it now in the example. Will add it later today.

marceloverdijk:
`filterControlHidden` is now renamed to `filterControlVisible` (with default value `true`).
But I'm struggling with that height fix and css...

UtechtDustin:
@marceloverdijk 
Adjust the `fixHeaderCSS` function, check if filter-control is visible or not.
If it is visble: use the current code, if it is not visible: unset the height.

Now you can call the fixHeaderCSS function while toggling.

marceloverdijk:
> @marceloverdijk
> Adjust the `fixHeaderCSS` function, check if filter-control is visible or not.
> If it is visble: use the current code, if it is not visible: unset the height.
> 
> Now you can call the fixHeaderCSS function while toggling.

I think the `fixHeaderCSS` is only called if the table `height` is set, right?
But filter control with a fixed height seems to be broken... https://live.bootstrap-table.com/code/marceloverdijk/3439 (using `1.16.0`)


And we also have mis-alignment @djhvscf [mentioned](https://github.com/wenzhixin/bootstrap-table/pull/5069#pullrequestreview-430741014).

UtechtDustin:
@marceloverdijk 
Oh okay, i dont checked yet the function (i thought its called always).
Then you have to find a other way, maybe some of us three has time to help out the next days.

Yes the table is broken with a fixed height #4869

marceloverdijk:
@UtechtDustin think I found it :-)

For each column that has no filter control an empty div is created with `.no-filter-control`. As this takes up some space we got the misalignment.

On creation and toggle I'm now not only hiding/showing the `.filter-control` elements, but also the `.no-filter-control` elements and now the misalignment is gone.

Note there is a bit of misalignment, but I see that also with `1.16.0` ([example](https://live.bootstrap-table.com/code/marceloverdijk/3443)).

Here are also updated example links:

- https://live.bootstrap-table.com/code/marceloverdijk/3261 (default)
- https://live.bootstrap-table.com/code/marceloverdijk/3442 (visible = false)

Thanks for your patience btw 👍 




PS: the more I looked at the code I wonder if this is correct:
https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/filter-control/bootstrap-table-filter-control.js#L484-L486
The class name seems incorrect, and should't this hide the `.no-filter-control` elements as well?

UtechtDustin:
@marceloverdijk Thank you for your contributen and the good communication :-)
Yes there is the wrong class on L485, you can fix that.

After this is fixed i check your code/changes :-)

marceloverdijk:
Cool I changed L485 from `UtilsFilterControl.getControlContainer(that).find('.filterControl').hide()` to `UtilsFilterControl.getControlContainer(that).find('.filter-control, .no-filter-control').hide()`

UtechtDustin:
Before: https://live.bootstrap-table.com/code/UtechtDustin/3444
After: https://live.bootstrap-table.com/code/UtechtDustin/3445

site/docs/extensions/filter-control.md
```diff
@@ -29,10 +29,22 @@ Dependence if you use the datepicker option: [bootstrap-datepicker](https://gith
...
 
+### filterControlVisible
+
+- **Attribute:** `data-filter-control-visible`
+
+- **type:** `Boolean`
+
+- **Detail:**
+
+   Set to false to hide the filter controls.
```
UtechtDustin, line 44: Set to `false` to hide the filter controls.

site/docs/extensions/filter-control.md
```diff
@@ -254,6 +266,10 @@ Dependence if you use the datepicker option: [bootstrap-datepicker](https://gith
 
 * Clear all the controls added by this plugin (similar to showSearchClearButton option).
 
+### toggleFilterControl
+
+* Toggle filter control.
```
UtechtDustin, line 271: Some more info would be greate e.g.   
Toggles the filter controls (Toggles the visibility of the filter controls)

marceloverdijk:
@UtechtDustin Documentation updated as well. 
After this is merge I will look into #5065.

djhvscf:
Good work! Thanks

marceloverdijk:
Thanks a lot for merging this! I will look into #5065 after work this evening.

PR #5083, opened 2020-06-11
Compare strings to prevent datatyp issues

### LINKED ISSUE #5008: filterBy not working with a select element
When using a dropdown select to create a filter. there is either no results found or it doesnt filter anything at all. 
```javascript
   $filter.change(function(){
      let filter_value = this.selectedOptions[0].value
      $('#selectval').html(filter_value)
      $table.bootstrapTable('filterBy', {
        id: [filter_value]                    
      })
    })
```
I included this fiddle to show it. Its using mostly the defaults for the filterBy method example. 
 [Online Example](https://live.bootstrap-table.com/code/sonicdm/2716)

UtechtDustin:
Fix #5008

Before: https://live.bootstrap-table.com/code/UtechtDustin/3349
After: https://live.bootstrap-table.com/code/UtechtDustin/3350

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -804,6 +804,12 @@ class BootstrapTable {
         ? Utils.escapeHTML(this.searchText) : this.searchText).toLowerCase()
       const f = Utils.isEmptyObject(this.filterColumns) ? null : this.filterColumns
 
+      for (const key in f) {
+        for (const valueKey in f[key]) {
+          f[key][valueKey] = f[key][valueKey].toString()
+        }
+      }
+
       // Check filter
       if (typeof this.filterOptions.filterAlgorithm === 'function') {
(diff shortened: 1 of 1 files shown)


wenzhixin:
I don't think it's a good idea to support both int and string. In this case, just only need to update this line to solve the problem:

```js
let filter_value = +this.selectedOptions[0].value
```

Example: https://live.bootstrap-table.com/code/wenzhixin/3361

There are some bug after the changed:

* `filterBy` with not array: https://live.bootstrap-table.com/code/wenzhixin/3362
* `filterBy` boolean: for example, the field value is `true` or `'true'`, they are not the same.

UtechtDustin:
Okay, then please open a PR with that fix.
I dont even knew that we support that, because its not documentated.

PR #5085, opened 2020-06-11
Added the methods "collapseRowByUniqueId" and "expandRowByUniqueId"

### LINKED ISSUE #5000: Methods i feel are missing or inconsistent (Removing row by index,Expand/Collapse row by unique id)
I might be missing these methods, or easy implementation but I just can't seem to find them in the documentation and examples.

Removing row by index. - You insert and update row by index, however removing row is suddenly by definite field and value or unique Id. I often missed ability to remove a row by its index instead of having to look for a value in that one specific row so that i can remove it.

Get specific cell data by Column name and index - There might be a way to achieve this through getData but I haven't found a way yet.

Expand/Collapse row by unique id.

If all of these are already implemented but i just can't understand them from examples then I am sorry.

UtechtDustin:
Fix #5000
Added the methods "collapseRowByUniqueId" and "expandRowByUniqueId"

Example: https://live.bootstrap-table.com/code/UtechtDustin/3356

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2861,6 +2861,16 @@ class BootstrapTable {
     this.trigger('expand-row', index, row, $element)
   }
 
+  expandRowByUniqueId (uniqueId) {
+    const row = this.getRowByUniqueId(uniqueId)
+
+    if (!row) {
+      return
+    }
+
+    this.expandRow(this.data.indexOf(row))
+  }
+
   collapseRow (index) {
     const row = this.data[index]
     const $tr = this.$body.find(Utils.sprintf('> tr[data-index="%s"][data-has-detail-view]', index))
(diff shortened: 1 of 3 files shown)


src/bootstrap-table.js
```diff
@@ -2861,6 +2861,11 @@ class BootstrapTable {
     this.trigger('expand-row', index, row, $element)
   }
 
+  expandRowByUniqueId (uniqueId) {
+    const rowId = this.options.data.indexOf(this.getRowByUniqueId(uniqueId))
+    this.expandRow(rowId)
+  }
```
wenzhixin, line 2867: Updated to:

```js
    const row = this.getRowByUniqueId(uniqueId)

    if (!row) {
      return
    }

    this.expandRow(this.data.indexOf(row))
```

You will find the problem when you enable the search.

wenzhixin:
Improved the code: https://live.bootstrap-table.com/code/wenzhixin/3389

wenzhixin:
Added example: https://github.com/wenzhixin/bootstrap-table-examples/commit/7476926f87464caba05407f6d7c1895b8f0c5074

PR #5106, opened 2020-06-18
WIP: Cookie async url load

### LINKED ISSUE #5105: cookie filter using select control is loaded before url data populated control
1.16 and develop

**Description**  

When cookie extension is enabled, filter is a select control and keys do not match values and data is loaded from url, the option is added before options are populated.

relates with #1840

**Example**  

Before: https://live.bootstrap-table.com/code/albfan/3504
After: https://live.bootstrap-table.com/code/albfan/3502

**Possible  (optional)**  

See this solution is more than a hack. It assumes that a select filled with url data has an two options (one empty and other with the cookie value) It gets the cookie value from that second option, and use as selected option.

In the example is barely perceptible, but on a laggy service you can see the option with the value for a while, until it is correctly reloaded.

albfan:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5105
 
**Example(s)?**
Before: https://live.bootstrap-table.com/code/albfan/3504
After: https://live.bootstrap-table.com/code/albfan/3502     

See second commit is just a fixture for url data (not intented to be merged), but if there's a way to create a laggy service, the PoC will be more obvious.

This interrelates cookie and filter, as the filter extension should be aware there could be a cookie value for a select and the default selected needs to be loaded from there. That would be the real fix.


albfan:
I think this PR should be turn into a draft https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/#:~:text=To%20convert%20a%20pull%20request,you%20to%20confirm%20the%20change.

If fixes the problem, but needs a rework to be easily maintanable and less "hacky"

djhvscf:
@albfan reminder

albfan:
Using tag 1.16 I was able to reproduce the problem:

- I put a laggy service doin a "sleep" for 5 seconds (time enough so async success happens after cookie fill the selected value) https://github.com/albfan/bootstrap-table/commit/83fabd14dcce3748e7eb596e02163580f733f2bd

- And I added a json so I can call a custome url service https://github.com/albfan/bootstrap-table/commit/0594566298f762f192f5022f9d1b9b81cfd3dde7

   ```
   {
     "0": "Zero",
     "1": "One",
     "2": "Two",
     "3": "Three",
     "4": "Four"
   }
   ```

   **NOTE**: It could be a good addition to be able to define json data for the online editor same as we have that "json/data1.json", "json/data2.json", "json/data3.json" instead of call the raw github content like `data-filter-data="url:https://raw.githubusercontent.com/albfan/bootstrap-table/laggy-service-1.16/filter-data.json"`


There I can reproduce on demand the problem:

- https://live.bootstrap-table.com/code/albfan/3957:

   bootstrap-cookie-async-1 16

But after move to version 1.17.0 or develop I can only see this the first time and then I cannot reproduce anymore. There's some change I'm not aware of that does the magic. If I was able to understand that "one time error" will reopen but I think this can be mark as resolved.

PR #5112, opened 2020-06-19
Hotfix/init container template without newlines

molostoff:
**Bug fix?**
yes

**New Feature?**
no

Removes whitespaces before html tags and newlines in according to jquery 1.9.1 node creation semantics (e.g. '<div></div>' and not '     <div> \n\n </div>')

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -73,25 +73,12 @@ class BootstrapTable {
     const bottomPagination = ['bottom', 'both'].includes(this.options.paginationVAlign)
       ? '<div class="fixed-table-pagination"></div>' : ''
 
-    this.$container = $(`
-      <div class="bootstrap-table ${this.constants.theme}">
-      <div class="fixed-table-toolbar"></div>
-      ${topPagination}
-      <div class="fixed-table-container">
-      <div class="fixed-table-header"><table></table></div>
-      <div class="fixed-table-body">
-      <div class="fixed-table-loading">
-      <span class="loading-wrap">
-      <span class="loading-text">${this.options.formatLoadingMessage()}</span>
-      <span class="animation-wrap"><span class="animation-dot"></span></span>
-      </span>
-      </div>
-      </div>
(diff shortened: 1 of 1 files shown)


molostoff:
> Could you share the link where jQuery mentiones that?
https://api.jquery.com/jQuery/#jQuery2

if it looks like HTML (i.e., it starts with <tag ... >). If not, the string is interpreted as a selector expression
----
the error is

Uncaught Error: Syntax error, unrecognized expression: <div class="bootstrap-table bootstrap3">
      <div class="fixed-table-toolbar"></div>
      
      <div class="fixed-table-container">
      <div class="fixed-table-header"><table></table></div>
      <div class="fixed-table-body">
      <div class="fixed-table-loading">
      <span class="loading-wrap">
      <span class="loading-text">Loading, please wait</span>
      <span class="animation-wrap"><span class="animation-dot"></span></span>
      </span>
      </div>
      </div>
      <div class="fixed-table-footer"><table><thead><tr></tr></thead></table></div>
      </div>
      <div class="fixed-table-pagination"></div>
      </div>
    at Function.Sizzle.error (components.min.js?v1.13:4421)
    at tokenize (components.min.js?v1.13:5076)
    at select (components.min.js?v1.13:5460)
    at Function.Sizzle [as find] (components.min.js?v1.13:3998)
    at init.find (components.min.js?v1.13:5576)
    at new init (components.min.js?v1.13:196)
    at jQuery (components.min.js?v1.13:62)

molostoff:
> Thanks. Could you please include a working example??
at your own, please. just checkout hotfix branch and you can see it. :)

UtechtDustin:
Cant reproduce it with jquery 1 and jquery 2.
Just as note: jquery 1 and jquery 2 are eol.
in my opinion we can close this PR, what do you think @wenzhixin @djhvscf 

If we still should remove the whitespaces, we schould only remove the first and last whitespaces.
I also think there are a few more place which have spaces around the html.

djhvscf:
@UtechtDustin Agree

PR #5160, opened 2020-07-09
Added autoMergeCells option

### LINKED ISSUE #5128: Is it possible to define individual rowspan's when using a table from data?
This is probably an edge case but I thought let's check if it's possible.

When using Bootstrap-Table with content form data, like:

```
<table id="table">
  <thead>
    <tr>
      <th data-field="place">Place</th>
      <th data-field="address">Address</th>
      <th data-field="resident">Resident</th>
    </tr>
  </thead>
</table>

<script>
  var $table = $('#table')

  $(function() {
    var data = [
      {
        'place': 'Atlanta',
        'address': 'Street 1',
        'resident': 'Sarah'
      },
      {
        'place': 'Boston',
        'address': 'Street 1',
        'resident': 'Ben'
      },
      {
        'place': 'Boston',
        'address': 'Street 2',
        'resident': 'John Doe'
      },
      {
        'place': 'Boston',
        'address': 'Street 2',
        'resident': 'Jane Doe'
      }
    ]
    $table.bootstrapTable({data: data})
  })
</script>
``` 

I wonder if it's possible to set the `rowspan` on certain fields (which I will compute upfront).
So to only display once Atlanta and Boston for place name, and only once Street 2 for Boston.

I believe my best option is to use a table from HTML in this case, or are there other options?

Something like https://live.bootstrap-table.com/code/marceloverdijk/3645

wenzhixin:
**Bug fix?**
no

**New Feature?**
yes/no

**Resolve an issue?**
Fix #5128

**Example(s)?**
https://live.bootstrap-table.com/code/wenzhixin/3877

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ ChangeLog
 - **New:** Added `formatted` parameter for `getData` method to get formatted data.
 - **New:** Added `paginationParts` option instead of `onlyInfoPagination`.
 - **New:** Added `sortReset` option to reset sort on third click.
+- **New:** Added support for auto merge the table body cells.
 - **Update:** Fixed `updateByUniqueId` method cannot update multiple rows bug.
 - **Update:** Fixed `insertRow` not write to source data array bug.
 - **Update:** Fixed events bug with `detailViewIcon` option.
@@ -31,6 +32,7 @@ ChangeLog
 - **Update:** Fixed table `border` bug when table is hidden.
 - **Update:** Fixed `showRow` method show all hidden rows bug.
 - **Remove:** Removed the `onlyInfoPagination` option.
+- **Remove:** Removed accent neutralise extension and moved it to core.
 
 #### Extensions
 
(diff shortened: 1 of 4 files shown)


UtechtDustin:
Could you explain why we should add a new option for that ?
For tables based on html data is also no option necessary.

wenzhixin:
The option is necessary, you can check these two examples:
* https://live.bootstrap-table.com/code/wenzhixin/3879
* https://live.bootstrap-table.com/code/wenzhixin/3878

wenzhixin:
We can know that the `undefined` column needs to merge cells based on html (`fromHtml`).
But the data option we cannot know the `undefined` column needs to render `-` or needs to merge cells.

UtechtDustin:
But we could check if in the data is `_<field>_colspan` or `_<field>_rowspan` and then set "autoMergeCells" to true (internal).

wenzhixin:
Although this is fine, the `data` is required to ensure that the format is correct.

UtechtDustin:
That was my thought :-)

wenzhixin:
OK, I will update tomorrow.

wenzhixin:
Updated: https://live.bootstrap-table.com/code/wenzhixin/3887

marceloverdijk:
Brilliant to have the `_<field>_rowspan` option. Works perfectly 👍

rvshah512:
Hello @wenzhixin,

While using _<field>_rowspan to merge cells with dynamic data, how to re-merge when Search or Sort option is used? Help much appreciated.

PR #5165, opened 2020-07-10
Improved tools (check if the examples are valid)

UtechtDustin:
**Description**
This PR will add a new check to our tool, it checks if the example in our documentation are formatted correctly `[name](url)` and checks if the file exists on our examples repository.

@wenzhixin currently the tool use the github api directly, but the api has a limit of 60 requests per hour which could be exceeded while developing. You could write a script which cache the json(s) on our server (quite simple cronjob).

While developing i also fixed 5 broken examples, there are two more examples which are broken (this is btw also the reason why the ci fails for this PR):

- [html](https://bootstrap-table.com/docs/api/table-options/#html)
- [virtualScrollItemHeight](https://bootstrap-table.com/docs/api/table-options/#virtualscrollitemheight)

i dont know what that options do, maybe one of you could add the examples (or change the url if there is already a example).

**Bug fix?**
no

**New Feature?**
yes

**Resolve an issue?**
no

**Example(s)?**
no

### DIFF
--- a/tools/check-api.js
+++ b/tools/check-api.js
@@ -3,7 +3,29 @@ require = require('esm')(module)
 const fs = require('fs')
 const chalk = require('chalk')
 const Constants = require('../src/constants/index.js').default
+const XMLHttpRequest = require('xmlhttprequest').XMLHttpRequest
 let errorSum = 0
+
+function loadExampleLink (url) {
+  const request = new XMLHttpRequest()
+  request.open('GET', url, false)
+  request.send(null)
+
+  const filesJson = JSON.parse(request.responseText)
+  const fileNames = []
+
+  for (const [key, value] of Object.entries(filesJson)) {
+    fileNames.push(value.name)
+  }
+
+  return fileNames
+}
+
+let exampleFiles = []
+exampleFiles = exampleFiles.concat(loadExampleLink('https://api.github.com/repos/wenzhixin/bootstrap-table-examples/contents/options'))
+exampleFiles = exampleFiles.concat(loadExampleLink('https://api.github.com/repos/wenzhixin/bootstrap-table-examples/contents/methods'))
+exampleFiles = exampleFiles.concat(loadExampleLink('https://api.github.com/repos/wenzhixin/bootstrap-table-examples/contents/column-options'))
+exampleFiles = exampleFiles.concat(loadExampleLink('https://api.github.com/repos/wenzhixin/bootstrap-table-examples/contents/welcomes'))
 
 class API {
   constructor () {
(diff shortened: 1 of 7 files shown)


wenzhixin:
Updated the code, I think the `html` is unnecessary and never used, so I removed it.

UtechtDustin:
@wenzhixin I dont used the filesystem to prevent that every fork/contributor who use the command must have the other repository as well.
Maybe we should check if the repository was found on that location, if not ignore that check and just output a warning that this check was skipped because the repository is missing.

wenzhixin:
OK,  can you help to add a waring?

UtechtDustin:
@wenzhixin sure let me check that late :-)

UtechtDustin:
@wenzhixin Added a warning :-)

tools/check-api.js
```diff
@@ -4,6 +4,20 @@ const fs = require('fs')
...
+let exampleFiles = []
+if (exampleFilesFound) {
+  exampleFiles = [
+    ...fs.readdirSync(exampleFilesFolder + 'welcomes'),
+    ...fs.readdirSync(exampleFilesFolder + 'options'),
+    ...fs.readdirSync(exampleFilesFolder + 'column-options'),
+    ...fs.readdirSync(exampleFilesFolder + 'methods')
+  ]
+} else {
+  console.log((chalk.yellow(chalk.bold('Warning: ') + 'Cant check if example files are correct formatted and has a valid url.')))
```
djhvscf, line 18: have a valid...*

tools/check-api.js
```diff
@@ -4,6 +4,20 @@ const fs = require('fs')
...
+  console.log((chalk.yellow(chalk.bold('Warning: ') + 'To enable that check, please clone the "bootstrap-table-examples" repository in the tools folder or create a symlink (if you already cloned the repository on a other path).')))
```
djhvscf, line 19: an other...*

tools/check-api.js
```diff
@@ -39,7 +54,21 @@ class API {
           if (this.ignore && this.ignore[key] && this.ignore[key].includes(name)) {
             continue
           }
-          if (!details[i + 1] || details[i + 1].indexOf(`**${name}:**`) === -1) {
+
+          const tmpDetails = details[i + 1].trim()
+          if (name === 'Example' && exampleFilesFound) {
+            const matches = exampleRegex.exec(tmpDetails)
+            if (matches === null) {
```
djhvscf, line 61: Could we validate this like a falsy value?

`if (!matches) {....`

Issue #5168, opened 2020-07-10
Search input field type


nvp-lemay:
Currently, the search field activated by "data-search" creates an input field with the type of "text". While this seems fine on the outset, implementing a "search" field has many added benefits. And simply changing the type to "search" would allow other small HTML5 browser features to be displayed; such as the small "x" at the end of the input field which clears the field.

Could this be an option?

PR #5170, opened 2020-07-11
Change the input type to search instead of text

djhvscf:
**Bug fix?**
no

**New Feature?**
yes

**Resolve an issue?**
#5168 

**Example(s)?**
Base example: https://live.bootstrap-table.com/code/djhvscf/3916
Filter control example: https://live.bootstrap-table.com/code/djhvscf/3917

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -737,7 +737,7 @@ class BootstrapTable {
       )
       const searchInputHtml = `<input class="${this.constants.classes.input}
         ${Utils.sprintf(' %s%s', this.constants.classes.inputPrefix, opts.iconSize)}
-        search-input" type="text" placeholder="${opts.formatSearch()}" autocomplete="off">`
+        search-input" type="search" placeholder="${opts.formatSearch()}" autocomplete="off">`
       let searchInputFinalHtml = searchInputHtml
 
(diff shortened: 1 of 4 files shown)


src/bootstrap-table.js
```diff
@@ -757,7 +757,7 @@ class BootstrapTable {
       this.$toolbar.append(html.join(''))
       const $searchInput = this.$toolbar.find('.search input')
       const handleInputEvent = () => {
-        const eventTriggers = `keyup drop blur ${Utils.isIEBrowser() ? 'mouseup' : ''}`
+        const eventTriggers = 'keyup drop blur mouseup'
```
wenzhixin, line 760: No need to judge whether it is IE?

wenzhixin:
```
git checkout develop
git pull
git checkout feature/5168
git rebase develop
```

src/bootstrap-table.js
djhvscf: No, since using search as input type we need to listen the mouseup because that event is fired when the x icon is clicked

nvp-lemay:
Question on when the changes for this fix will actually make it to a release. It appears to have stalled in the develop branch and now is going on 2 months. Thanks for an update.

UtechtDustin:
@nvp-lemay i agree with you, we will release a new version soon.

PR #5178, opened 2020-07-13
Footer colspan and server response

### LINKED ISSUE #2276: colspan with footer
i created a example on jsfiddle
http://jsfiddle.net/kxzj2qj3/

is it possible to do a colspan only for the totalTextFormatter?

i tried 

```
function totalTextFormatter(data) {
  return '<colspan="3"> Total';
}
```

i would like to do something like 
[image table sum](http://www.excel-easy.com/data-analysis/images/tables/total-row.png)

### LINKED ISSUE #3046: No way to set footer values when server-side pagination is used

UtechtDustin:
**Description**
This PR adds the ability to use colspans in the footer.
I also added the ability to use the values from the footer (or from a server response) in the footerFormatter.

**Bug fix?**
no

**New Feature?**
yes

**Resolve an issue?**
Fix #3046
Fix #2276

**Example(s)?**
FooterFormatter with value from footer: https://live.bootstrap-table.com/code/UtechtDustin/3974
Footer colspan: https://live.bootstrap-table.com/code/UtechtDustin/3975
Footer multiple colspans: https://live.bootstrap-table.com/code/UtechtDustin/3976
Both works with paginationSide `Server`, an json should looks like:
```
{
  "rows": [
    {
      "id": 0,
      "name": "Item 0",
      "price": "$0",
      "amount": 3
    },
    {
      "id": 1,
      "name": "Item 1",
      "price": "$1",
      "amount": 4
    },
    {
      "id": 2,
      "name": "Item 2",
      "price": "$2",
      "amount": 8
    }
  ],
  "footer": {
    "id": "fid",
    "_id_colspan": 2,
    "name": "fname"
  }
}
```

### DIFF
--- a/site/docs/api/column-options.md
+++ b/site/docs/api/column-options.md
@@ -228,7 +228,7 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
   The function, takes two parameter:
 
   * `data`: Array of all the data rows.
-  * `value`: If footer data isset, the value of the footer column.
+  * `value`: If footer data is set, the value of the footer column.
 
   the function should return a string with the text to show in the footer cell.
 
--- a/site/docs/api/column-options.md
+++ b/site/docs/api/column-options.md
@@ -225,9 +225,10 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 
   The context (this) is the column Object.
 
-  The function, take one parameter:
+  The function, takes two parameter:
 
-  * `data`: Array of all the  data rows.
+  * `data`: Array of all the data rows.
+  * `value`: If footer data isset, the value of the footer column.
 
   the function should return a string with the text to show in the footer cell.
 
--- a/site/docs/api/column-options.md
+++ b/site/docs/api/column-options.md
@@ -225,7 +225,7 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 
   The context (this) is the column Object.
 
-  The function, takes two parameter:
+  The function, takes two parameters:
 
   * `data`: Array of all the data rows.
   * `value`: If footer data is set, the value of the footer column.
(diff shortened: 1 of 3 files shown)


UtechtDustin:
@djhvscf @wenzhixin
I need some help, the features works technically but the width of the colspans in the footer has not the same width of the columns of the table.
Maybe one of you could help with that :-)

site/docs/api/column-options.md
```diff
@@ -225,9 +225,10 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 
   The context (this) is the column Object.
 
-  The function, take one parameter:
+  The function, takes two parameter:
 
-  * `data`: Array of all the  data rows.
+  * `data`: Array of all the data rows.
+  * `value`: If footer data isset, the value of the footer column.
```
djhvscf, line 231: Is set*?

site/docs/api/column-options.md
```diff
@@ -225,9 +225,10 @@ The column options is defined in `jQuery.fn.bootstrapTable.columnDefaults`.
 
   The context (this) is the column Object.
 
-  The function, take one parameter:
+  The function, takes two parameter:
```
djhvscf, line 228: parameters *

PR #5185, opened 2020-07-16
fix a colspan issue while using the detailformatter

### LINKED ISSUE #5184: 1.17.1版本 Bug

**Description**  

启用 detailView，并添加操作列，当数据为空时，表格显示错位，详细如图
1.16.0 版本不存在该问题
**Example**  

image

**Possible  (optional)**  

**Additional context**

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5184 

**Example(s)?**
Before: https://live.bootstrap-table.com/code/UtechtDustin/4016
After: https://live.bootstrap-table.com/code/UtechtDustin/4017

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1545,7 +1545,7 @@ class BootstrapTable {
     // show no records
     if (!hasTr) {
       this.$body.html(`<tr class="no-records-found">${Utils.sprintf('<td colspan="%s">%s</td>',
-        this.getVisibleFields().length,
+        this.getVisibleFields().length + Utils.getDetailViewIndexOffset(this.options),
         this.options.formatNoMatches())}</tr>`)
     } else {
       if (!this.options.virtualScroll) {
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1538,7 +1538,7 @@ class BootstrapTable {
     // show no records
     if (!hasTr) {
       this.$body.html(`<tr class="no-records-found">${Utils.sprintf('<td colspan="%s">%s</td>',
-        this.getVisibleFields().length,
+        this.getVisibleColumnCount(),
```
wenzhixin, line 1541:
```
this.getVisibleFields().length +
  (Utils.hasDetailViewIcon(this.options) && this.options.detailViewAlign !== 'right' ? 1 : 0)

Comment (wenzhixin) [inline src/bootstrap-table.js:1541]:
```
this.getVisibleFields().length +
  (Utils.hasDetailViewIcon(this.options) && this.options.detailViewAlign !== 'right' ? 1 : 0)

wenzhixin:
We can use `Utils.hasDetailViewIcon` method.

src/bootstrap-table.js
UtechtDustin: I agree we can use the Utils.hasDetailViewIcon method, but we should use a function as wrapper for that.
If we have in the future new features which must be also checked, the code will be ugly.
So we can prevent it right now.

What do you think @wenzhixin

wenzhixin: OK, let me updated this PR.

wenzhixin:
Updated and fixed a new problem with `mergeCells`.

Before: https://live.bootstrap-table.com/code/wenzhixin/4125
After: https://live.bootstrap-table.com/code/wenzhixin/4126

UtechtDustin:
Looks good for me, good job.

PR #5222, opened 2020-07-29
Fixed sort bug when the field is 0

### LINKED ISSUE #5217: sortable column breaks starting in v1.17.0
The issue happens in 1.17.0 and 1.17.1. (It works fine in 1.16.0.)

**Description**  

Some sortable columns are unable to sort

**Example**  
The web site that demonstrates the problem is at the URL: http://www.showskimmer.com/?t=Pixar+movies&btv=1.17.1
Note that you can specify the version of bootstrap-table in the "btv" parameter. If that parameter is omitted, it defaults to 1.16.0, which is the working version.

To reproduce the problem... Note that the table has 3 sortable columns: Year, Rating and Title. The Rating and Title columns work fine. If you click the Rating column header twice, the table will respond by being sorted on the ratings in ascending order on the first click and descending order on the second click. Similarly, if you click the Title column header twice, the table will respond by being sorted on the titles in ascending order on the first click and descending order on the second click. But, if you click the Year column header twice, the first click sorts the table by year in ascending order, and the second click makes all of the sort column headers unresponsive (which is the bug).

I have reproduced this bug on the most recent versions of Safari/Mac, Chrome/Windows and Firefox/Windows.

wenzhixin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5217 

**Example(s)?**
Before: https://live.bootstrap-table.com/code/mikewertheim/4160
After: https://live.bootstrap-table.com/code/wenzhixin/4165

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2166,7 +2166,7 @@ class BootstrapTable {
       (
         this.searchText ||
         this.options.customSearch ||
-        this.options.sortName ||
+        this.options.sortName !== undefined ||
         !Utils.isEmptyObject(this.filterColumns) ||
         !Utils.isEmptyObject(this.filterColumnsPartial)
       ) && (!params || !params.unfiltered)


UtechtDustin:
Some more seconds and i would push my branch for that fix :+1:

wenzhixin:
@UtechtDustin  👍

PR #5232, opened 2020-07-31
Add support of clientSidePagination to addrbar extension

kerog:
**Bug fix?**
no

**New Feature?**
yes

**Resolve an issue?**
no

**Example(s)?**
I can't add example because editor haven't address bar.

### DIFF
--- a/src/extensions/addrbar/bootstrap-table-addrbar.js
+++ b/src/extensions/addrbar/bootstrap-table-addrbar.js
@@ -44,7 +44,7 @@ function _buildUrl (dict, url = window.location.search) {
     // 搜索name=general这种形式的字符串(&是分隔符)
     const pattern = `${key}=([^&]*)`
     const targetStr = `${key}=${val}`
-    if (targetStr.includes('undefined'))
+    if (val === undefined )
       continue
 
     /*
--- a/src/extensions/addrbar/bootstrap-table-addrbar.js
+++ b/src/extensions/addrbar/bootstrap-table-addrbar.js
@@ -63,6 +63,23 @@ function _buildUrl (dict, url = window.location.search) {
   return url
 }
 
+/*
+* function: _updateHistoryState
+* var _prefix = this.options.addrPrefix || ''
+* _updateHistoryState(_prefix)
+* returns void
+*/
+function _updateHistoryState ( _prefix ) {
+  const params = {}
+  params[`${_prefix}page`] = this.options.pageNumber,
+  params[`${_prefix}size`] = this.options.pageSize,
+  params[`${_prefix}order`] = this.options.sortOrder,
+  params[`${_prefix}sort`] = this.options.sortName,
+  params[`${_prefix}search`] = this.options.searchText
+  // h5提供的修改浏览器地址栏的方法
+  window.history.pushState({}, '', _buildUrl(params))
+}
+
 $.extend($.fn.bootstrapTable.defaults, {
   addrbar: false,
   addrPrefix: ''
@@ -72,7 +89,6 @@ $.BootstrapTable = class extends $.BootstrapTable {
   init (...args) {
     if (
       this.options.pagination &&
-      this.options.sidePagination === 'server' &&
       this.options.addrbar
     ) {
       // 标志位, 初始加载后关闭
@@ -86,23 +102,24 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
       const _prefix = this.options.addrPrefix || ''
       const _onLoadSuccess = this.options.onLoadSuccess
+      const _onPageChange = this.options.onPageChange
 
       this.options.onLoadSuccess = data => {
         if (this.addrbarInit) {
           this.addrbarInit = false
         } else {
-          const params = {}
-          params[`${_prefix}page`] = this.options.pageNumber,
-          params[`${_prefix}size`] = this.options.pageSize,
-          params[`${_prefix}order`] = this.options.sortOrder,
-          params[`${_prefix}sort`] = this.options.sortName,
-          params[`${_prefix}search`] = this.options.searchText
-          // h5提供的修改浏览器地址栏的方法
-          window.history.pushState({}, '', _buildUrl(params))
+          _updateHistoryState(_prefix)
         }
 
         if (_onLoadSuccess) {
           _onLoadSuccess.call(this, data)
+        }
+      }
+      this.options.onPageChange = (number,size) => {
+        _updateHistoryState(_prefix)
+
+        if (_onPageChange) {
+          _onPageChange.call(this,number,size)
         }
       }
     }
--- a/src/extensions/addrbar/bootstrap-table-addrbar.js
+++ b/src/extensions/addrbar/bootstrap-table-addrbar.js
@@ -44,6 +44,8 @@ function _buildUrl (dict, url = window.location.search) {
     // 搜索name=general这种形式的字符串(&是分隔符)
     const pattern = `${key}=([^&]*)`
     const targetStr = `${key}=${val}`
+    if (targetStr.includes('undefined'))
+      continue
 
     /*
      * 如果目标url中包含了key键, 我们需要将它替换成我们自己的val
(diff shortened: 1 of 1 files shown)


src/extensions/addrbar/bootstrap-table-addrbar.js
```diff
@@ -63,6 +63,23 @@ function _buildUrl (dict, url = window.location.search) {
...
 
+/*
+* function: _updateHistoryState
+* var _prefix = this.options.addrPrefix || ''
+* _updateHistoryState(_prefix)
+* returns void
+*/
+function _updateHistoryState ( _prefix ) {
+  const params = {}
+  params[`${_prefix}page`] = this.options.pageNumber,
```
UtechtDustin, line 74: That cant work, because of `this`.

UtechtDustin: Add `this` as parameter then it should work.
Its not the best way to solve it, but the extension must be rewrite anyways.

UtechtDustin: It also would be greate to check if `order`, `sort` and/or `search` are `empty` or `undefined`.
If that is the case we should not add them as url parameter (`order=undefined` looks bad).

UtechtDustin:
https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/addrbar/bootstrap-table-addrbar.js#L119-L121

This part will fail with the page size (because the pageSize is selectable), which means the url parameter is useless.

kerog:
> 
> 
> https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/addrbar/bootstrap-table-addrbar.js#L119-L121
> 
> This part will fail with the page size (because the pageSize is selectable), which means the url parameter is useless.

Only reason for keep this, is share full url between 2 users and initial set the bootstrap-table, of parameters like search, page, limit, etc.

UtechtDustin:
> Only reason for keep this, is share full url between 2 users and initial set the bootstrap-table, of parameters like search, page, limit, etc.

I think you missunderstand what i mean.
I dont mean to remove a url parameter, but we need a other solution for the pageSize/size parameter because currently you cant use them.

kerog:
> 
> 
> > Only reason for keep this, is share full url between 2 users and initial set the bootstrap-table, of parameters like search, page, limit, etc.
> 
> I think you missunderstand what i mean.
> I dont mean to remove a url parameter, but we need a other solution for the pageSize/size parameter because currently you cant use them.

Ok.

src/extensions/addrbar/bootstrap-table-addrbar.js
kerog: I've change call of this function and removes 'undefined' from urls

src/extensions/addrbar/bootstrap-table-addrbar.js
```diff
@@ -86,25 +105,26 @@ $.BootstrapTable = class extends $.BootstrapTable {
...
 
         if (_onLoadSuccess) {
           _onLoadSuccess.call(this, data)
         }
       }
+      this.options.onPageChange = (number,size) => {
+        _updateHistoryState(this,_prefix)
+
+        if (_onPageChange) {
+          _onPageChange.call(this,number,size)
```
djhvscf, line 125: I think we can use apply in order to send all arguments.

Something like yourFunction.apply(...args)

src/extensions/addrbar/bootstrap-table-addrbar.js
```diff
@@ -63,6 +65,24 @@ function _buildUrl (dict, url = window.location.search) {
...
+* returns void
+*/
+function _updateHistoryState ( table,_prefix ) {
+  const params = {}
+  params[`${_prefix}page`] = table.options.pageNumber,
+  params[`${_prefix}size`] = table.options.pageSize,
+  params[`${_prefix}order`] = table.options.sortOrder,
+  params[`${_prefix}sort`] = table.options.sortName,
+  params[`${_prefix}search`] = table.options.searchText
+  // h5提供的修改浏览器地址栏的方法
```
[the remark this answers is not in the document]
kerog: I only extract this part of code to new function, to remove duplicates function call.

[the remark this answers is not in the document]
UtechtDustin: You can delete it, this comment is not needed (used a translator).

src/extensions/addrbar/bootstrap-table-addrbar.js
kerog: I do it, like in L104-106.

src/extensions/addrbar/bootstrap-table-addrbar.js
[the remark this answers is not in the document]
kerog: Ok I remove it.

src/extensions/addrbar/bootstrap-table-addrbar.js
```diff
@@ -44,6 +44,8 @@ function _buildUrl (dict, url = window.location.search) {
     // 搜索name=general这种形式的字符串(&是分隔符)
     const pattern = `${key}=([^&]*)`
     const targetStr = `${key}=${val}`
+    if (targetStr.includes('undefined'))
```
UtechtDustin, line 47: `typeof targetStr === undefined`

kerog: This don't works because targetStr is string like this
```
sort=undefined
```

UtechtDustin: oh your right, then `val === undefined` should work i guess.

kerog: Ok, I change this line.

src/extensions/addrbar/bootstrap-table-addrbar.js
kerog: Ok, I changed this.

src/extensions/addrbar/bootstrap-table-addrbar.js
```diff
@@ -63,6 +65,23 @@ function _buildUrl (dict, url = window.location.search) {
...
 }
 
+/*
+* function: _updateHistoryState
+* var _prefix = this.options.addrPrefix || ''
+* var table = this
+* _updateHistoryState( table,_prefix)
+* returns void
+*/
+function _updateHistoryState ( table,_prefix ) {
```
wenzhixin, line 75: I think we can move this function into BootstrapTable class.

kerog: This function use _buildUrl from L42-L66

UtechtDustin: @wenzhixin do you mean the whole extension or just this function ?
What would be improvement if we would add this to the core (if you dont mean the whole extension)

kerog: I mean only the function _buildUrl is using by _updateHistoryState, and if you plan move function _updateHistoryState you must also move _buildUrl.

src/extensions/addrbar/bootstrap-table-addrbar.js
```diff
@@ -44,6 +44,8 @@ function _buildUrl (dict, url = window.location.search) {
     // 搜索name=general这种形式的字符串(&是分隔符)
     const pattern = `${key}=([^&]*)`
     const targetStr = `${key}=${val}`
+    if (val === undefined )
```
UtechtDustin, line 47: Could be changed to the normal syntax with `{}`, but as the extension should be rewritten someday its ok.

djhvscf:
@wenzhixin @kerog any update here?

kerog:
> 
> 
> @wenzhixin @kerog any update here?

Hi I'm waiting for your accept of changes (I change 'call' to 'action')

djhvscf:
Ok, please resolve the conversations first

kerog:
done

wenzhixin:
Merged into develop: https://github.com/wenzhixin/bootstrap-table/commits/develop

PR #5235, opened 2020-08-04
Fixed an issue with vertical scrolling event

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
https://stackoverflow.com/questions/63239938/bootstrap-table-sticky-header-doesnt-move-left-right-with-table-content-when-sc

**Example(s)?**
Scroll to the right (the header will not be positioned)
Before: https://live.bootstrap-table.com/code/UtechtDustin/4247
After: https://live.bootstrap-table.com/code/UtechtDustin/4248

### DIFF
--- a/src/extensions/sticky-header/bootstrap-table-sticky-header.js
+++ b/src/extensions/sticky-header/bootstrap-table-sticky-header.js
@@ -56,6 +56,11 @@ $.BootstrapTable = class extends $.BootstrapTable {
       .on('scroll', () => this.renderStickyHeader())
   }
 
+  horizontalScroll () {
+    super.horizontalScroll()
+    this.$tableBody.on('scroll', () => this.matchPositionX())
+  }
+
   renderStickyHeader () {
     const that = this
     this.$stickyHeader = this.$header.clone(true, true)


wenzhixin:
It seems that the problem is caused by the `showFoter` option: https://live.bootstrap-table.com/code/wenzhixin/4249

UtechtDustin:
Yes it is, as described in the SO post.

src/extensions/sticky-header/bootstrap-table-sticky-header.js
```diff
@@ -42,6 +42,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
     $(window).off(resizeEvent).on(resizeEvent, () => this.renderStickyHeader())
     $(window).off(scrollEvent).on(scrollEvent, () => this.renderStickyHeader())
     this.$tableBody.off('scroll').on('scroll', () => this.matchPositionX())
+    this.$tableBody[0].onscroll = () => this.matchPositionX()
```
wenzhixin, line 45:
```js    
    if (this.options.showFooter) {
      this.$tableBody[0].onscroll = () => this.matchPositionX()
    }
```

PR #5239, opened 2020-08-05
The checked property is ignored on checkbox columns where the field has a value

TantumLabs:
The checked property is ignored if the field has a value.

For example:
I am setting the checkbox column to the Id field so there is always a value. If I set the formatter to checked: false the checkbox is still checked because the field has a value.

`function checkboxFormatter(){
return {
checked: false,
disabled: true
}`

**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**

**Example(s)?**
https://live.bootstrap-table.com/code/TantumLabs/4276

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1454,7 +1454,8 @@ class BootstrapTable {
         type = column.radio ? 'radio' : type
 
         const c = column['class'] || ''
-        const isChecked = (typeof(value.checked) !== 'undefined' ? value.checked : (value === true || value_ || value && value.checked) && value !== false)
+        const isChecked = Utils.isObject(value) && value.hasOwnProperty('checked') ?
+          value.checked : (value === true || value_) && value !== false
         const isDisabled = !column.checkboxEnabled || (value && value.disabled)
 
         text = [
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1454,7 +1454,7 @@ class BootstrapTable {
         type = column.radio ? 'radio' : type
 
         const c = column['class'] || ''
-        const isChecked = (value === true || value_ || (value && value.checked)) && value !== false
+        const isChecked = (typeof(value.checked) !== 'undefined' ? value.checked : (value === true || value_ || value && value.checked) && value !== false)
         const isDisabled = !column.checkboxEnabled || (value && value.disabled)
 
         text = [
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -15,6 +15,10 @@ export default {
     })
     return flag ? str : ''
   },
+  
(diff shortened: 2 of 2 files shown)


src/utils/index.js
```diff
@@ -15,6 +15,10 @@ export default {
     })
     return flag ? str : ''
   },
+  
+  isObject (val) {
+    return val !== null && typeof val === 'object' && !Array.isArray(val)
```
djhvscf, line 20: Could we use `Object.prototype.toString.call(val) === '[object Object]'`?

UtechtDustin: Maybe `val instanceof Object` ? :-)

src/bootstrap-table.js
```diff
@@ -1454,7 +1454,8 @@ class BootstrapTable {
         type = column.radio ? 'radio' : type
 
         const c = column['class'] || ''
-        const isChecked = (value === true || value_ || (value && value.checked)) && value !== false
+        const isChecked = Utils.isObject(value) && value.hasOwnProperty('checked') ?
+          value.checked : (value === true || value_) && value !== false
```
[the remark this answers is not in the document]
TantumLabs: That worked!

UtechtDustin:
@TantumLabs Please check the conversations!
@djhvscf asked something, so your answer "it works" makes no sense, we dont only want working code, we also want good code!
Also you ignored my suggestion for the `isObject` function.

src/utils/index.js
```diff
@@ -17,7 +17,7 @@ export default {
   },
   
   isObject (val) {
-    return val !== null && typeof val === 'object' && !Array.isArray(val)
+    return val instanceof Object && !Array.isArray(val)
```
UtechtDustin, line 20: if `val` is a `object`, its obviously not an `Array`.
So the second check can be removed.

src/bootstrap-table.js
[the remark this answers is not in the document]
djhvscf: What do you mean? Did you test my suggestion?

src/utils/index.js
djhvscf: @TantumLabs  any news?

djhvscf:
@TantumLabs any update here?

src/utils/index.js
wenzhixin: It's OK: https://live.bootstrap-table.com/code/wenzhixin/4670

src/bootstrap-table.js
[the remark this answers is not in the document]
wenzhixin: We cannot use `value || value_` instead, because the `value` maybe `undefined`.

wenzhixin (approved):

PR #5267, opened 2020-08-20
fixed height table : auto scroll to detail-icon clicked

Abbas10:
fixed height table
-------------------
**detail view enhancement**
- auto scroll to expanded row at top of the fixed-table-body, when user click on expand icon for the detail view.


UtechtDustin:
You edited the dist file(s) instead of the src file, just as info.

djhvscf:
@Abbas10 please create a new PR with your change and also remember to add an online example to test your code.

PR #5283, opened 2020-09-08
Supporting style

### LINKED ISSUE #5270: How to keep table element attributes after launching bootstraptable?
Hi, I'm encountering a small issue referring table element attributes. It's simple like this:

[html file]
`<thead> ... </thead>`
`<tbody>`
`...`
`<td style="background-color:red">...</td>`
`...`
`</tbody>`
`</table>`

[js file]
`$("#mytable").bootstrapTable();`

This js script will actually remove the background color of that td, and I'd like to keep it.

Any help would be appreciated!

djhvscf:
**Bug fix?**
No

**New Feature?**
Yes

**Resolve an issue?**
Fix #5270

**Example(s)?**
https://live.bootstrap-table.com/code/djhvscf/4561

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -139,6 +139,7 @@ class BootstrapTable {
     }
 
     this._headerTrClasses = []
+    this._headerTrStyles = []
     this.$header.find('tr').each((i, el) => {
       const $tr = $(el)
       const column = []
@@ -162,6 +163,9 @@ class BootstrapTable {
 
       if ($tr.attr('class')) {
         this._headerTrClasses.push($tr.attr('class'))
+      }
+      if ($tr.attr('style')) {
+        this._headerTrStyles.push($tr.attr('style'))
       }
     })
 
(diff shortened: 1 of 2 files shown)


src/bootstrap-table.js
```diff
@@ -1336,6 +1340,7 @@ class BootstrapTable {
       Utils.sprintf(' %s', htmlAttributes.length ? htmlAttributes.join(' ') : undefined),
       Utils.sprintf(' id="%s"', Array.isArray(item) ? undefined : item._id),
       Utils.sprintf(' class="%s"', style.classes || (Array.isArray(item) ? undefined : item._class)),
+      Utils.sprintf(' style="%s"', Array.isArray(item) ? undefined : item._style),
```
UtechtDustin, line 1343: Im not sure what happens if you apply `undefined` on the `sprintf` function.
Could you check that ?

UtechtDustin (changes requested):

src/bootstrap-table.js
djhvscf: It returns empty when the parameters are undefined https://jsfiddle.net/djhvscf/xv0f6cyd/

UtechtDustin: So it would make more sense to check it with a if to prevent empty style attributes (`style=""`).

wenzhixin: It's OK: https://github.com/wenzhixin/bootstrap-table/blob/develop/src/utils/index.js#L2

wenzhixin (approved):

src/bootstrap-table.js
UtechtDustin: oh okay, i thought %s would be replaced with an empty string :D

PR #5296, opened 2020-09-14
Enable CSS transitions for loading (fixes #5295)

### LINKED ISSUE #5295: Enable CSS transitions for showLoading/hideLoading
**Description**  
When loading data, a loading template is displayed, which currently works by toggling the CSS `display` property from `none`/`flex`.

I'd like to use CSS transitions with the loading template, such as `ease-in-out`, which would make the appearance of the loading template seem "smoother". However, CSS transitions don't work with the `display` property

As an alternative, the CSS `visibility` and `opacity` properties can be used instead, for example:

```css
.fixed-table-loading {
    display: flex;
    transition: visibility 0s, opacity 0.15s ease-in-out;
    opacity: 0;
    visibility: hidden;
}

.fixed-table-loading.open {
    visibility: visible;
    opacity: 1;
}
```

The idea here is that instead of toggling a single CSS property on the element, you toggle the `open` class on the element. This approach also makes it easier for end-users to customise further, as they can simply override the `open` class with their own CSS (customisation is actually difficult now, since a CSS property is set on the element itself).

This is actually pretty straightforward to achieve, even moreso since `loadingTemplate` made it's debut in #5003 - I really appreciate this library, and I'm happy to contribute with a PR for this, if you want it?

cocowalla:
**Bug fix?**
no

**New Feature?**
yes

**Resolve an issue?**
Fix #5295

**Examples**
https://live.bootstrap-table.com/code/cocowalla/4642

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2845,7 +2845,7 @@ class BootstrapTable {
   }
 
   showLoading () {
-    this.$tableLoading.css('display', 'flex')
+    this.$tableLoading.toggleClass('open', true)
 
     let fontSize = this.options.loadingFontSize
 
@@ -2860,7 +2860,7 @@ class BootstrapTable {
   }
 
   hideLoading () {
-    this.$tableLoading.css('display', 'none')
+    this.$tableLoading.toggleClass('open', false)
   }
 
   togglePagination () {
--- a/src/themes/_theme.scss
+++ b/src/themes/_theme.scss
@@ -199,12 +199,20 @@
       .fixed-table-loading {
         align-items: center;
         background: $background;
-        display: none;
+        display: flex;
         justify-content: center;
         position: absolute;
         bottom: 0;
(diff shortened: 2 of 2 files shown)


UtechtDustin:
Could you please provide us two examples (before/after) using our [editor](https://live.bootstrap-table.com/) ?
You maybe have to copy your css changes in a style part, because the editor will not recompile the css.

cocowalla:
@UtechtDustin I've added an example [here](https://live.bootstrap-table.com/code/cocowalla/4642), but to see it working you will need to click **Libraries** and enter this as the branch: `cocowalla:feat/loading-enhancement` - for some reason the branch keeps reverting back to `develop`.

UtechtDustin:
Thanks for the contribution, it seems the editor is buggy.
@wenzhixin could you please check and fix that ?

UtechtDustin (approved):

PR #5298, opened 2020-09-14
use the main dataset instead of the current one

### LINKED ISSUE #5278: After deleting the selected row, the deleted row can be found when querying the selected row
Check the source code and find that the operation is this.options.data When getselections, the query is this.data Is this a bug or does it need special treatment?

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5278

**Example(s)?**
Bug1:
Check the first checkbox, delete the row.
Do the same for the second row => Deletion is not working

Bug2:
Check the first checkbox, delete the row.
Click on "getSelections", there is still the removed row.

Before: https://live.bootstrap-table.com/code/UtechtDustin/4645
After: https://live.bootstrap-table.com/code/UtechtDustin/4646

This issue also removes the `getAllSelections` which no more needed, because of the option [maintainMetaData](https://bootstrap-table.com/docs/api/table-options/#maintainmetadata).
If you have arguments against it, please explain them.

### DIFF
--- a/site/docs/api/methods.md
+++ b/site/docs/api/methods.md
@@ -167,16 +167,6 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
 - **Example:** [Filter By](https://examples.bootstrap-table.com/#methods/filter-by.html)
 
-## getAllSelections
-
-- **Parameter:** `undefined`
-
-- **Detail:**
-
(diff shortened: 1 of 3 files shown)


djhvscf (approved):
For me it is ok, but would like to hear something from @wenzhixin

wenzhixin (approved):

PR #5302, opened 2020-09-15
fix/5282

### LINKED ISSUE #5282: Duplicated data when adding rows to empty tables
1.17.1

**Description** 
After using the method "insertRow" to add data to an empty table, if you use it again to add a row to another empty table, the data will be duplicated.

**Example**  

Steps to reproduce:
1. Press the button "insertRow" from the first table
2. Press the button "insertRow" from the second table
3. New data from first table is also added to the second table

https://live.bootstrap-table.com/code/joaossantos/4558

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5282

**Example(s)?**
Before: https://live.bootstrap-table.com/code/joaossantos/4558
After: https://live.bootstrap-table.com/code/UtechtDustin/4658

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -407,7 +407,7 @@ class BootstrapTable {
     } else if (type === 'prepend') {
       this.options.data = [].concat(data).concat(this.options.data)
     } else {
-      data = data || this.options.data
+      data = data || [...this.options.data]
       this.options.data = Array.isArray(data) ? data : data[this.options.dataField]
     }
 
--- a/site/docs/api/table-options.md
+++ b/site/docs/api/table-options.md
@@ -1146,7 +1146,7 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
 
 - **Default:** `false`
 
-- **Example:** [Search Selector](https://examples.bootstrap-table.com/#options/table-search-selector.html)
+- **Example:** [Search Selector](https://examples.bootstrap-table.com/#options/search-selector.html)
 
 ## searchText
 


src/bootstrap-table.js
```diff
@@ -192,6 +192,10 @@ class BootstrapTable {
       })
     })
 
+    if (this.options.data.length === 0) {
```
djhvscf, line 195: Not sure why this line fix the sync between two tables

djhvscf (changes requested):

src/bootstrap-table.js
UtechtDustin: I'm also not sure, but without that javascript will use a shared array.
You also dont have the issue if one of the table have data (one row is enough), because there use not shared arrays.

djhvscf: So, if we change something in this.options both tables will be affected?

UtechtDustin: This only occurs in this edge case (Both table has no init data(rows) also this only happens with `this.options.data`.

djhvscf: ok, thanks @UtechtDustin . @wenzhixin do you have any idea? I think the fix is good because it is working now but it looks to me like a workaround

UtechtDustin: Just a explanation for you @djhvscf, if you do
```
let x = []
let y = x
y.push('test')
alert(x)
```
Example: https://live.bootstrap-table.com/code/UtechtDustin/4659

You will see `test`, because javascript dont copy the array. Instead of a copying, javascript will set the internal reference/pointer of the array (variable x) to variable y. Which means both variables share the same array.

I'm not sure where this happens in the bootstrap-table so this PR is just a workaround to prevent bugs.

djhvscf: Yes, i know that. But we should remove those kind of problems since every table should be isolated

UtechtDustin: oh okay, i thought you dont know that because of you question :D
> So, if we change something in this.options both tables will be affected?

wenzhixin: Thanks, @UtechtDustin and @djhvscf , I will check this problem.

wenzhixin:
I find the problem, if `options.data` of the table is not set, it will use `DEFAULTS.data` by default, they are the same.

djhvscf (approved):

PR #5314, opened 2020-09-20
Check if the buttons option returned an array of buttons

### LINKED ISSUE #5313: strange behaviour using the new button feature
**Description**  

trying to implement your new feature (buttons in the upper right section), i am struggeling about this problem.

of course only the new version (v1.18.0) is used and caches have been cleared.

i am not able to reconstruct this problem in your example-code-system, but my local example with this problem is similar to https://live.bootstrap-table.com/code/snoopy72/4831, so i tried nothing too special :)

i am using these settings for the following screenshot (rest of code like in my example 4831):

```js
buttons: "buttonsFunction",
buttonsOrder: "['btntest','refresh','columns','export','autoRefresh']",
```
bt001

reading the html with the browser developer tools i get:

```html
<div class="columns columns-right btn-group pull-right">
   <button class="btn btn-default" type="button" name="refresh" aria-label="Ansicht neu laden" title="Ansicht neu laden"><i class=" las la-sync"></i> </button>
   <div class="keep-open btn-group" title="Spalten auswählen">
      <button class="btn btn-default dropdown-toggle" type="button" data-toggle="dropdown" aria-label="Columns" title="Spalten auswählen">
        <i class=" las la-columns"></i>
        <span class="caret"></span>
      </button>
      <ul class="dropdown-menu" role="menu">
         <li class="dropdown-item-marker" role="menuitem"><label>[cut the 10 field to shorten code]</label></li>
      </ul>
   </div>
   <div class="export btn-group">
      <button class="btn btn-default dropdown-toggle" aria-label="Export" data-toggle="dropdown" type="button" title="Daten exportieren">
        <i class=" las la-download"></i>
        <span class="caret"></span>
      </button>
      <ul class="dropdown-menu" role="menu">
         <li role="menuitem" class="" data-type="csv"><a href="#">CSV</a></li>
         <li role="menuitem" class="" data-type="txt"><a href="#">TXT</a></li>
         <li role="menuitem" class="" data-type="excel"><a href="#">MS-Excel</a></li>
         <li role="menuitem" class="" data-type="xml"><a href="#">XML</a></li>
      </ul>
   </div>
   <button class="btn btn-default" type="button" name="0"></button>
   <button class="btn btn-default" type="button" name="1"></button>
   <button class="btn btn-default" type="button" name="2"></button>
   <button class="btn btn-default" type="button" name="3"></button>
   <button class="btn btn-default" type="button" name="4"></button>
   <button class="btn btn-default" type="button" name="5"></button>
   <button class="btn btn-default" type="button" name="6"></button>
   <button class="btn btn-default" type="button" name="7"></button>
   <button class="btn btn-default" type="button" name="8"></button>
   <button class="btn btn-default" type="button" name="9"></button>
   <button class="btn btn-default" type="button" name="10"></button>
   <button class="btn btn-default" type="button" name="11"></button>
   <button class="btn btn-default" type="button" name="12"></button>
   <button class="btn btn-default" type="button" name="13"></button>
   <button class="btn btn-default" type="button" name="14"></button>
   <button class="auto-refresh btn btn-default" type="button" name="autoRefresh" title="Auto"><i class=" las la-clock"></i></button>
</div>
```

that list of 15 times `<button class="btn btn-default" type="button" name="0"></button>` is irritation for me.

disabling the table-options "buttons" and "buttonsOrder", the buttons are displayed like normal (then of course without my new button ...). so this behaviour seems to has its origin within the button feature.

disabling only the option "buttonsOrder" keeps that strange buttons and only changes the ordering (what is not wondering):
bt002

my fault ?

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5313

**Example(s)?**
Before: https://live.bootstrap-table.com/code/UtechtDustin/4832
After:  https://live.bootstrap-table.com/code/UtechtDustin/4833

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -687,26 +687,28 @@ class BootstrapTable {
       let buttonHtml
 
       if (buttonConfig.hasOwnProperty('html')) {
-        buttonHtml = Utils.calculateObjectValue(opts, buttonConfig.html)
+        if (typeof buttonConfig.html === 'function') {
+          buttonHtml = buttonConfig.html()
+        } else if (typeof buttonConfig.html === 'string') {
+          buttonHtml = buttonConfig.html
+        }
       } else {
         buttonHtml = `<button class="${this.constants.buttonsClass}" type="button" name="${buttonName}"`
 
         if (buttonConfig.hasOwnProperty('attributes')) {
           for (const [attributeName, value] of Object.entries(buttonConfig.attributes)) {
-            buttonHtml += ` ${attributeName}="${Utils.calculateObjectValue(opts, value)}"`
+            buttonHtml += ` ${attributeName}="${value}"`
           }
         }
 
         buttonHtml += '>'
 
         if (opts.showButtonIcons && buttonConfig.hasOwnProperty('icon')) {
-          const icon = Utils.calculateObjectValue(opts, buttonConfig.icon)
-
-          buttonHtml += `${Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, icon) } `
+          buttonHtml += `${Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, buttonConfig.icon) } `
         }
 
         if (opts.showButtonText && buttonConfig.hasOwnProperty('text')) {
-          buttonHtml += Utils.calculateObjectValue(opts, buttonConfig.text)
+          buttonHtml += buttonConfig.text
         }
 
         buttonHtml += '</button>'
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -46,7 +46,10 @@ class BootstrapTable {
       Utils.sprintf(`${buttonsPrefix}%s`, opts.iconSize)
     ].join(' ').trim()
 
-    this.buttons = Utils.calculateObjectValue(this, opts.buttons, [], [])
+    this.buttons = Utils.calculateObjectValue(this, opts.buttons, [], {})
+    if (typeof this.buttons !== 'object') {
+      this.buttons = {}
+    }
   }
 
   initLocale () {
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -140,7 +140,7 @@ export default {
     return this.cachedWidth
   },
 
-  calculateObjectValue (self, name, args = [], defaultValue) {
+  calculateObjectValue (self, name, args, defaultValue) {
     let func = name
 
     if (typeof name === 'string') {


wenzhixin:
I think this line is no need to change: https://github.com/wenzhixin/bootstrap-table/commit/93c4b0f5e6bd0799b01b0e92f2533a3fee9b2e9a#diff-8968c62593320351ced6a0ef776ef599R139.

Improved the code.

src/bootstrap-table.js
```diff
@@ -46,7 +46,7 @@ class BootstrapTable {
       Utils.sprintf(`${buttonsPrefix}%s`, opts.iconSize)
     ].join(' ').trim()
 
-    this.buttons = Utils.calculateObjectValue(this, opts.buttons, [], [])
+    this.buttons = Utils.calculateObjectValue(opts, opts.buttons, undefined, [])
```
UtechtDustin: This and Line 170 must be reverted, i also tested it but it cause some issue with other features i guess.

src/bootstrap-table.js
```diff
@@ -690,20 +690,18 @@ class BootstrapTable {
 
         if (buttonConfig.hasOwnProperty('attributes')) {
           for (const [attributeName, value] of Object.entries(buttonConfig.attributes)) {
-            buttonHtml += ` ${attributeName}="${Utils.calculateObjectValue(opts, value)}"`
+            buttonHtml += ` ${attributeName}="${value}"`
```
UtechtDustin, line 693: If we merge this change, its no more possile to use functions!

src/bootstrap-table.js
```diff
@@ -690,20 +690,18 @@ class BootstrapTable {
...
         if (opts.showButtonIcons && buttonConfig.hasOwnProperty('icon')) {
-          const icon = Utils.calculateObjectValue(opts, buttonConfig.icon)
-
-          buttonHtml += `${Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, icon) } `
+          buttonHtml += `${Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, buttonConfig.icon) } `
         }
 
         if (opts.showButtonText && buttonConfig.hasOwnProperty('text')) {
-          buttonHtml += Utils.calculateObjectValue(opts, buttonConfig.text)
+          buttonHtml += buttonConfig.text
```
UtechtDustin, line 704: Same here

src/bootstrap-table.js
```diff
@@ -690,20 +690,18 @@ class BootstrapTable {
...
           }
         }
 
         buttonHtml += '>'
 
         if (opts.showButtonIcons && buttonConfig.hasOwnProperty('icon')) {
-          const icon = Utils.calculateObjectValue(opts, buttonConfig.icon)
-
-          buttonHtml += `${Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, icon) } `
+          buttonHtml += `${Utils.sprintf(this.constants.html.icon, opts.iconsPrefix, buttonConfig.icon) } `
```
UtechtDustin, line 700: same here

src/utils/index.js
```diff
@@ -167,6 +167,7 @@ export default {
 
     if (
       !func &&
+      args &&
```
UtechtDustin, line 170: I also tested this but i guess this could breake some setups!

UtechtDustin:
@wenzhixin please check the conversation!

src/bootstrap-table.js
wenzhixin: image

I don't think we need to support function here

UtechtDustin: I think we should support it, because there are usecases where a static string is not enough (e.g. to use as text `user` or `users` based on some data).
So we should improve the documentation instead of remove this feature.

wenzhixin (approved):

PR #5331, opened 2020-09-23
Use the cellStyle option also for the card view

### LINKED ISSUE #5324: data-cell-style apply in cardview
**Description**  

The attribute data-cell-style is working perfect in table-view but when change to card-view the spplied style disapeared, for example:

in table view function cellStyle works, add class "hecha" or "desconocida" to corresponding cell

`1) <td class="hecha" style="">Realizada</td>`

`2) <td class="desconocida" style="">Desconocida</td>`

but if i change to card view, this class is not applied to the corresponding cell so the data has no style:

`1) <div class="card-view"><span class="card-view-title" style="">Estado</span><span class="card-view-value">Realizada</span></div>`
`2) <div class="card-view"><span class="card-view-title" style="">Estado</span><span class="card-view-value">Desconocida</span></div>`

It would be very interesting that the cell-style would be applied to the data in both views, so the data with css style will look same.

NOTE: sorry for my bad English.

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5324 

**Example(s)?**
Before: https://live.bootstrap-table.com/code/UtechtDustin/4886
After: https://live.bootstrap-table.com/code/UtechtDustin/4887

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1607,9 +1607,9 @@ class BootstrapTable {
 
         if (this.options.cardView) {
           const cardTitle = this.options.showHeader ?
-            `<span class="card-view-title"${style_}>${Utils.getFieldTitle(this.columns, field)}</span>` : ''
+            `<span class="card-view-title ${cellStyle.classes}"${style_}>${Utils.getFieldTitle(this.columns, field)}</span>` : ''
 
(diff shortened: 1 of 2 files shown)


i-m-d:
Great news for the fix !!!!!

djhvscf (approved):

PR #5340, opened 2020-09-24
Collapsing detail view on collapsing groups

### LINKED ISSUE #5339: Group By v2 Collapsing does not collapse detail view expanded row
1.18.0

**Description**  
Utilizing detail view and group by v2 extension, if a row detail view is expanded, and then the group by row is clicked to collapse, the detail row is not collapsed

**Example**  
https://live.bootstrap-table.com/code/togradywk/4905

**Possible  (optional)**  
I resolved this issue by amending line 7639 in bootstrap-table.js:
        `if ($tr.attr('data-parent-index')) {
          $tr.after('<tr class="detail-view" data-parent-index="${$tr.attr('data-parent-index')}"><td 
                      colspan="${$tr.children('td').length}"></td></tr>');
        } else {
          $tr.after(Utils.sprintf('<tr class="detail-view"><td colspan="%s"></td></tr>', $tr.children('td').length));
        }`
Not sure if its acceptable to accommodate extensions in the primary bootstrap-table js.  

**Additional context**  
Chromium 83

UtechtDustin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5339

**Example(s)?**
Before: https://live.bootstrap-table.com/code/UtechtDustin/4906
After: https://live.bootstrap-table.com/code/UtechtDustin/4907

### DIFF
--- a/src/extensions/group-by-v2/bootstrap-table-group-by.js
+++ b/src/extensions/group-by-v2/bootstrap-table-group-by.js
@@ -190,16 +190,14 @@ BootstrapTable.prototype.initBody = function (...args) {
     if (this.options.groupByToggle) {
       this.$container.off('click', '.groupBy')
         .on('click', '.groupBy', function () {
-          const groupIndex = $(this).closest('tr').data('group-index')
+          const $this = $(this)
+          const groupIndex = $this.closest('tr').data('group-index')
+          const $groupRows = that.$body.find(`tr[data-parent-index=${groupIndex}]`)
 
-          $(this).toggleClass('expanded collapsed')
-          $(this).find('span').toggleClass(`${that.options.icons.collapseGroup} ${that.options.icons.expandGroup}`)
-
-          that.$body.find(`tr[data-parent-index=${groupIndex}]`).toggleClass('hidden')
-
-          const indexes = $(`tr[data-parent-index="${groupIndex}"]`).map((i, element) => $(element).data('index'))
-
-          indexes.each(index => that.collapseRow(index))
+          $this.toggleClass('expanded collapsed')
+          $this.find('span').toggleClass(`${that.options.icons.collapseGroup} ${that.options.icons.expandGroup}`)
+          $groupRows.toggleClass('hidden')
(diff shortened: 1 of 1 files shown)


src/extensions/group-by-v2/bootstrap-table-group-by.js
```diff
@@ -190,9 +190,16 @@ BootstrapTable.prototype.initBody = function (...args) {
     if (this.options.groupByToggle) {
       this.$container.off('click', '.groupBy')
         .on('click', '.groupBy', function () {
+          const groupIndex = $(this).closest('tr').data('group-index')
```
djhvscf, line 193: Now that we're here could we make a const variable for `$(this)` something like `$this` and then replace it in this click function?

src/extensions/group-by-v2/bootstrap-table-group-by.js
```diff
@@ -190,9 +190,16 @@ BootstrapTable.prototype.initBody = function (...args) {
...
+
           $(this).toggleClass('expanded collapsed')
           $(this).find('span').toggleClass(`${that.options.icons.collapseGroup} ${that.options.icons.expandGroup}`)
-          that.$body.find(`tr[data-parent-index=${$(this).closest('tr').data('group-index')}]`).toggleClass('hidden')
+
+          that.$body.find(`tr[data-parent-index=${groupIndex}]`).toggleClass('hidden')
```
djhvscf, line 198: I think also that we can save this in a const variable and use it here and in the next line

src/extensions/group-by-v2/bootstrap-table-group-by.js
```diff
@@ -190,9 +190,16 @@ BootstrapTable.prototype.initBody = function (...args) {
...
+
+          const indexes = $(`tr[data-parent-index="${groupIndex}"]`).map((i, element) => $(element).data('index'))
```
djhvscf, line 200: Instead of map we could use foreach or each and call the collapseRow method in the same loop.

djhvscf (changes requested):

djhvscf (approved):

PR #5352, opened 2020-09-29
Feature/5351 shorted locale

### LINKED ISSUE #5351: Question about shorted lacale
**Description**  
Hi, in PR#1033 (https://github.com/wenzhixin/bootstrap-table/pull/1033) was introduced shorted way to set locale instead of 
`locale: 'en-US'`
 just simple 
`locale: 'en'`
In version 1.18.0 I cannot set shorted version of locale eg. 'pl' instead 'pl-PL'.

**Example (optional)**  
Can't provide example (in sandbox locale is not loading (always in English))

kerog:
**Bug fix?**
no

**New Feature?**
yes

**Resolve an issue?**
Fix #5351

**Example(s)?**
Sandbox not working with localization only shows English.

Additional add missing translations for Polish language pack.


src/locale/bootstrap-table-pl-PL.js
```diff
@@ -1,11 +1,11 @@
 /**
  * Bootstrap Table Polish translation
- * Author: zergu <michal.zagdan @ gmail com>
+ * Author: kerogos <kerog @ wp pl>
```
wenzhixin, line 3:
```js
 * Author: zergu <michal.zagdan @ gmail com>
 * Update: kerogos <kerog @ wp pl>
```

djhvscf (changes requested):
Please change the header of the  file that @wenzhixin mentioned

UtechtDustin:
Also it would be greate if you add a sentence to the documentation which inform about this "feature" :-)

kerog:
> 
> 
> Also it would be greate if you add a sentence to the documentation which inform about this "feature" :-)

Ok, I found file Readme.md in src/locale folder, i add there information about all locale with short codes and missing translations.

Can I add simply reference to the this file in documentation?

UtechtDustin:
The Readme.md is the wrong file, you can find the Documentation here: https://github.com/wenzhixin/bootstrap-table/blob/develop/site/docs/api/localizations.md

Just add an Sentence like `Each Localization has a long and a short name, e.g. "af-ZA" and "af"`, we dont need a reference to this issue :-)

kerog:
During search of all locale I found in bootstrap-table locale dir translation for the non existing locale es-SP. I leave it, but you must consider delete it.

src/locale/bootstrap-table-pl-PL.js
kerog: I changed it.

djhvscf:
How can we manage the localisations like es-CR? Since we have a lot of es-*

Any idea?

kerog:
> 
> 
> How can we manage the localisations like es-CR? Since we have a lot of es-*
> 
> Any idea @kerog ?

David, If you have locale differ from es-ES than you must use full locale eg 'es-CR', or if someone do translation for Deutch language for Austria, then he must use full locale 'de-AT', 'de' is reserved for 'de-DE'.

In src/locale/Readme.md I propose how to add new language based on https://www.alchemysoftware.com/livedocs/ezscript/Topics/Catalyst/Language.htm

djhvscf:
Ok, I was thinking that we could use the CR of es-CR as a identifier also.

The problem is what happen if we find more than one coincidence?

kerog:
> 
> 
> Ok, I was thinking that we could use the CR of es-CR as a identifier also.
> 
> The problem is what happen if we find more than one coincidence?

I think we must set short code only for most popular language, like en is short form for en-US, unless there will be problem with eg. Argentinian Spanish (es-AR) witch is short form of Arabian (Saudi Arabia).
We must use code from iso-639-1, eg CR is not for Costa Rica. There is language Cree

kerog:
Hi, what is wrong with travis-ci?
I'm waiting almost 24h for status from  travis. :(

UtechtDustin:
@wenzhixin Please check that.

djhvscf:
I just restarted the build. Let's wait

djhvscf (approved):

PR #5401, opened 2020-10-24
Fix for bug #5400

### LINKED ISSUE #5400: group-by-formatter not working
1.18 and 1.17 tested

**Description**  
The data-group-by-formatter is not working at all. The value set in the html attribute is ignored. 

**Example**  
Use
https://live.bootstrap-table.com/example/extensions/group-by-v2.html

Add data-group-by-formatter value of "myFunc" to the table and any javascript function to the scripts section, such as:
`function myFunc(a,b,c)
  {
    return "MMM";
  }`

The table view is unchanged - expected result is to have "MMM" in the group rows.

**Possible  cause**  
In my opinion the problem originates from [the following lines in the code of _bootstrap-table-group-by.js_](https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/group-by-v2/bootstrap-table-group-by.js#L155)

`if (typeof (that.options.groupByFormatter) === 'function') {
        formattedValue = that.options.groupByFormatter(item.name, item.id, item.data)
      }`

The type of _groupByFormatter_ will always be 'string' as HTML data attributes are of string type. Therefore the check always fails and the formatter function is never called.

CaelumNigre:
**Bug fix?**
Yes

**New Feature?**
No

**Resolve an issue?**
Fix #5400

**Example(s)?**
I was unable to use bootstrap-table online testbed, so here are working examples:
bug demo:
 http://tpcg.io/mttmbnIW

bug fixed demo
 http://tpcg.io/rGchGEvo

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -189,7 +189,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
         keys.forEach(key => {
           const thisColumn = that.columns[that.fieldsColumnsIndex[key]]
           const fval = (fp[key] || '').toLowerCase()
-          let value = Utils.getItemField(item, key, false)
+          let value = Utils.unescapeHTML(Utils.getItemField(item, key, false))
           let tmpItemIsExpected
 
           if (fval === '') {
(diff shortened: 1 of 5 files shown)


wenzhixin (approved):

wenzhixin:
Thanks for your contribution!
https://live.bootstrap-table.com/code/wenzhixin/5202

christianwengert:
This is cool and I could really need it. Is there a release planned so this is included?

Thanks for your work!

PR #5414, opened 2020-11-02
Bugfix/fix 5403

### LINKED ISSUE #5403: Incorrect row unchecked on unchecking after applying search filter.
**Bootstraptable version(s) affected**: v1.18.0

**Description**  
Steps to reproduce: 
1. Select all rows using checkbox.
2. Search for some keyword.
3. Uncheck a row from the search results.
4. Clear search box.
5. The table unchecks using index, not value. So, if first row was unchecked from search results, then first row is also unchecked from selections made in Step 1.

**Example**  
https://live.bootstrap-table.com/code/supmanyu/5191

djhvscf:
**Bug fix?**
Yes

**New Feature?**
No

**Resolve an issue?**
Fix #5403 

**Example(s)?**
From URL: https://live.bootstrap-table.com/code/djhvscf/5281
From Data: https://live.bootstrap-table.com/code/djhvscf/5282
From HTML: https://live.bootstrap-table.com/code/djhvscf/5283

### DIFF
--- a/site/docs/api/methods.md
+++ b/site/docs/api/methods.md
@@ -647,6 +647,7 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
   Update the localizations format text.
 
 - **Example:** [Update Format Text](https://examples.bootstrap-table.com/#methods/update-format-text.html)
+
 ## updateRow
 
 - **Parameter:** `params`
@@ -660,3 +661,13 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
   * `replace` (optional): set to `true` to replace the row instead of extending.
 
 - **Example:** [Update Row](https://examples.bootstrap-table.com/#methods/update-row.html)
+
+## dataItem
+
+- **Parameter:** `jQuery row element`
+
+- **Detail:**
+
+  Returns the data item to which the specified table row is bound.
+
+- **Example:** [Data Item](https://examples.bootstrap-table.com/#methods/data-item.html)
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -443,7 +443,7 @@ class BootstrapTable {
 
   initDataUID () {
     for (let i = 0; i < this.options.data.length; i++) {
-      const uid = Utils.getRowGUID()
+      const uid = Utils.getRowUID()
 
       this.options.data[i]._uid = uid
       this.data[i]._uid = uid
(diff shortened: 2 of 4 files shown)


UtechtDustin:
Please resolve the CI Issues

djhvscf:
@UtechtDustin they are related to the docs. Please review the examples repo and approve my pr :)

djhvscf:
https://github.com/wenzhixin/bootstrap-table-examples/pull/418

djhvscf:
https://github.com/wenzhixin/bootstrap-table-examples/pull/418

wenzhixin:
I don't think to add `_uid` to a row is a good idea, the bug is caused by this: https://github.com/wenzhixin/bootstrap-table/pull/5298/files. Let me have a check.

djhvscf:
@wenzhixin why do you think that? For me is a good approach to link a `tr` to a specific object in our `data` array.

djhvscf:
@wenzhixin any input here? I would like to discuss why _uid is not a good idea

wenzhixin:
It will change the original data: https://live.bootstrap-table.com/code/wenzhixin/5314

djhvscf:
Oh, yes. Actually we're iterating over both data arrays. But we'd iterate over the this.data array only and see how this behaves

wenzhixin:
I think the key issue is the bug caused by the previous modification, which did not exist before v1.18.0: https://github.com/wenzhixin/bootstrap-table/pull/5298/files.
In addition, if there is a `_uid` field in his data, it may conflict. My suggestion is to find a better solution and re-evaluate it.

djhvscf:
> I think the key issue is the bug caused by the previous modification, which did not exist before v1.18.0: https://github.com/wenzhixin/bootstrap-table/pull/5298/files.
> In addition, if there is a `_uid` field in his data, it may conflict. My suggestion is to find a better solution and re-evaluate it.

I got you, but I think the main problem here is that this.options.data and this.data are arrays that we use for different purposes but sometimes we use them in the wrong way.
For example, this.options.data should be immutable for me, and this.data should be the one which is filtered and etc..

What do you think @wenzhixin @UtechtDustin ?

UtechtDustin:
I agree with @wenzhixin, i think there is a better solution for that.

djhvscf:
Well, could you give us some ideas? I'm not sure that I'm very clear about the issue with this solution.

@wenzhixin @UtechtDustin

wenzhixin:
I think it's a better solution for the problem: #5437

djhvscf:
ok, close this pr @wenzhixin @UtechtDustin

Issue #1393, opened 2015-09-09
Accessibility keyboard when sorting with <th>


aureliet:
Hello Wenzhinxin (and the others),

I don't see anything, in order to make the table accessible, when we want to sort it with the `<th>` like in my example : 
http://jsfiddle.net/6nj4nnwu/

Did I miss something ?

I see it is possible, in https://www.datatables.net/

On the datatables.net tables, when I am on a `<th>` through the tabulation, when I press Enter, it sorts it.

You will say "So why don't you use datatables ?"
Well... I tried it too, but I really prefer your bootstrap-table ;-)

Thank you very much,

Aurélie

djhvscf:
I think this is so easy to implement, let me try it @aureliet

aureliet:
Thank you, @djhvscf ! :-)

And I'm trying to understand how to make work the fixed columns, which are presented here, too ;-)
http://issues.wenzhixin.net.cn/bootstrap-table/#extensions/fixed-columns.html

I'm new on Github, so maybe I don't search as I should do.

djhvscf:
Did you check this? https://github.com/wenzhixin/bootstrap-table-examples/blob/master/extensions/fixed-columns.html

aureliet:
Yeah, thank you ! :-)

djhvscf:
See this: http://jsfiddle.net/e3nk137y/3062/ @aureliet

aureliet:
Hello @djhvscf 
I'll try it tomorrow, when I'll be at work, to see if I can make it work with my project :-)
Thank you again :-)

aureliet:
Impeccable ! it works very well on my project !
Thank you again, @djhvscf !

PR #5432, opened 2020-11-13
fix/5427

### LINKED ISSUE #5427: Filter control with sortable
1.18

**Description**  
I have data-search, data-sortable, data-filter-control option. In data-search, I have used search on enter key. so when I try to search on a column-based there I need to hit enter to search, but when I hit the enter key the column gets searched with text and the same column which is getting sorted. The sort should not happen.

**Example**  
https://live.bootstrap-table.com/code/rahulpattanshetty/5375

data-search-on-enter-key="true" should not be applied to data-filter-control input box

UtechtDustin:
Maybe i miss something, but i dont know why we should need that code.
We only have inputs in the header which can be selected with the keyboard (shift).

**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #5427

**Example(s)?**
Before: https://live.bootstrap-table.com/code/rahulpattanshetty/5375
After: https://live.bootstrap-table.com/code/UtechtDustin/5401


djhvscf:
This is to support tab and space behavior over the headers. I think we shouldn't remove this

wenzhixin:
Agree with @djhvscf

UtechtDustin:
Could you please explain that ?
Im not able (or don't know how) to sort with tab/space.

djhvscf:
Well... after a review looks like that functionality was removed. We usually add tabindex to the th elements in order to support tab+enter = sort

UtechtDustin:
So the question is, can we merge this PR ?
This code will cause the bug of #5427, the sort via space+tab is currently not possible and i don't think that a lot of ppl will sort via tab+space.

djhvscf (approved):

djhvscf:
For my perspective it is ok. @wenzhixin what do you think?

UtechtDustin:
@wenzhixin  What do you think ?

UtechtDustin:
@wenzhixin Still waiting :-)

wenzhixin:
The related issue and commit: #1393  https://github.com/wenzhixin/bootstrap-table/commit/7e489d06cd0b87c4736010dd2e62735abc3805f6

UtechtDustin:
@wenzhixin the issue and the commit are from 2015.
As this feature is gone for years and nobody complains about it we should decide if this feature makes sense (as it also cause # #5427).
I dont know anybody who use the tabulator to navigate to a table header for sorting.

What do you think @wenzhixin @djhvscf ?

djhvscf:
Well, we can remove that feature but let's add it as a breaking change in our release notes

UtechtDustin:
@djhvscf i don't think we have to note that, as this feature stopped work a long time ago.

djhvscf:
Got it

UtechtDustin:
@wenzhixin still waiting :-)

wenzhixin (approved):

PR #5443, opened 2020-11-16
Fix the bug of multiSelectRow

Yang-Ming-Wang:
Hi, I Just found if a user select a certain range of table raw (like select index 8 and 11),
multiSelectRow doesn't work expectedly. 

Fix it by using int compare rather than alphabetical-ordered Array.sort.

If you have any concern, please let me know. Thanks.

**Bug fix?**
yes

**New Feature?**
no

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2830,9 +2830,10 @@ class BootstrapTable {
 
     if (this.options.multipleSelectRow) {
       if (this.multipleSelectRowShiftKey && this.multipleSelectRowLastSelectedIndex >= 0) {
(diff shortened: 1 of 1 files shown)


UtechtDustin:
Thank you for the PR, could you please create two example (before and after the PR) using our [editor] (https://live.bootstrap-table.com/) ? :-)

Yang-Ming-Wang:
Sure, Here is the examples.

Before: https://live.bootstrap-table.com/example/options/multiple-select-row.html
After: https://live.bootstrap-table.com/code/Yang-Ming-Wang/5427

UtechtDustin:
Works fine, thank you very much!

UtechtDustin (approved):

PR #5489, opened 2020-12-17
Fix multi-sort

Naruto-kyun:
this.data is change, then this.initBody() load this.options.data, so reorder this.options.data instead

Bug fix?
yes

New Feature?
no

Resolve an issue?
Fix wenzhixin#4616

Example(s)?
Do manual asc on both columns
Before: https://live.bootstrap-table.com/code/wenzhixin/853
After: https://live.bootstrap-table.com/code/Naruto-kyun/5802

### DIFF
--- a/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
+++ b/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
@@ -744,7 +744,7 @@ BootstrapTable.prototype.onMultipleSort = function () {
     return cmp(arr1, arr2)
   }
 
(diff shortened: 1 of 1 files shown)


UtechtDustin (changes requested):
Sure that would fix this problem, but it would create new issues.
With youre change the original data would be sorted!

We have to find a other way to solve that, but thanks for the try :-)

wenzhixin:
@UtechtDustin is right, for example, use the `sortReset` option will show the problem: https://live.bootstrap-table.com/code/wenzhixin/5832.

PR #5494, opened 2020-12-17
keep opened details on update/insert

aureliendavid:
Hello, 

I have refreshed the following pull request: https://github.com/wenzhixin/bootstrap-table/pull/4571

The goal is to keep detail panes opened when using `updateByUniqueId()` and `insertRow()`

This is related to these issues: #1899, #1498, #1603

* The issue can be tested here: https://live.bootstrap-table.com/code/aureliendavid/5814

With official release when a row is expanded and then we click one of the button to insert or update a row, the opened details panes get closed.

* The fix can be tested here:  https://live.bootstrap-table.com/code/aureliendavid/5815

With this version of the lib (based on this PR) the details panes stay opened when clicking the buttons.


src/bootstrap-table.js
```diff
@@ -1653,15 +1653,34 @@ class BootstrapTable {
...
+        if (uniqueId && item.hasOwnProperty(uniqueId)) {
+          const itemUniqueId = item[uniqueId]
+
+          const oldTr = this.$body.find(Utils.sprintf('> tr[data-uniqueid="%s"][data-has-detail-view]', itemUniqueId))
+
+          if (oldTr.next().is('tr.detail-view')) {
+
+            toExpand.push(i)
+
+            if (updatedUid && itemUniqueId === updatedUid) {
```
djhvscf, line 1677: There is a empty `if`. Let's invert the logic to remove the `else` section

src/bootstrap-table.js
```diff
@@ -1653,15 +1653,34 @@ class BootstrapTable {
...
       hasTr = hasTr || !!tr
       if (tr && typeof tr === 'string') {
+
+        const uniqueId = this.options.uniqueId
+        if (uniqueId && item.hasOwnProperty(uniqueId)) {
+          const itemUniqueId = item[uniqueId]
+
+          const oldTr = this.$body.find(Utils.sprintf('> tr[data-uniqueid="%s"][data-has-detail-view]', itemUniqueId))
+
+          if (oldTr.next().is('tr.detail-view')) {
```
djhvscf, line 1673: Let's save `oldTr.next()` in a const variable just in case we need to use some lines after

src/bootstrap-table.js
```diff
@@ -3112,14 +3135,14 @@ class BootstrapTable {
...
 
-    if ($tr.next().is('tr.detail-view')) {
-      return
-    }
-
     if (this.options.detailViewIcon) {
       $tr.find('a.detail-icon').html(Utils.sprintf(this.constants.html.icon, this.options.iconsPrefix, this.options.icons.detailClose))
     }
 
+    if ($tr.next().is('tr.detail-view')) {
```
djhvscf, line 3142: Why are we moving this?

djhvscf (changes requested):

djhvscf (changes requested):
Review CI errors https://travis-ci.org/github/wenzhixin/bootstrap-table/jobs/750253501

src/bootstrap-table.js
aureliendavid: The idea of moving this is to be able to call `expandRow()` in the cases where we already copied the whole details pane html. In this case expandRow will just switch the details icon to "minus" and return. Otherwise we would either have an expanded row with a "plus" icon, or we'd have to copy the icon switching code somewhere else.

aureliendavid:
requested changes made in b0b3d85

djhvscf (approved):

djhvscf:
Looks good to me. @wenzhixin @UtechtDustin please take a look

thedoudou07:
Hello @aureliendavid, I am using your fix and the panels stay opened while updating a parent cell as you said but I loose the events/functions (columns editable, row collapsible) attached to the table. Do you have any idea ?

UtechtDustin:
> Looks good to me. @wenzhixin @UtechtDustin please take a look

I'm fine with that, but we should be aware that this solution should not be forever.
As on some other events the detail view will be still collapsed.

@aureliendavid Thank you for the contribution!

PR #5504, opened 2020-12-21
Add support for flat JSON to filter-control

### LINKED ISSUE #5503: Extension filter-control: Automatic population of select input column filters doesn't work woth flat JSON
1.18.1 with bootstrap v4.5.3

**Description**  
When using nested JSON objects as input data (flat JSON mode) for the bootstrap table filter control does not automatically populate the select input of the filter control in the column heading. The select stays empty. It does work fine with 'normal' JSON data as evidenced by the filter-control example.

**Example**  
https://live.bootstrap-table.com/code/ch1lln/5930

**Cause**  
The problem is caused by line 251 in src\extensions\filter-control\utils.js
`let fieldValue = data[i][field]`

Here field will contain the flat JSON 'path' like 'object.value' and this will not work with the above code. I will try to whip up a quick fix.

ch1lln:
Added helper function to get a value from a nested object by a flat json path.

**Bug fix?**
Yes

**New Feature?**
No

**Resolve an issue?**
Fix #5503 

**Example(s)?**
Before: https://live.bootstrap-table.com/code/ch1lln/5930
After: https://live.bootstrap-table.com/code/ch1lln/5933

### DIFF
--- a/src/extensions/filter-control/utils.js
+++ b/src/extensions/filter-control/utils.js
@@ -228,9 +228,9 @@ export function hasSelectControlElement (selectControl) {
 export function getObjectValueByString (obj, memberpath) {
   var pathComponents = memberpath.split('.')
   for (var i = 0; i < pathComponents.length; ++i) {
-      var k = pathComponents[i]
-      if (k in obj) {
-          obj = obj[k]
+      var path = pathComponents[i]
+      if (path in obj) {
+          obj = obj[path]
       } else {
           return null
       }
--- a/src/extensions/filter-control/utils.js
+++ b/src/extensions/filter-control/utils.js
@@ -248,7 +248,7 @@ export function initFilterSelectControls (that) {
 
       for (let i = 0; i < z; i++) {
         // Added a new value
-        let fieldValue = Utils.getItemField(data[i],field,false)
+        let fieldValue = Utils.getItemField(data[i], field, false)
         const formatter = that.options.editable && column.editable ? column._formatter : that.header.formatters[j]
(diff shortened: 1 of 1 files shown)


src/extensions/filter-control/utils.js
```diff
@@ -225,7 +225,18 @@ export function isFilterDataNotGiven ({ filterData }) {
 export function hasSelectControlElement (selectControl) {
   return selectControl && selectControl.length > 0
 }
-
+export function getObjectValueByString (obj, memberpath) {
```
djhvscf, line 228: I think we could move this to Utils file.. Not this one, the main one

src/extensions/filter-control/utils.js
```diff
@@ -225,7 +225,18 @@ export function isFilterDataNotGiven ({ filterData }) {
...
+  var pathComponents = memberpath.split('.')
+  for (var i = 0; i < pathComponents.length; ++i) {
+      var k = pathComponents[i]
```
djhvscf, line 231: Let use `path` as variable name

djhvscf (changes requested):

src/extensions/filter-control/utils.js
ch1lln: Thanks for the pointer, I checked src/Utils/index.js and actually found a function that does what we need.

I have no overview over the codebase so I didn't know it existed.

I have removed my function and used the one already in global utils.

The change is now a one-liner 👍

djhvscf (approved):

Issue #5532, opened 2021-01-07
sticky-header offsetY ignores value


atomkind:
**Bootstraptable version(s) affected**:

v1.18.1

**Description**  

if sticky-header offsetY has any value > 0 it just put the header about 55px y-offset from the top and not with the value given in `stickyHeaderOffsetY: '2'`
regardles of the value, it put the sticky-header always on the same position.

**Example**  

[https://live.bootstrap-table.com/code/atomkind/6154](https://live.bootstrap-table.com/code/atomkind/6154)

**Possible  (optional)**  

**Additional context**

syedsanahassan:
@wenzhixin needs to add a "px" at this line for this offset to work properly

https://github.com/wenzhixin/bootstrap-table/blob/develop/src/extensions/sticky-header/bootstrap-table-sticky-header.js#L111

Why?
```top: 0``` works but if it's a number other than 0 then we need a ```top: 60px```

PR #5591, opened 2021-02-19
Updated the stickyHeader offset options to number

### LINKED ISSUE #4755: sticky header offsetY not working
Hello Thanks for this project - First time using it today and its really useful! 

I have been trying the sticky Header extention and it seems to be working - Unless you try and add the stickyHeaderOffsetY 

using Local host - Uniserver - apache / php and Google chrome 

Downloaded today build to 1.15.5  

Using config   
`$table.bootstrapTable({
      stickyHeader: true,
      stickyHeaderOffsetY: 0 ,
      stickyHeaderOffsetLeft: '3em',
      stickyHeaderOffsetRight: '3em',
    });`

Outputs 
  Table looks good 
image

On scroll - Sticky seems fine .. needs offset 

image

checking on inspect : top:0px inline style
image
 

Update Config : 
`  $table.bootstrapTable({
      stickyHeader: true,
      stickyHeaderOffsetY: 60 ,
      stickyHeaderOffsetLeft: '3em',
      stickyHeaderOffsetRight: '3em',
    });`

Header is no longer Sticky : 

image

On inspect top inline style is no longer applied  

image

Unfortunately there is no on-line example at the moment, Just trying on local 

I have a quick fix by adding extra CSS
 
.sticky-header-container.fix-sticky.fixed-table-container {top: 60px;}

### LINKED ISSUE #4814: StickyHeader not working with Bootstrap container or fullscreen or offset-y
I can't get StickyHeader to work with a Bootstrap container:

1. The sticky header "jumps" out of the container, as soon as it is supposed to be at the top.
2. As well data-sticky-header-offset-y seems to have no effect: whatever value is entered, the offset is very large and always the same.
3. If fullscreen is used, the sticky header is not working at all.

Live code: [live snippet](https://live.bootstrap-table.com/code/aflogit/1612)

Replication
1. Scroll down: header "jumps" out of container
2. Change data-sticky-header-offset-y variable in HTML: no effect
3. Click on fullscreen and scroll down: header is not fixed

wenzhixin:
**Bug fix?**
yes

**New Feature?**
no

**Resolve an issue?**
Fix #4755
Fix #4814
#5532

**Example(s)?**
Before: https://live.bootstrap-table.com/code/dreamerworks/1701
After: https://live.bootstrap-table.com/code/wenzhixin/6645

### DIFF
--- a/site/docs/extensions/sticky-header.md
+++ b/site/docs/extensions/sticky-header.md
@@ -37,11 +37,11 @@ This is an extension which provides a sticky header for the table when scrolling
 
 - **attribute:** `data-sticky-header-offset-left`
 
-- **type:** `Number | String`
+- **type:** `Number`
 
 - **Detail:**
 
-   Set the left offset of the sticky header container.
+   Set the left offset of the sticky header container. If the body padding left is `60px`, this value would be `60`.
 
 - **Default:** `0`
 
@@ -49,11 +49,11 @@ This is an extension which provides a sticky header for the table when scrolling
 
 - **attribute:** `data-sticky-header-offset-right`
 
-- **type:** `Number | String`
+- **type:** `Number`
 
 - **Detail:**
 
-   Set the right offset of the sticky header container.
+   Set the right offset of the sticky header container. If the body padding right is `60px`, this value would be `60`.
 
 - **Default:** `0`
 
@@ -65,6 +65,6 @@ This is an extension which provides a sticky header for the table when scrolling
 
 - **Detail:**
 
-   Set the Y offset from the top of the window to pin the sticky header. If there is a fixed navigation bar with a height of 60px, this value would be `60`.
+   Set the Y offset from the top of the window to pin the sticky header. If there is a fixed navigation bar with a height of `60px`, this value would be `60`.
 
 - **Default:** `0`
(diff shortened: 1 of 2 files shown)


UtechtDustin:
I'm not sure if it makes sense to define px hardcoded. Maybe we should define a option for that, similar to [widthUnit](https://bootstrap-table.com/docs/api/column-options/#widthunit).

wenzhixin:
https://github.com/wenzhixin/bootstrap-table/issues/4755#issuecomment-666310189

> The .offset() and the .height() function returning px, so in my opinion stickyHeaderOffsetY should always be px or you have to calc how much px the % or em is you need for the const end.

I think the unit always be `px` is ok,  we can calc how much `px` the `%` or `em` something like this: https://github.com/wenzhixin/bootstrap-table-examples/pull/425

djhvscf:
Just thinking out loud.. Could we add an option where if the `stickyHeaderOffsetLeft` is not set, then the extension assumes that it needs to calculate automagically?

djhvscf (approved):

wenzhixin:
There may be different uses of cases, making it difficult to implement calculating automagically.

djhvscf:
For me this is ok, @UtechtDustin what do you think?

UtechtDustin (approved):

UtechtDustin:
It's fine for me.

PR #5617, opened 2021-03-06
Fix #5614 - Parameters with accents

### LINKED ISSUE #5614: Search parameters with accent
from CDN

**Description**  
Returned string from queryParams function contains an accent like : `?first_name.startswith=José`
Get request is not encoded correctly `?first_name.startswith=Jos%C3%A9` and then no match is found.
Html file is recorded in utf-8. Meta tag utf-8 is set.
Same issue using or not `data-search-accent-neutralise`.

**Possible  (optional)**  
Is is possible the form is not utf-8 encoded when submitting?

Mamisoa

ucusita:
If you use php language, the solution may be:
function myUrlEncode($string) {
    $entities = array('%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%2B', '%24', '%2C', '%2F', '%3F', '%25', '%23', '%5B', '%5D');
    $replacements = array('!', '*', "'", "(", ")", ";", ":", "@", "&", "=", "+", "$", ",", "/", "?", "%", "#", "[", "]");
    return str_replace($entities, $replacements, urlencode($string));
}

In the javascript language:
var uri = "my test.asp?name=ståle&car=saab";
var res = encodeURI(uri);

### DIFF
--- a/site/_data/nav.yml
+++ b/site/_data/nav.yml
@@ -21,6 +21,7 @@
     - title: Auto Refresh
     - title: Cookie
     - title: Copy Rows
+    - title: Custom View
     - title: Defer URL
     - title: Editable
     - title: Export


UtechtDustin:
Why do you opened a merge request for that ?!

PR #5619, opened 2021-03-07
translating locales for pt-PT

davidsbatista:
adding the local in pt-PT

### DIFF
--- a/src/locale/bootstrap-table-pt-PT.js
+++ b/src/locale/bootstrap-table-pt-PT.js
@@ -5,7 +5,7 @@
 
 $.fn.bootstrapTable.locales['pt-PT'] = $.fn.bootstrapTable.locales['pt'] = {
   formatCopyRows () {
-    return 'Copy Rows'
+    return 'Copiar Linhas'
   },
   formatPrint () {
     return 'Print'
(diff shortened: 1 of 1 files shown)


src/locale/bootstrap-table-pt-PT.js
```diff
@@ -24,19 +24,19 @@ $.fn.bootstrapTable.locales['pt-PT'] = $.fn.bootstrapTable.locales['pt'] = {
     return `A mostrar ${pageFrom} at&eacute; ${pageTo} de ${totalRows} linhas`
   },
   formatSRPaginationPreText () {
-    return 'previous page'
+    return 'p&aacute;gina anterior'
```
UtechtDustin, line 27: Why the html entites ?
`Á` is part of utf8 so i guess we dont need the html entities.

UtechtDustin (changes requested):

src/locale/bootstrap-table-pt-PT.js
davidsbatista: saw a mix of html entities and utf8, and decided to stick with just one; do you prefer all in UTF8 instead?

UtechtDustin: I dont see a reason to use html entites, without the code is more readable :-)

davidsbatista: updated! thanks for the quick feedback :)

UtechtDustin (approved):

UtechtDustin:
Thanks for your contribution 👍🏻

PR #5625, opened 2021-03-09
Update API Docs: point out suppression of footerField explicitly

cnbeining:
**Bug fix?**

No

**New Feature?**

No

**Resolve an issue?**

No

It seems that documentation for [footerfield](https://bootstrap-table.com/docs/api/table-options/#footerfield) is rather ambiguous: clarification of when `footerField` shall NOT work should be great.

BTW the [example](https://examples.bootstrap-table.com/#options/footer-field.html#view-source) for `footerfield` is not working: a separate PR shall be raised.

Thanks,


cnbeining:
See: https://github.com/wenzhixin/bootstrap-table/pull/5625

djhvscf (approved):

Issue #5653, opened 2021-03-24
Optional compile js for development


ko22009:
Add feature to config optional compiling scr js files.
Manually I commented and rename path to specific file.
It will be useful pass by arguments

UtechtDustin:
I'm not sure what do you mean with this issue, could you explain it further ?

ko22009:
For example, in development I don't want compile everytime vue js wrapper and another files (locales, extensions).
I changed one file and want compile only one.

ko22009:
#5655

UtechtDustin:
But you can use the src's files for the development enviroment, we only have to compile if we release a new version.
So i'm not sure if we need this feature.

ko22009:
@djhvscf, How I can compile with watching specific file? How I can create example file? But I have not looked for example folder.

wenzhixin:
Added readme: https://github.com/wenzhixin/bootstrap-table/pull/5674/commits/7533354af13423e843ff9593700a59976138a214

PR #5655, opened 2021-03-24
Improve rollup building

ko22009:


ko22009:
#5655

UtechtDustin:
https://github.com/wenzhixin/bootstrap-table/issues/5653#issuecomment-806227532

@wenzhixin @djhvscf what do you think ?

djhvscf:
What I do usually to test locally is run the rollup script adding the --watch parameter. After that I add the reference to the file generated in the local dist folder. This is because I want the environment as when the version is released.

ko22009:
> What I do usually to test locally is run the rollup script adding the --watch parameter. After that I add the reference to the file generated in the local dist folder. This is because I want the environment as when the version is released.

good solution.

UtechtDustin:
I don't see a problem to use the source files for the development, as we only compile it to ECMAScript X, minify the plugin etc.
So @wenzhixin would see issues while realeasing a new version.

djhvscf:
For me this looks good.. Since when we try to compile them it takes a lot of time compiling the locales and the files

ko22009:
@djhvscf, give you example with watch arguments?

PR #5657, opened 2021-03-26
Fix #5637 one colspan in header

### LINKED ISSUE #5637: Toggle Columns in grouped headers
**Description**  
When a grouped header (aka nested) has 1 nested level (like "Price" and "Total" in the example) there is a strange behaviour: I would expect a way to hide the whole column, yet when I hide "Total", "Price" remains visible. Is there a way to hide that?
You can see it works fine in case the 2nd level header has more than 1 element: I hide `Id` and `Name` and `Key` is automatically hidden.

**Expected behaviour**  
"Price" header being hidden when I hide "Total".

**Actual behaviour**  
"Price" stays visible.

**Example**  
https://live.bootstrap-table.com/code/antonioaltamura/6940

ko22009:
https://live.bootstrap-table.com/code/ko22009/7031

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -60,28 +60,32 @@ export default {
         flag[i][j] = false
       }
     }
-
-    for (let i = 0; i < columns.length; i++) {
-      for (const r of columns[i]) {
+    for (let i = columns.length - 1; i >= 0; i--) {
+      for (let z = columns[i].length - 1; z >= 0; z--) {
+        const r = columns[i][z]
         const rowspan = r.rowspan || 1
         const colspan = r.colspan || 1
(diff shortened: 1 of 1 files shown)


ko22009:
#5637

ko22009:
Found problem with that template https://live.bootstrap-table.com/code/ko22009/7039

djhvscf:
By that, you mean we can close this PR?

ko22009:
@djhvscf, yes, I found place of mess, but I cannot create correct algorithm.

djhvscf:
If you know where the fix could apply, let's take more time to test and then open a PR. Thanks @ko22009

PR #5726, opened 2021-05-08
Merging master into develop. Just to keep develop updated

djhvscf:
**Bug fix?**

**New Feature?**

**Resolve an issue?**

**Example(s)?**


wenzhixin:
I don't think we should need this PR. I'll release a new version in these two days.

PR #5817, opened 2021-08-02
Fix/5813

### LINKED ISSUE #5813: Column FilterData function example
I'm using the data-filter-data attribute to customize the select filter values of a column through a function but the select values don't work.

I would like an example to use the function on this attribute.

**data-filter-data=func:functionName**

UtechtDustin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5813

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

Reacreation of the filter control elements, if data was loaded using the `load` method.

**💡Example(s)?**

Before: https://live.bootstrap-table.com/code/UtechtDustin/8508
After: https://live.bootstrap-table.com/code/UtechtDustin/8509

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -154,6 +154,12 @@ $.BootstrapTable = class extends $.BootstrapTable {
     super.init()
   }
 
+  load (data) {
+    super.load(data)
+
+    UtilsFilterControl.createControls(this, UtilsFilterControl.getControlContainer(this))
+  }
+
   initHeader () {
     super.initHeader()
 
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -157,6 +157,10 @@ $.BootstrapTable = class extends $.BootstrapTable {
   load (data) {
     super.load(data)
 
+    if (!this.options.filterControl) {
+      return
+    }
+
     UtilsFilterControl.createControls(this, UtilsFilterControl.getControlContainer(this))
   }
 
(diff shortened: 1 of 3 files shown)


src/extensions/filter-control/utils.js
```diff
@@ -353,6 +353,8 @@ export function createControls (that, header) {
     $.each(header.find('th'), (i, th) => {
       const $th = $(th)
 
+      $th.find('.filter-control').remove()
```
djhvscf, line 356: Is this needed?

djhvscf (changes requested):

src/extensions/filter-control/utils.js
UtechtDustin: Yes, without that we would append the input/selects again, again and again.
Which means we would have X Input/Elements per Header.

djhvscf:
Thanks @UtechtDustin

src/extensions/filter-control/bootstrap-table-filter-control.js
```diff
@@ -154,6 +154,12 @@ $.BootstrapTable = class extends $.BootstrapTable {
     super.init()
   }
 
+  load (data) {
+    super.load(data)
+
+    UtilsFilterControl.createControls(this, UtilsFilterControl.getControlContainer(this))
```
djhvscf, line 160: Should we add an if statement?

```js
if (!this.options.filterControl) {
  return;
}
```

djhvscf (approved):

src/extensions/filter-control/bootstrap-table-filter-control.js
UtechtDustin: yes we should, thanks for the hint.
Added in the last commit.

djhvscf (approved):

PR #5818, opened 2021-08-03
Feature/5815

### LINKED ISSUE #5815: Copy Rows without html code
The Copy-Rows extension brings all the value of the cell even if it is an html code, for example when I create a column for the operations using `OperateFormatter`, the html code is copied to the clipboard. 
**Here is an example**
[Example1](https://live.bootstrap-table.com/code/levyaugustoriosaude/8460)

There had to be an option that excludes a column from being copied. Looking at other extensions like **export** or **print,** it has the functions `data-print-ignore="true"` or `data-force-hide="true"`. Is there a similar option for copy-rows?
Could you have a new function `copy-ignore = “false”` as default value, to make this function?

In my attempts to avoid this I realized that using colspan and rowspan, copy rows only takes the column that is doing the rowspan, however it only behaves like this if it has at least two columns inside the colspan, as in the example below:
[Example2](https://live.bootstrap-table.com/code/levyaugustoriosaude/8459)

And in case there is a `<a href=” ”>` tag in the cell and it is not a column of operations, I could use the `text()` functionality, so as not to copy the html code and just get the text

UtechtDustin:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5815

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

Added a new column option `ignoreCopy` to prevent copying the column data.
Added a new column option `rawCopy` to copy the raw value instead of the formatted one.

**💡Example(s)?**

The column `Item Name` will be ignored while copying.
Ignore Column: https://live.bootstrap-table.com/code/UtechtDustin/8512

The raw value of the column `Item Price` will be copied.
Raw copy: https://live.bootstrap-table.com/code/UtechtDustin/8538

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/extensions/copy-rows.md
+++ b/site/docs/extensions/copy-rows.md
@@ -82,6 +82,19 @@ This extension adds functionality for copying selected rows to the clipboard. Cu
 
 - **Default:** `false`
 
+### rawCopy
+
+- **Attribute:** `data-raw-copy`
+
+- **type:** `Boolean`
+
+- **Detail:**
+
+  Set `true` to copy the raw value instead the formatted one.   
+  If no formatter is used, this option has no effect.
+
+- **Default:** `false`
+
 ## Icons
 
 - copy: 'fa-copy'
--- a/site/docs/extensions/copy-rows.md
+++ b/site/docs/extensions/copy-rows.md
@@ -68,7 +68,21 @@ This extension adds functionality for copying selected rows to the clipboard. Cu
 
 - **Default:** `false`
 
-### Icons
+## Column options
+
+### ignoreCopy
+
+- **Attribute:** `data-ignore-copy`
+
+- **type:** `Boolean`
+
+- **Detail:**
+
+  Set `true` to ignore this column while copying.
+
+- **Default:** `false`
+
+## Icons
 
 - copy: 'fa-copy'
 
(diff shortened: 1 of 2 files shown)


djhvscf (changes requested):
I have read the root issue and what I understood is that the extension is copying the html instead of the content.. Should we just use a regex to remove the html characters?

UtechtDustin:
@djhvscf 
> There had to be an option that excludes a column from being copied. Looking at other extensions like export or print, it has the functions data-print-ignore="true" or data-force-hide="true". Is there a similar option for copy-rows?
Could you have a new function copy-ignore = “false” as default value, to make this function?

UtechtDustin:
I could add a option to copy the raw value of the column, i don't think we should remove the html later.

djhvscf:
> I could add a option to copy the raw value of the column instead of the formatted one, i don't think we should remove the html later.

ok ok.. if you agree let's add the option

UtechtDustin:
Added a new column option to copy the unformatted value :-)

djhvscf (approved):

PR #5819, opened 2021-08-03
fix/5812

### LINKED ISSUE #5812: When using data-search-selector it doesn't detect search-text
All? tried on 5 and on the live editor.

**Description**  
I am setting up a data-search-selector to use sticky search and search-text just change the value of that input, but doesn't execute the search.

**Example**  
https://live.bootstrap-table.com/code/eliezra236/8439

**Possible  (optional)**  

I made some attempts of changing it way later (trough button that execute:   
```
$("#table").bootstrapTable("refreshOptions", {
    searchText: "testing",
  });
```
And it does the same thing, changes the text in the input but doesn't execute the search.

UtechtDustin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5812
**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Fixed the comparision of search inputs (DOM element and jQuery Element/Object).

**💡Example(s)?**

Before: https://live.bootstrap-table.com/code/eliezra236/8439
After: https://live.bootstrap-table.com/code/UtechtDustin/8513

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -914,12 +914,9 @@ class BootstrapTable {
       }
 
       const $searchInput = Utils.getSearchInput(this)
+      const $currentTarget = currentTarget instanceof jQuery ? currentTarget : $(currentTarget)
 
-      if (
-        (currentTarget instanceof jQuery && currentTarget.is($searchInput)) ||
-        (!(currentTarget instanceof jQuery) && currentTarget === $searchInput[0]) ||
-        $(currentTarget).hasClass('search-input')
-      ) {
+      if ($currentTarget.is($searchInput) || $currentTarget.hasClass('search-input')) {
         this.searchText = text
         this.options.searchText = text
       }
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -913,7 +913,11 @@ class BootstrapTable {
         return
       }
 
-      if (currentTarget === Utils.getSearchInput(this)[0] || $(currentTarget).hasClass('search-input')) {
+      if (
+        (currentTarget instanceof jQuery && currentTarget.is(Utils.getSearchInput(this))) ||
+        (!(currentTarget instanceof jQuery) && currentTarget === Utils.getSearchInput(this)[0]) ||
+        $(currentTarget).hasClass('search-input')
+      ) {
         this.searchText = text
         this.options.searchText = text
       }
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -913,9 +913,11 @@ class BootstrapTable {
         return
       }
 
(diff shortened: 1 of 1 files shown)


src/bootstrap-table.js
```diff
@@ -913,7 +913,11 @@ class BootstrapTable {
         return
       }
 
-      if (currentTarget === Utils.getSearchInput(this)[0] || $(currentTarget).hasClass('search-input')) {
+      if (
+        (currentTarget instanceof jQuery && currentTarget.is(Utils.getSearchInput(this))) ||
```
djhvscf, line 917: What about to get the `Utils.getSearchInput(this)` in a variable and then just evaluate it in this if statement?

src/bootstrap-table.js
```diff
@@ -913,7 +913,11 @@ class BootstrapTable {
...
+        (!(currentTarget instanceof jQuery) && currentTarget === Utils.getSearchInput(this)[0]) ||
+        $(currentTarget).hasClass('search-input')
```
djhvscf, line 919: At the end we could convert currentTarget to a jQuery object and then evaluate if that object has the corresponding class.. I don't think we could have performance hit if we create a jQuery object from a jQuery object... Do you know that?

djhvscf (approved):

djhvscf (changes requested):
please review my previous comments

src/bootstrap-table.js
UtechtDustin: I'm not sure what do you mean, we don't create a jQuery object on that if.
We just check if the currentTarget is a jQuery object, if it is a jQuery we compare it using the `is` function.
If it is a dom element we get the dom element from the jquery object and just compare it with `===`

src/bootstrap-table.js
UtechtDustin: I agree it's a micro optimization :-)

src/bootstrap-table.js
djhvscf: What about this?
```js
const searchInput = Utils.getSearchInput(this)
currentTarget = $(currentTarget)
if (
        currentTarget.is(searchInput) ||
        currentTarget.hasClass('search-input')
...
```

UtechtDustin: I guess creating a new jquery object will cost more performance.
So i could invert it, i can check if it is a jQuery Object and set `currentTarget` to a domElement.
In the if i would use the vanilla js functions for the comparison.

What do you mean ? :-)

UtechtDustin:
@djhvscf please check again.

djhvscf (approved):

Issue #5781, opened 2021-07-05
Invert saved columns in cookie extentions


Naruto-kyun:
**Description**  
Today, when you use the cookie plugin to save current visible columns, you should ask all users to check the visibility of new columns inside an already saved table.
The idea is to save hidden columns and not visible columns. So in case of new column, it will be visible by default.

I can do the PR quickly, but before i just want to know if the idea have a sense to you.

djhvscf:
Hi @Naruto-kyun that does make sense. Please open the PR and let's review it.

PR #5825, opened 2021-08-09
Cookie fix new columns

Naruto-kyun:
**🤔Type of Request**
- [x] **Improvement**

See issue: https://github.com/wenzhixin/bootstrap-table/issues/5781

**📝Changelog**

- [x] **Extensions**

When the table evolved and new columns are added, they will not be hidden by default anymore.

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/Naruto-kyun/8580
After: https://live.bootstrap-table.com/code/Naruto-kyun/8582

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/extensions/cookie.md
+++ b/site/docs/extensions/cookie.md
@@ -178,9 +178,9 @@ toc: true
 
 - **Detail:**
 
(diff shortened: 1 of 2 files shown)


djhvscf:

PR #5918, opened 2021-10-24
fix/5917

### LINKED ISSUE #5917: updateRow does not work as expected after sorting
1.18.3

**Description**  
When sorting a table and using updateRow the wrong index is changed

**Example**  
https://live.bootstrap-table.com/code/Friends4U/9365
https://live.bootstrap-table.com/code/Friends4U/9367 (with cookie after sorting once)

UtechtDustin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5917 

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Fixed an issue with with the row index, if table was sorted and the method `updateRow will be used.

**💡Example(s)?**

Sort by name and click multiple times the action in the first row.

Before: https://live.bootstrap-table.com/code/Friends4U/9365
After: https://live.bootstrap-table.com/code/UtechtDustin/9368

**☑️Self Check before Merge**


djhvscf:
> @wenzhixin it also seems there is still a bug with the test's (or in that case with one of the tests)

it's weird because if we rerun it, it works

src/bootstrap-table.js
```diff
@@ -2532,9 +2532,15 @@ class BootstrapTable {
...
 
       if (params.hasOwnProperty('replace') && params.replace) {
-        this.options.data[params.index] = params.row
+        const oldRow = this.data[params.index]
+
+        Object.keys(oldRow).forEach(function (key) {
+          oldRow[key] = null
+        })
+
+        $.extend(this.data[params.index], $.extend(oldRow, params.row))
```
djhvscf, line 2541: What happens if you do:

```js
this.data[params.index] = params.row
```

I think the logic is too complicated

djhvscf (changes requested):

src/bootstrap-table.js
UtechtDustin: That was my first commit for the replacing part.
But as i already wrote in the post above, if i reassigne the value (`= params.row`) it don't work.
If i use the too complicated ugly fix, it works fine locally but not on the editor.

I'm quite confused why it only works with `$.extend`  and why the complicated fix don't work on the editor.

PR #5941, opened 2021-11-11
Fixed CVE security problem

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**📝Changelog**
Fixed CVE security problem

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/wenzhixin/9594
After: https://live.bootstrap-table.com/code/wenzhixin/9595

CVE problem: https://nvd.nist.gov/vuln/detail/CVE-2021-23472

> This affects all versions of package bootstrap-table. A type confusion vulnerability can lead to a bypass of input sanitization when the input provided to the `escapeHTML` function is an array (instead of a string) even if the escape attribute is set.

Ref: 
* https://github.com/component/escape-html/blob/master/index.js
* https://stackoverflow.com/questions/40263803/native-javascript-or-es6-way-to-encode-and-decode-html-entities

**☑️Self Check before Merge**

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -206,29 +206,27 @@ export default {
   },
 
   escapeHTML (text) {
-    if (typeof text === 'string') {
+    if (!text) {
       return text
-        .replace(/&/g, '&amp;')
-        .replace(/</g, '&lt;')
-        .replace(/>/g, '&gt;')
-        .replace(/"/g, '&quot;')
-        .replace(/'/g, '&#039;')
-        .replace(/`/g, '&#x60;')
     }
-    return text
+    return text.toString()
+      .replace(/&/g, '&amp;')
+      .replace(/</g, '&lt;')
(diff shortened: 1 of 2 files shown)


src/utils/index.js
```diff
@@ -206,29 +206,27 @@ export default {
   },
 
   escapeHTML (text) {
-    if (typeof text === 'string') {
+    if (!text) {
```
djhvscf, line 209: In which case this can't be a string?

djhvscf (changes requested):

src/utils/index.js
djhvscf: Nevermind.. I just saw it!

djhvscf (approved):

Issue #5898, opened 2021-10-18
autoRefresh


Scnck:
it seems you need to have "autoRefresh: true" for auto refresh to work but according to the documentation it's not required?

UtechtDustin:
Because of our [buttons](https://bootstrap-table.com/docs/api/table-options/#buttons) feature (which also will be used internally), this is already implemented.
You can disable the button using `data-show-auto-refresh="false"`.
Example: https://live.bootstrap-table.com/code/UtechtDustin/9602

PR #5946, opened 2021-11-11
Improved the documentation for the auto refresh extension

UtechtDustin:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [x] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
ref #5898

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/auto-refresh/bootstrap-table-auto-refresh.js
+++ b/src/extensions/auto-refresh/bootstrap-table-auto-refresh.js
@@ -8,6 +8,7 @@ const Utils = $.fn.bootstrapTable.utils
 
 $.extend($.fn.bootstrapTable.defaults, {
   autoRefresh: false,
+  showAutoRefresh: true,
   autoRefreshInterval: 60,
   autoRefreshSilent: true,
   autoRefreshStatus: true,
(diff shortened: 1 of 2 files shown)


src/extensions/auto-refresh/bootstrap-table-auto-refresh.js
```diff
@@ -8,6 +8,7 @@ const Utils = $.fn.bootstrapTable.utils
 
 $.extend($.fn.bootstrapTable.defaults, {
   autoRefresh: false,
+  showAutoRefresh: true,
```
djhvscf, line 11: We need more logic for this option, right?

djhvscf (changes requested):

src/extensions/auto-refresh/bootstrap-table-auto-refresh.js
UtechtDustin: No it's already implemented because of our [buttons](https://bootstrap-table.com/docs/api/table-options/#buttons) feature.
At all we don't really need that assignment there (because the default value of `show<name>` is always true), but i guess it's better for our understanding if we read the source code :-)

wenzhixin (approved):

wenzhixin (approved):

PR #5583, opened 2021-02-08
Improving filter control

djhvscf:
Improving filter control extension

Default options:

- [x] filterControl
- [x] filterControlVisible
- [x] onColumnSearch
- [x] onCreatedControls
- [x] alignmentSelectControlOptions:
- [x] filterTemplate
- [x] disableControlWhenSearch (removed)
- [x] searchOnEnterKey
- [x] showFilterControlSwitch
- [x] filterControlContainer

Column options:
- [x] filterControl // input
- [x] filterControl //select
- [x] filterControl // datepicker
- [x] filterDataCollector
- [x] filterData
- [x] filterDatepickerOptions
- [x] filterStrictSearch
- [x] filterStartsWithSearch
- [x] filterControlPlaceholder
- [x] filterDefault
- [x] filterOrderBy // asc
- [x] filterOrderBy // desc

Scenarios
- [x]  When height is not set //input
- [x]  When height is set // input
- [x]  When height is not set // select
- [x]  When height is set // select
- [x]  When height is not set // datetimepicker
- [x]  When height is set // datetimepicker
- [x]  Show Columns when height is not set
- [x]  Show Columns when height is set

Integrations
- [x]  Cookie integration
- [x] ~~MultipleSelect with basic configuration~~
- [ ] ~~MultipleSelect when height is set (Facing css issue. See https://github.com/wenzhixin/bootstrap-table/pull/5583#issuecomment-821907389)~~
- [x] ~~MultipleSelect when height is not set~~
- [ ] ~~MultipleSelect when multiple is set (Facing css issue. See https://github.com/wenzhixin/bootstrap-table/pull/5583#issuecomment-821907389)~~

Update docs
- [x]  Add/Remove documentation

Fix https://github.com/wenzhixin/bootstrap-table/issues/5541 (https://live.bootstrap-table.com/code/djhvscf/9665)
Fix https://github.com/wenzhixin/bootstrap-table/issues/5588 (https://live.bootstrap-table.com/code/djhvscf/9666)
Fix https://github.com/wenzhixin/bootstrap-table/issues/4998 (https://live.bootstrap-table.com/code/djhvscf/9667)
Fix https://github.com/wenzhixin/bootstrap-table/issues/5468 (https://live.bootstrap-table.com/code/djhvscf/9690)
Fix #5684 (https://live.bootstrap-table.com/code/djhvscf/9691)
[...]

PR #5959, opened 2021-11-21
fix/5828

### LINKED ISSUE #5828: The filter-control input is clearing while searching
The filter-control input is clearing while searching and the filter cannot be cleared.

Tried to change table settings but it persists.

screencast

An example of my table configuration:

https://live.bootstrap-table.com/code/matheusilario/8621

This example works perfectly. But in the image you can see that in the application no.

whizzrd:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5828 

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

filter controls no longer clear on server search

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/filter-control/utils.js
+++ b/src/extensions/filter-control/utils.js
@@ -329,6 +329,11 @@ export function createControls (that, header) {
       addedFilterControl = true
 
       if (column.searchable && that.options.filterTemplate[nameControl]) {
+        let value = column.filterDefault
+
+        if (that.filterColumnsPartial && that.filterColumnsPartial[column.field]) {
+          value = that.filterColumnsPartial[column.field]
+        }
         html.push(
           that.options.filterTemplate[nameControl](
             that,
@@ -336,7 +341,7 @@ export function createControls (that, header) {
             column.filterControlPlaceholder ?
               column.filterControlPlaceholder :
               '',
-            column.filterDefault
+            value
           )
         )
       }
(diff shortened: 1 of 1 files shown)


yeezy69:
With the changes from 5abee4a the search text remains in the input fields, but the select fields are now empty.

whizzrd:
You are right. I think the problem lies in filtercontrol/utils.js @ syncControls

https://github.com/wenzhixin/bootstrap-table/blob/b42124f671e2af685ac440c8d68a1522e1696c80/src/extensions/filter-control/utils.js#L514 <https://github.com/wenzhixin/bootstrap-table/blob/b42124f671e2af685ac440c8d68a1522e1696c80/src/extensions/filter-control/utils.js#L514>

where it bails when height is not set, however it seems that $header is empty so it doesn't find any controls to sync

whizzrd:
> Change [5abee4a](https://github.com/wenzhixin/bootstrap-table/commit/5abee4a5558f66f40c3c121d29c17f723524271d) works for input fields, but the select fields remain empty :(

this should fix that (in my case when using data-filter-data="url:..." on the columns with data-filter-control="select"

yeezy69:
I tried something like you in a4d2918 too, but it didn't work with the cookie extension :(

whizzrd:
> I tried something like you in [a4d2918](https://github.com/wenzhixin/bootstrap-table/commit/a4d2918a324d0ea1484ff14624bc7ca03f7cc662) too, but it didn't work with the cookie extension :(
> (in my case i using data-filter-data="url:...")

That may be a separate issue: for me when enabling the cookie plugin it does initialise the filter controls with the values from "filterControl" cookie. Could you file a separate issue with repro for that please?

djhvscf:
The issue #5828 is fixed in https://github.com/wenzhixin/bootstrap-table/pull/5583

whizzrd:
> The issue #5828 is fixed in #5583

Thanks, that looks like a really nice improvement to filter-control

PR #6045, opened 2022-02-05
Fixed an issue with custom search filter if the value is an object

### LINKED ISSUE #6044: Filter custom search doesn't work if field is an object

1.19.1

### Description

When using a custom search in filter control, if the field is an object (instead of a string or number), the custom search will pass along the string "[object Object]" to the custom search instead of the object itself. This used to work in 1.18.3, it is not working in 1.19.1.

### Example(s)

See https://live.bootstrap-table.com/code/jonatanschroeder/10437, if the tag is filtered by any value no rows are selected. If you check the console you notice that the search function receives "[object Object]" instead of the tag object itself.

### Possible Solutions

_No response_

### Additional Context

_No response_

UtechtDustin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6044 

**📝Changelog**

- [x] **Core**
- [x] **Extensions**

Fixed an issue with custom search filter if the value is an object (none array)

**💡Example(s)?**
Filter by tag to test the bug.
Before: https://live.bootstrap-table.com/code/jonatanschroeder/10437
After: https://live.bootstrap-table.com/code/UtechtDustin/10438

**☑️Self Check before Merge**


UtechtDustin:
@wenzhixin the editor is broken i can't use a specific branch for the after example.
The branch always switch back to `develop`.

wenzhixin (approved):

PR #6052, opened 2022-02-12
Fixed error when the column events are undefined

### LINKED ISSUE #6048: Missing eventhandler (data-events) reported as Uncaught TypeError: Invalid attempt to spread non-iterable instance

1.19.1

### Description

it is probably related to what has been reported in https://github.com/wenzhixin/bootstrap-table/issues/5212

we were moving from an older bootstrap-table version to 1.19.1
since then we had several issues with the error 
Uncaught TypeError: Invalid attempt to spread non-iterable instance

the eventhandler has been declaratively defined in the HTML template

```
<table id="tableKonvertierauftraege"
...
    <th data-field="aktion" data-formatter="konvauftraege_formatter_edit"
       data-events="konvauftraege_events_edit"
```

through the data-events attribute

If the implementation of the handler konvauftraege_events_edit is missing, we get then TypeError when the template is loaded and the table initialized

I would agree that an event handler that is declared for a field should also be implemented.
what is troublesome here is that the error message is not helpful to find the cause of the problem easily.

### Example(s)

_No response_

### Possible Solutions

Provide better error message to find and fix problem in customer application faster.

### Additional Context

_No response_

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6048

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Fixed error when the column events are undefined.

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/wenzhixin/10508
After: https://live.bootstrap-table.com/code/wenzhixin/10509

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1852,6 +1852,10 @@ class BootstrapTable {
         events = Utils.calculateObjectValue(null, events)
       }
 
+      if (!events) {
+        throw new Error(`Unknown event in the scope: ${_events}`)
+      }
+
       const field = this.header.fields[i]
       let fieldIndex = this.getVisibleFields().indexOf(field)
 
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -280,6 +280,7 @@ export default {
     if (
       !func &&
       typeof name === 'string' &&
+      args &&
       this.sprintf(name, ...args)
     ) {
       return this.sprintf(name, ...args)


wenzhixin:
Updated @UtechtDustin

UtechtDustin:
Don't work for me, or is the editor buggy again ?

wenzhixin:
> Don't work for me, or is the editor buggy again?

I think it is a cache problem. You can open the Network panel and check the `Disable cache` option.
image

UtechtDustin (approved):

Issue #6007, opened 2021-12-25
Problem with cookie and columns


Friends4U:
### Bootstraptable version(s) affected

1.19.1

### Description

When you select columns that were loaded with data-visible="false", they will not be saved in the cookie if you select them.

### Example(s)

https://live.bootstrap-table.com/code/Friends4U/10051

### Additional Context

P.S. The cardview is also not restored from cookie if mobileResponsive: true, see this topic: https://github.com/wenzhixin/bootstrap-table/issues/5864

Friends4U:
By the way, in 1.18.3 this did work :(

Friends4U:
And with the older version is you DO select more columns than default, the actions on cardview do not work anymore after a page refresh.

PR #6064, opened 2022-03-01
Fixes #5993 #6007

### LINKED ISSUE #5993: Cookie Extension doesn't save state of a at start invisible column incorrect

1.19.1

### Description

When having an column, which is at start invisible (data-visible="false") the state isn't saved correctly. when loading the page again with cookie extension activated all settings are loaded correctly (sorting, etc) but the column is again invisible but before reload it was made visible.
The problem accours with 1.19.1, with 1.18.3 there this bug doesn't exist.
Maybe it has to do something with this commit: https://github.com/wenzhixin/bootstrap-table/commit/f3242d1e39e6b2255ba68d85d23e5d1290fa3523

### Example(s)

https://live.bootstrap-table.com/code/simonsolutions/9948

### Possible Solutions

_No response_

### Additional Context

_No response_

inietov:
**🤔Type of Request**
- [x] **Bug fix**

**🔗Resolves an issue?**
Fix #5993 #6007
**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

When using the cookie extension, if a user starts a column as non visible but then in the interface toggles the check to show that column, when refreshing the page or if they exit the page and then got back, the columns marked as non visible still doesn't appear.

I think this is an error, because the condition always check if the column is marked as visible, ignoring the columns that the cookie contains, and I believe the cookie must take precedence, since it's more probable that the user wants to view the columns that they already checked as visible.

Special thanks to @simonsolutions to pointing in the right direction!!. 

**💡Example(s)?**

https://live.bootstrap-table.com/code/inietov/10650
**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/cookie/bootstrap-table-cookie.js
+++ b/src/extensions/cookie/bootstrap-table-cookie.js
@@ -497,7 +497,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
           return columnField === column.field
         })
 
-        column.visible = (filteredColumns.length > 0 || !column.switchable) && column.visible
+        column.visible = (filteredColumns.length > 0 || !column.switchable)
       }
     }
   }


simonsolutions (approved):

djhvscf:
Please provide an [Online Example](https://live.bootstrap-table.com/) to show your problem, thanks!

inietov:
> Please provide an [Online Example](https://live.bootstrap-table.com/) to show this PR working, thanks!

Sorry, I edited the PR with the example, I hope everything is good now. Let me know if something else is needed. Thanks!!

wenzhixin:
ref #6078
I will close this PR. Thanks anyway.

PR #6067, opened 2022-03-04
Fix #6066

### LINKED ISSUE #6066: Editable extension modify (escape) data stored in bootstrap table

1.19.1

### Description

The `initBody` function of editable extension call `getData({ escape: true });`. The escaping screw up arrays/objects in the data and the function also store this screwed data into the main bootstrap table object `options.data`

### Example(s)

http://jsfiddle.net/vxLk6r2s/3/

Here after clicking getData link I would expect to see: `[{ key:1, val:1 },{ key:2, val:2 }]`, not `"[object Object],[object Object]"`.

Note: you can get the expected result e.g. by removing the `editable` from the column.

### Possible Solutions

The `getData` function shall make deep copy of the data before escaping them. Will prepare PR.

### Additional Context

_No response_

kkozlik:
…e function #6066

**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #6066

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

Make deep copy of the data to not modify original object by the escape function

**💡Example(s)?**

http://jsfiddle.net/vxLk6r2s/3/

Here after clicking getData link I would expect to see: `[{ key:1, val:1 },{ key:2, val:2 }]`, not `"[object Object],[object Object]"`.

Note: you can get the expected result e.g. by removing the `editable` from the column.

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -172,7 +172,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
   }
 
   getData (params) {
-    const data = super.getData(params)
+    const data = structuredClone(super.getData(params))
 
     if (params && params.escape) {
       for (const row of data) {


djhvscf:
The example is not working.

kkozlik:
> The example is not working.

I am sorry, I had to forgot save my changes in the fiddle. I have updated the example link to the correct fiddle, that should work...

djhvscf:
The example is still not working. please use our [Online Editor](https://live.bootstrap-table.com/) to show your fix, thanks!

kkozlik:
> The example is still not working. 

What do you mean by "not working"? It works for me. 
After clicking getData link I get "[object Object],[object Object]" instead of expected: [{ key:1, val:1 },{ key:2, val:2 }].

> please use our [Online Editor](https://live.bootstrap-table.com/) to show your fix, thanks!

I was not able to make the editable extension working in this online editor. My biggest trouble with this editor is that I do not know what javascript code is already included and which I have to additionally include to make the extension working. Can you advice here?

UtechtDustin:
@kkozlik 
On the Issue template we wrote that we require a example using our Editor!
So jsfiddle are no valid examples for us, because of a few reasons..

Just select the Extensions you want and the editor will add the needed (bootstrap-table) files.
Here is a info page: https://bootstrap-table.com/docs/online-editor/
You also could check similar examples on the top.

PR #6074, opened 2022-03-10
Fixed escape column option not override table option bug.

### LINKED ISSUE #6071: column.escape does not override table.escape

1.19.1

### Description

Setting `escape:true` for the whole table and `escape:false` for a single column I would expect that data in all columns except this one will be escaped. But actually all data are escaped. Setting `escape` to `false` for single column does not disable escaping.

The opposite case: `escape: false` for whole table and `escape: true` for a single column works as expected.

### Example(s)

_No response_

### Possible Solutions

_No response_

### Additional Context

_No response_

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6071

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Fixed escape column option does not override table option bug.

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/wenzhixin/10759
After: https://live.bootstrap-table.com/code/wenzhixin/10760

**☑️Self Check before Merge**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ jobs:
     - stage: test
       language: node_js
       node_js:
-        - 12
+        - 14
       name: "Lint src and check docs"
       cache:
         npm: true
@@ -22,7 +22,7 @@ jobs:
     - stage: test
       language: node_js
       node_js:
-        - 12
+        - 14
       name: "Cypress Test"
       cache:
         npm: true
@@ -40,7 +40,7 @@ jobs:
       rvm:
         - 2.4.1
       before_install:
-        - nvm install 12
+        - nvm install 14
       script: ./deploy.sh
       if: branch = master AND type = push
       cache: bundler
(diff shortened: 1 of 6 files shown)


UtechtDustin (changes requested):
In the after example the column `Item Name` is not escaped, though the column don't have `data-escape="false"`.

djhvscf (approved):
I think it is working as expected based on the original issue

UtechtDustin (approved):
I guess i had again a cache issue with the editor, the after example works fine now.
@wenzhixin could you check/improve that ?

wenzhixin:
Yes, it's a cache problem. I have updated the code, I think we can merge this PR now.

Fanduzi:
看起来又引入了新bug. 因为当我的bootstrap-table从1.19.1升级到1.20后, 原本正常的字符乱码了. 我只设置了column escape没有设置table escape. 
不变动任何代码, 只是将bootstrap-table降级会1.19.1后, 一切又正常了
image
image

PR #6077, opened 2022-03-12
WIP: Feature/add

### LINKED ISSUE #6057: onColumnSwitch is not triggered when "Toggle All" is clicked
### Description

During my implementation of server-side pagination and sorting (for which I [cannot use the supported means of server-side pagination](https://github.com/wenzhixin/bootstrap-table-examples/issues/450#issuecomment-1031731127)), I added a function `onColumnSwitch` to set a cookie so that column switch states are remembered when paging and sorting.

However, I noted that when columns are switched using the "Toggle All" checkbox, those toggled columns do not result in a call to the `onColumnSwitch` callback function...

It would be nice if it did.  Alternatively, if you provided an event for `onToggleAll`, I could handle it without needing to use `onAll`.

### LINKED ISSUE #6058: Unable to set sortable flag via updateColumnTitle()
### Description

I would like to be able to enable the `sortable` option in a specific column _after_ I got the results (from the server).
To give a bit of context, I want to enable `sortable` only if the number of results is below a threshold.
It seems `updateColumnTitle` is not the right way, as it doesn't update the `sortable` option.
https://live.bootstrap-table.com/code/antonioaltamura/10565

I also tried `refreshOptions()`, which actually works, but it forces a refresh of the whole table, and that includes an extra remote request to fetch the data, and I would like to avoid that. As a side, if you use `refreshOptions` in `responseHandler` or `onLoadSuccess` you create an endless loop.
https://live.bootstrap-table.com/code/antonioaltamura/10567

### Example(s)

https://live.bootstrap-table.com/code/antonioaltamura/10565

https://live.bootstrap-table.com/code/antonioaltamura/10567

wenzhixin:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6058

**📝Changelog**

* Added update sortable support of updateColumnTitle method.

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/antonioaltamura/10565
After: https://live.bootstrap-table.com/code/wenzhixin/10788

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -3296,13 +3296,27 @@ class BootstrapTable {
       return
     }
 
-    this.columns[this.fieldsColumnsIndex[params.field]].title =
-      this.options.escape ? Utils.escapeHTML(params.title) : params.title
+    const column = this.columns[this.fieldsColumnsIndex[params.field]]
 
-    if (this.columns[this.fieldsColumnsIndex[params.field]].visible) {
+    column.title = this.options.escape ? Utils.escapeHTML(params.title) : params.title
+
+    if (params.hasOwnProperty('sortable')) {
+      column.sortable = params.sortable
+    }
+
+    if (column.visible) {
       this.$header.find('th[data-field]').each((i, el) => {
         if ($(el).data('field') === params.field) {
-          $($(el).find('.th-inner')[0]).text(params.title)
+          const $el = $($(el).find('.th-inner')[0]).text(params.title)
+
+          if (params.hasOwnProperty('sortable')) {
+            if (params.sortable) {
+              $el.addClass('sortable both')
+            } else {
+              $el.removeClass('sortable both asc desc')
+            }
+          }
+          $(el).data(column)
           return false
         }
       })


UtechtDustin (changes requested):
I don't think it makes sense that you can update the sortable option via method `updateColumnTitle`.
As the method name already says update column TITLE.

Then we have a method which name says update title but also could update the sortable option.

wenzhixin:
Do you think add a new method like `updateColumnSortable`?

UtechtDustin:
I guess a method `updateColumn` or `updateColumnOptions` makes more sense. so we can implement mehr stuff.
But i wonder that we don't have a method like this.

src/bootstrap-table.js
```diff
@@ -3288,13 +3289,27 @@ class BootstrapTable {
...
+          const $el = $($(el).find('.th-inner')[0]).text(params.title)
+
+          if (params.hasOwnProperty('sortable')) {
+            if (params.sortable) {
+              $el.addClass('sortable both')
+            } else {
+              $el.removeClass('sortable both asc desc')
+            }
+          }
+          $(el).data(column)
```
djhvscf, line 3312: we're instantiating a lot of jquery object with $(el). I think we could use something like const element = $(el);

djhvscf (changes requested):

UtechtDustin:
Any news here @wenzhixin  ?

wenzhixin:
> But i wonder that we don't have a method like this.

I think it's okay to set it up directly with the `refreshOptions` method.

So let's close this PR?

djhvscf:
We can either change the method name or implement a new method where we can change all the column config

UtechtDustin:
I think we should add a new method which allowes to change all column options as @djhvscf said, but i guess this has no high priority yet.
So the "workaround" using the `refreshOptions`  is ok for now.

PR #6103, opened 2022-04-17
Add BS Tooltips for buttons in toolbar

CosmoFox:
Added  'data-bs-toggle': 'tooltip' and  'data-bs-placement': 'bottom' for buttons in toolbar (fullscreen, toggle, refresh, paginationSwitch)

**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [X] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

- [X] **Core**
- [ ] **Extensions**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -597,7 +597,9 @@ class BootstrapTable {
         event: this.togglePagination,
         attributes: {
           'aria-label': opts.formatPaginationSwitch(),
-          title: opts.formatPaginationSwitch()
+          title: opts.formatPaginationSwitch(),
+          'data-bs-toggle': 'tooltip',
+          'data-bs-placement': 'bottom'
         }
       },
       refresh: {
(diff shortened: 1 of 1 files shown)


UtechtDustin:
Thank you for the PR, i have a few points:

- I guess there are more buttons as this four
- It makes sense to make the placement configurable instead of hardcoded `bottom`
- It should only be added if the used theme is a bootstrap theme
- Does all bootstrap Version have the same tooltip mechanism (data attributes) ?
- Please provide us an example using our [editor](https://live.bootstrap-table.com/).
- Please use the same syntax, `title` don't use (single) qoutes.

CosmoFox:
> * Please use the same syntax, `title` don't use (single) qoutes.
rows with **title** are original code, i added only **,** at the end of row

For other points i need more time, cos it was fastfix for my project.

UtechtDustin:
You're right its on the source code, but we would love if you could fix it :-)

UtechtDustin:
Any new here @CosmoFox ?

CosmoFox:
closing, will open new one for new version

PR #6112, opened 2022-04-25
Fixed exportOptions option cannot support data attribute

### LINKED ISSUE #6107: How to change the csv delimiter
### Description

Hello,

I tried to find a simple way to change the character used as csv delimiter but I didn't find any option to do it... it would be great to have this option: in my country (Italy) the standard is to use the semicolon (see wikipedia: Semicolons are often used instead of commas in many European in order to use the comma as the decimal separator and, possibly, the period as a decimal grouping character.)

The dirty trick to solve this problem is simple (change the character in the hard coded csvSeparator variable in tableExport.min.js) but that's not right way to do things...

Thanks

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6107 

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

Fixed the `exportOptions` option cannot support the data attribute.

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/wenzhixin/11324
After: https://live.bootstrap-table.com/code/wenzhixin/11323

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/export/bootstrap-table-export.js
+++ b/src/extensions/export/bootstrap-table-export.js
@@ -84,6 +84,10 @@ $.BootstrapTable = class extends $.BootstrapTable {
         exportTypes = types.map(t => t.slice(1, -1))
       }
 
+      if (typeof o.exportOptions === 'string') {
+        o.exportOptions = Utils.calculateObjectValue(null, o.exportOptions)
+      }
+
       this.$export = this.$toolbar.find('>.columns div.export')
       if (this.$export.length) {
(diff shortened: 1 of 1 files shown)


wenzhixin:
@UtechtDustin

UtechtDustin (approved):

PR #6145, opened 2022-05-10
Fixed XSS vulnerability bug by onCellHtmlData

### LINKED ISSUE #4005: Export Bug with data-filter-control="select"
Having Select Filtering on a Column when exporting will cause every option in the select filter to be placed in the title header for the column.

Example Can be seen Here in Date Column when exporting to MS-Excel.

https://codepen.io/jahid-webdev/pen/ObegKN

### LINKED ISSUE #6144: XSS vulnerability with Table Export extension

1.20.0

### Description

Hello and thank you for the wonderful library! We use it extensively in our app. However, I think we've identified an XSS vulnerability in the Export plug-in.

If you set the exportOptions in your Bootstrap Table to true, then you can force arbitrary Javascript to execute (see the attached PoC). The problem is actually in the jQuery Table Export plug-in, and I've reported it to them as well. But I figure you might also want to fix it here, just in case.

I think the problem can be worked-around by using a corrected onCellHtmlData callback method - which it looks like the library is already attempting to do. However, as evidenced by the vuln, I think for some reason that callback isn't getting executed, and the default onCellHtmlData callback is firing instead, and that default implementation does appear to be vulnerable.

### Example(s)

https://live.bootstrap-table.com/code/uberbrady/11033

Found by @uberbrady

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6144 
Fix #4005

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

Fixed XSS vulnerability bug by onCellHtmlData.

**💡Example(s)?**
* #6144
Before: https://live.bootstrap-table.com/code/uberbrady/11033
~~After: https://live.bootstrap-table.com/code/wenzhixin/11422~~
After: https://live.bootstrap-table.com/code/wenzhixin/11486

* #4005
https://live.bootstrap-table.com/code/wenzhixin/11423

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/export/bootstrap-table-export.js
+++ b/src/extensions/export/bootstrap-table-export.js
@@ -23,15 +23,7 @@ $.extend($.fn.bootstrapTable.defaults, {
   showExport: false,
   exportDataType: 'basic', // basic, all, selected
   exportTypes: ['json', 'xml', 'csv', 'txt', 'sql', 'excel'],
-  exportOptions: {
-    onCellHtmlData (cell, rowIndex, colIndex, htmlData) {
-      if (cell.is('th')) {
-        return cell.find('.th-inner').text()
-      }
-
-      return htmlData
-    }
-  },
+  exportOptions: {},
   exportFooter: false
 })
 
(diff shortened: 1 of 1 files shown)


djhvscf:
For the export plugin the example is not working, this is the error: Cannot read properties of undefined (reading 'fileName')

djhvscf (changes requested):
For the export plugin the example is not working, this is the error: Cannot read properties of undefined (reading 'fileName')

The filter control example works as expected

UtechtDustin:
Don't work for me either, but i get the error message `Uncaught TypeError: o.exportOptions is undefined`

wenzhixin:
Updated the example: https://live.bootstrap-table.com/code/wenzhixin/11486

djhvscf (approved):

Issue #6024, opened 2022-01-10
Add TypeScript definitions for the Extensions


JJ-M:
### Description

Currently there is a TypeScript definition for core BootstrapTable code but not for the extensions, so if you coding are in TypeScript and try to use an Extension the TypeScript compiler will complain as it does not recognise the additional attributes (For reference I'm trying to use Export). The only workaround for this it to try to extend the interface definitions in a new definition file.

I considered creating a pull request for this, but realise this needs a architectural decision on whether to extend the current index.d.ts to include all the options and types for the Extensions or whether to add a definition for each extension....

wenzhixin:
I haven't used TypeScript yet, do you have any ideas? @DustinUtecht @djhvscf

UtechtDustin:
I also haven't used TypeScript yet, but i guess we can extend the [index.d.ts](https://github.com/wenzhixin/bootstrap-table/blob/develop/index.d.ts).
Looks quite easy, if im correct the syntax always looks like `<option name>?: <data type>`.
As soon i have time i will check it!

JJ-M:
If you are going to have a look, please can you fix this one too - just some missing ?s ... #6023

UtechtDustin:
@JJ-M sure, could you maybe create a skeleton Typescript  repository for testing ?

nwalters512:
Weighing in re: approach per request from @UtechtDustin over on #6153. From what I've seen in the wild on other projects, trying to write TypeScript types for things with plugin systems (especially those where the plugins are implicit by adding another script on the page) is inevitably tricky precisely because of this issue that was pointed out in the original request: 

> this needs a architectural decision on whether to extend the current index.d.ts to include all the options and types for the Extensions or whether to add a definition for each extension....

If you add the plugin types to the "core" types, you end up in a situation where type checking succeeds even though someone may have forgotten to enable a certain extension, so you'd get runtime failures. Alternatively, you could force the user to construct a type that describes which plugins they have enabled. You could still have the same problem as above if you forget to keep things in sync, but it's less likely.

An example of this that I've seen in the wild is actually for another table library: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table (see "Example type file" in that readme). You get more precise types at the cost of a lot of manual work. There's also really no way for you to use different types in different parts of your app (say, if you include the `Export` plugin on one page but not on another).

[...]

PR #6153, opened 2022-05-13
types: Allow event handlers to return void

nwalters512:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Followup to https://github.com/wenzhixin/bootstrap-table/pull/6116 that does the more correct thing of allowing the handler functions to not return anything at all. This is more correct than the previous behavior, as having a handler function without an explicit `return ...` wasn't allowed by the previous types.

**☑️Self Check before Merge**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -149,51 +149,51 @@ export interface BootstrapAjaxParams {
 }
 
 export interface BootstrapTableOptions {
-  onCheck?: (row: any, $element: JQuery<HTMLElement>) => boolean | undefined;
+  onCheck?: (row: any, $element: JQuery<HTMLElement>) => boolean | void;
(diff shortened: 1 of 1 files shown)


UtechtDustin (approved):
Thank you!

UtechtDustin:
@nwalters512 could you maybe help us out with #6024 ?

nwalters512:
@UtechtDustin I don't have a ton of bandwidth to help out with the implementation right now, but I can weight in re: approach on that ticket.

PR #6190, opened 2022-05-30
Add else condition when one of the element is null

### LINKED ISSUE #6189: Bug on nested search with null child

1.20.2 

### Description

When you've a field that gets its value from a nested property (example: id.userId) and the lowest level is null, when you make a search what the formatter receives is the upper object, in this case id:

### Example(s)

https://live.bootstrap-table.com/code/lvalladares/11643

You can search "first" or "second" and you'll see how nothing is filtered and there is an error in the console because the formatter is getting an object instead an string or null, like this:

Uncaught Error: Syntax error, unrecognized expression: test-[object Object]

### Possible Solutions

This if: https://github.com/wenzhixin/bootstrap-table/blob/8a01294dabd4310a42d381f89d15c4f8ac5eefe0/src/bootstrap-table.js#L1035

needs an else, that makes value = null if one of the elements in the chain is null

### Additional Context

_No response_

lvalladares:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #6189

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

In order to avoid returning parent value, adds a condition that return null if one of the elements is null

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1034,6 +1034,9 @@ class BootstrapTable {
             for (let i = 0; i < props.length; i++) {
               if (value[props[i]] !== null) {
                 value = value[props[i]]
+              } else {
+                value = null
+                break;
               }
             }
           } else {


wenzhixin:
Please provide [before example and after example](https://live.bootstrap-table.com) to show the problem. Thanks!

lvalladares:
Before: https://live.bootstrap-table.com/code/lvalladares/11643
After: https://live.bootstrap-table.com/code/lvalladares/11658

Try to search on both

UtechtDustin (approved):

PR #6275, opened 2022-07-18
fix/6274

### LINKED ISSUE #6274: Multi sort extension: showMultiSortButton default behavior

1.18.0 and above

### Description

Version 1.18.0 introduced a (faulty in my opinion) behavior of the multi-sort extension:

https://github.com/wenzhixin/bootstrap-table/commit/93c4b0f5e6bd0799b01b0e92f2533a3fee9b2e9a#diff-72b9dc228479d427dd79d91fc54c0ee5558df0d5ef969dad6c73a3f693b02669R595

The code responsible for adding the multi-sort button was moved outside the `if (this.options.showMultiSort)` statement. As a consequence all bootstrap tables with the multi sorting disabled (parameter `data-show-multi-sort` set to false) have the multi-sort button visible (unless explicitly hidden with the `data-show-multi-sort-button` parameter set to false). Although visible the button doesn't have any onclick handlers attached so it doesn't work anyways.

The last bootstrap-table version which works as expected is 1.16.0 (for some reason version 1.17.1 is producing a javascript runtime error, even in the fiddle above).

I've found the following issue back from 2020 which mentions this problem: https://github.com/wenzhixin/bootstrap-table/issues/5343 however the main focus there was that the `data-show-multi-sort-button` parameter was **ignored** altogether. The issue was fixed by allowing you to explicitly hide the button. Unfortunately the default behavior of showing the button even when the multi sort is disabled remained as it was. I have found no other issues mentioning the problem.

### Example(s)

Here's the fiddle to recreate the error (version 1.18.0, but any version greater than that still works):

https://live.bootstrap-table.com/code/tomalla5120x/12120

Change the bootstrap-table version to 1.16.0 to see the problem disappear.

### Possible Solutions

I've fixed the problem temporarily by guarding the code for adding the multi sort button with the following if statement: `if (this.options.showMultiSort && this.options.showMultiSortButton)` (i.e. adding the `this.options.showMultiSort` condition suffices).

### Additional Context

_No response_

UtechtDustin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #6274 

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

The multiple sort button will no longer rendered if the extension is not enabled.

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/tomalla5120x/12120
After: https://live.bootstrap-table.com/code/UtechtDustin/12121

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
+++ b/src/extensions/multiple-sort/bootstrap-table-multiple-sort.js
@@ -584,7 +584,7 @@ BootstrapTable.prototype.initToolbar = function (...args) {
     that.onMultipleSort()
   }
 
-  if (this.options.showMultiSortButton) {
+  if (this.options.showMultiSort && this.options.showMultiSortButton) {
     this.buttons = Object.assign(this.buttons, {
       multipleSort: {
         html: Utils.sprintf(theme.html.multipleSortButton,


wenzhixin (approved):

ksodari:
I think it is better to make this feature reverse.
By default **disable** the render of multi sort button 
show-multi-sort-button="false" 
That way, we don't have to keep disabling everywhere throughout the system. 

Just to give you a scenario 
-----------------------------
- I have system where I am using bootstrap-table in every pages wherever needed 
- The npm packages are minified and imported in the application globally from default layout template. 
- Now, when I globally import multisort extension as well
- It automatically renders multisort button even when I don't need it.
Hence, it would be easier to just enable in which table it is required throughout the system by adding render true option on table initialization or html table properties. 

Just some thought for better use case.

UtechtDustin:
@ksodari i don't know what do you mean ?
It's disabled by default, also its a bad practice to import the bootstrap-table (and other lib's) globally.
You should only import that libs which are needed on this page, that is also the reason why webpack etc. exists.

Example: https://live.bootstrap-table.com/code/UtechtDustin/12234

PR #6281, opened 2022-07-22
Show/hide/toggle multiple rows

### LINKED ISSUE #5711: hideRow for multiple rows
Hi,

there is a method `hideRow` that allows you to hide particular row by index or unique id. However when we want to hide many rows in the table with more data and you execute `hideRow` in a loop, the problem with performance may be huge.
I figured out it's because of this code executed for each row:

```
this.initBody(true);
this.initPagination();
```

The solution would be to write another method accepting an array of indexes/unique ids, like this one:

```js
_toggleRows(params, visible) {
        // do the same as _toggleRow but for an array
        ...
        // execute these 2 functions once
        this.initBody(true);
        this.initPagination();
      }
```

This would ensure that last two methods are executed just once, not every hidden row.
To note that for showing already hidden rows there is no such problem, as we have `getHiddenRows(true)`.

9r0x:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Improvement**
- [x] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5711

**📝Changelog**

* Added APIs for show/hide/toggle multiple rows at a time

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**

[https://live.bootstrap-table.com/code/9r0x/12229](https://live.bootstrap-table.com/code/9r0x/12229)

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/api/methods.md
+++ b/site/docs/api/methods.md
@@ -285,18 +285,18 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
 - **Example:** [Show/Hide Loading](https://examples.bootstrap-table.com/#methods/show-hide-loading.html)
 
-## hideRow
+## hideRows
 
 - **Parameter:** `params`
 
 - **Detail:**
 
-  Hide the specified row. The params must contain at least one of the following properties:
+  Hide the specified row(s). The params must contain at least one of the following properties:
 
-  * `index`: the row index.
-  * `uniqueId`: the value of the uniqueId for that row.
+  * `indexes`: the array of row indexes.
+  * `uniqueIds`: the array of uniqueIds for rows.
 
-- **Example:** [Show/Hide Row](https://examples.bootstrap-table.com/#methods/show-hide-row.html)
+- **Example:** [Show/Hide Rows](https://examples.bootstrap-table.com/#methods/show-hide-row.html)
 
 ## insertRow
 
@@ -496,18 +496,18 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
 
 - **Example:** [Show/Hide Loading](https://examples.bootstrap-table.com/#methods/show-hide-loading.html)
 
-## showRow
+## showRows
 
 - **Parameter:** `params`
 
 - **Detail:**
 
-  Show the specified row. The params must contain at least one of the following properties:
+  Show the specified row(s). The params must contain at least one of the following properties:
 
-  * `index`: the row index.
-  * `uniqueId`: the value of the uniqueId for that row.
+  * `indexes`: the array of row indexes.
+  * `uniqueIds`: the array of uniqueIds for rows.
 
-- **Example:** [Show/Hide Row](https://examples.bootstrap-table.com/#methods/show-hide-row.html)
+- **Example:** [Show/Hide Rows](https://examples.bootstrap-table.com/#methods/show-hide-row.html)
 
 ## toggleDetailView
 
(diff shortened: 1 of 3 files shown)


9r0x:
Hi, is there a general guideline for updating changelog and examples?

Also, since this at the moment slightly breaks the previous API, please let me know if there's a better way to incorporate the features.

UtechtDustin:
For the changelog just describe what this PR does (Adds a new methods [...]).
We should avoid the BC, you could keep the old functions which calls the new functions internaly with one row (or did i miss something?)
Also please provide us an example using our [editor](https://live.bootstrap-table.com/) (as described in the issue template).

9r0x:
There are 3 ways I could think of 
1. Keep the same singular(e.g. `hideRow`) form, but internally allows an array of rows as input; simple but less intuitive because all parameters(row, index, uniqueID) are singular.
2. Add additional plural form(e.g. `hideRows`), which makes it a bit confusing with 6(vs 2 before) functions representing similar features
3. Replace singular form with plural form, simplest and most intuitive, but breaks API.

I would proceed with 2 but hide the singular form from the doc, so the doc would be less confusing while still internally supports historical versions

9r0x:
It seems the Online editor keeps using an old version of my code before my last force-push, is there a way to refresh the cache or whatever that causes the editor to use an old version of the library?

I by accident added an unused require statement in the previous push, and immediately force-pushed with the fix, then the editor still reports the missing library.
Screen Shot 2022-07-24 at 13 38 08

9r0x:
Change log:
* Added APIs for show/hide/toggle multiple rows at a time


Example:
[https://live.bootstrap-table.com/code/9r0x/12229](https://live.bootstrap-table.com/code/9r0x/12229)

site/docs/api/methods.md
```diff
@@ -285,18 +285,18 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
...
 - **Example:** [Show/Hide Loading](https://examples.bootstrap-table.com/#methods/show-hide-loading.html)
 
-## hideRow
+## hideRows
 
 - **Parameter:** `params`
 
 - **Detail:**
 
   Hide the specified row. The params must contain at least one of the following properties:
```
UtechtDustin, line 294: `Hide the specified row(s)[...]`

src/bootstrap-table.js
```diff
@@ -2691,36 +2691,64 @@ class BootstrapTable {
...
+
+  _toggleRow (row, visible) {
     if (!row) {
       return
     }
-
     const index = Utils.findIndex(this.hiddenRows, row)
 
-    if (!visible && index === -1) {
+    if (visible === undefined) {
```
UtechtDustin, line 2740: Could you somehow improve this if, ifelse, else part ?
`this.hiddenRows.push(row)` and  `this.hiddenRows.splice(index, 1)` are duplicated here, so we could merge the if's.

site/docs/api/methods.md
```diff
@@ -496,18 +496,18 @@ The calling method syntax: `$('#table').bootstrapTable('method', parameter)`.
...
 
   Show the specified row. The params must contain at least one of the following properties:
 
-  * `index`: the row index.
-  * `uniqueId`: the value of the uniqueId for that row.
+  * `indexes`: the array of row indexes.
+  * `uniqueIds`: the array of uniqueIds for rows.
 
-- **Example:** [Show/Hide Row](https://examples.bootstrap-table.com/#methods/show-hide-row.html)
+- **Example:** [Show/Hide Rows](https://examples.bootstrap-table.com/#methods/show-hide-row.html)
```
UtechtDustin, line 510: This would link to a example which don't show hiding multiple rows at once, could you please improve the example (https://github.com/wenzhixin/bootstrap-table-examples) ?
It would also be good if the example contains:
- show/hide/toggle one row
- show/hide/toggle multiple rows

UtechtDustin: any news here @9r0x ?

UtechtDustin:
Lets us know if you are still intrested to help with this PR, we will reopen the PR then.

Issue #6251, opened 2022-07-08
Filter control : select box with name containing "&"


sdespont:
### Bootstraptable version(s) affected

1.20.2

### Description

When rows have name containing "&", the select list show the name without the "&" char, which doesn't permit to filter by this name.

In this example, the name should be "Z&B", but "ZB" is listed. Selecting "ZB" doens't permit to display the "Z&B" rows:

![image](https://user-images.githubusercontent.com/646632/177977497-3ae41bda-a87e-46fe-b69d-868c54c20a6b.png)

### Example(s)

https://live.bootstrap-table.com/code/sdespont/12051

UtechtDustin:
Duplicate of #6130

Issue #6130, opened 2022-05-06
Bug filter-control when using & in value


Zecrum:
### Bootstraptable version(s) affected

1.20.0

### Description

Since the version 1.20.0 when we use "&", the select filter control does not work.

### Example(s)

With 1.19.1 (work) : https://live.bootstrap-table.com/code/Zecrum/11402
With 1.20.0 (don't work) : https://live.bootstrap-table.com/code/Zecrum/11400

### Possible Solutions

Fix in issue #5385

sdespont:
@UtechtDustin PR provided : https://github.com/wenzhixin/bootstrap-table/pull/6323

djhvscf:
@sdespont merged

PR #6323, opened 2022-08-09
Fix filter control special char

sdespont:
**🤔Type of Request**
- [X] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

https://github.com/wenzhixin/bootstrap-table/issues/6251
https://github.com/wenzhixin/bootstrap-table/issues/6130

**📝Changelog**

- [ ] **Core**
- [X] **Extensions**

Because HTML is removed from the cell value in order to extract only the text, encoded char like `&` is removed too resulting that the filetring of cell containing this char is not possible.

**💡Example(s)?**

Before : https://live.bootstrap-table.com/code/sdespont/12377
After : https://live.bootstrap-table.com/code/sdespont/12376

**☑️Self Check before Merge**

- [X] Doc is updated/provided or not needed
- [X] Demo is updated/provided or not needed
- [X] Changelog is provided or not needed

### DIFF
--- a/src/extensions/filter-control/utils.js
+++ b/src/extensions/filter-control/utils.js
@@ -89,7 +89,7 @@ export function sortSelectControl (selectControl, orderBy) {
   for (let i = 0; i < $selectControl.options.length; i++) {
     tmpAry[i] = new Array()
     tmpAry[i][0] = $selectControl.options[i].text
-    tmpAry[i][1] = $selectControl.options[i].value
+    tmpAry[i][1] = "test"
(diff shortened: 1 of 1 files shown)


UtechtDustin:
The Example work fine, but i'm not sure if this could create sercurity issues e.g. xss.
@djhvscf It's your main code, so please check it :-)

djhvscf (approved):

PR #6346, opened 2022-08-21
WIP: v1.21.1

wenzhixin:
# Next release

# v1.21.1

### Plan time

* 2022-09-10 or 2022-09-20


wenzhixin:
Hello @UtechtDustin @djhvscf, after merged a pull request, please help to update the changelog above.
ref: https://github.com/wenzhixin/bootstrap-table/blob/develop/CHANGELOG.md?plain=1#L4

UtechtDustin:
Does it makes sense to use a open PR for this ?
Or could it make sense to update the changelog.md within the PR's directly ?
This has a few advantages:
- We can't forgett a changelog entry (write a check which fails if the changelog.md was not modified)
- Other developer who submit PR's also can help to complete the changelog
- If we checkout any commit in the develop branch we want, we can see in the changelog which until this commit.
- We don't have a open PR for the next version

What we have to do for this:
- Add documentation that the dev's have to add a changelog entry and how it should look like.

What do you think @wenzhixin @djhvscf ?

wenzhixin:
> Add documentation that the dev's have to add a changelog entry and how it should look like.
> Add a check if the changelog.md was modified.

It's great.

Issue #5526, opened 2021-01-05
Plan to remove jQuery


djhvscf:
**Description**  
We should be planning the process to remove jQuery. I know that this could be a huge work but I think it worth. My proposal is that we need to have a strong unit testing and then we need to start removing jQuery.

I was working on this and you can see the progress in this branch

https://github.com/wenzhixin/bootstrap-table/tree/feature/modular_js

Please let me know your thoughts

ufhy:
up

TiKevin83:
This would be really helpful given bootstrap 5's dropping of jQuery. I was able to drop a dependency on jQuery on https://runs.tas.bot/runs.html by migrating from bootstrap 4 to 5, but when implementing more advanced table functionality using boostrap-table I had to add jQuery back in (though it does seem to at least work fine with jQuery slim).

jimmywarting:
bump

UtechtDustin:
Duplicate of #4796
@jimmywarting please check this issue, there is a recent answer :-)

Issue #4796, opened 2020-01-20
Using BootstrapTable without jQuery


WeanaSchnitzl:
Hi,

i tried using BootstrapTable without jQuery. I ran into some issues where i don't know, if it is a bug or a feature ;-) 
 The following options caused trouble in the "NOT-jQuery" constructor:

Line 3871 : formatLoadingMessage is not checked if exists before used or no default defined
Line 3946: columns[0] is accessed without checking if columns ist set and is array and array len > 0
Line 4382:  formatSearch called withoout checking if exists or no default defined
Line 4383: formatClearSearch same issue
Line 5531: cannot read property 'apply' of undefined

i stopped searching when i saw the code in 5531 because i didn't understand it (i'm new to js!)

hope i don't blame me and this helps you!

Regards
Werner

wenzhixin:
It is not supported at this time and will be removed in future versions.

paritoshromy:
Any update on v2

UtechtDustin:
@paritoshromy No, this issue the largest one which needs a lot of time.
At the moment this issue has no high priority.

jimmywarting:
having to depend on jQuery is a tiny bit of a turn of for me to even consider starting using `bootstrap-table` but knowing that it might eventually be removed maybe i can look the other way...

hope like v2 ditch all jQuery & IE support (considering IE is mostly dead now anyway)

PR #6348, opened 2022-08-21
Use object.assign instead of extend

jimmywarting:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
- part of #4796 #5526 (got to start out small)

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/auto-refresh/bootstrap-table-auto-refresh.js
+++ b/src/extensions/auto-refresh/bootstrap-table-auto-refresh.js
@@ -44,19 +44,17 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
   initToolbar (...args) {
     if (this.options.autoRefresh) {
-      this.buttons = assign(this.buttons, {
-        autoRefresh: {
-          html: `
-            <button class="auto-refresh ${this.constants.buttonsClass}
-              ${this.options.autoRefreshStatus ? ` ${this.constants.classes.buttonActive}` : ''}"
-              type="button" name="autoRefresh" title="${this.options.formatAutoRefresh()}">
-              ${ this.options.showButtonIcons ? Utils.sprintf(this.constants.html.icon, this.options.iconsPrefix, this.options.icons.autoRefresh) : ''}
-              ${ this.options.showButtonText ? this.options.formatAutoRefresh() : ''}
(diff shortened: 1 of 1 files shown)


UtechtDustin:
I'm not sure if i'm a fan to use only `assign` instead of `Object.assign`.
Does this have any advantages, because it's not that readable (especially for new devs in the project).

Also i'm not sure if we should replace $.extend with `Object.assign` or if we should write our own method for that (i think we need a custom function for deep copy either way).
The `extend` function from `jquery`  is faster then `Object.assign` if [this benchmark](https://www.measurethat.net/Benchmarks/Show/3447/0/objectassign-vs-spread-vs-jqueryextend) is correct.

jimmywarting:
Executed the test myself... it showed that spread was a clear winner and that jQuery.extend was the slowest. 
https://www.measurethat.net/Benchmarks/ShowResult/326203

using `const { assign } = Object` instead of `Object.assign` makes it a tiny bit more friendlier for the other stuff to be GC, helps with code minification and also has a faster path lookup to increase perf

anyhow this stuff didn't extend anything deeply so nothing like extend deep functionallity had to be written

UtechtDustin:
Firefox: https://www.measurethat.net/Benchmarks/ShowResult/326214

> using const { assign } = Object instead of Object.assign makes it a tiny bit more friendlier for the other stuff to be GC, helps with code minification and also has a faster path lookup to increase perf b/c it's cached to a more closer local scope

I think our build script automatically optimize that, or not @wenzhixin  ?

wenzhixin:
> I think our build script automatically optimize that.

Yes

UtechtDustin:
@wenzhixin @djhvscf your opinions about this ?

wenzhixin:
I think we can modify the code in all `src`, not just this extension.

UtechtDustin:
@wenzhixin and what do you think about the benchmarks ?

wenzhixin:
Updated in #6560

PR #6349, opened 2022-08-21
[WIP]Changelog action and documentation

UtechtDustin:
This PR will add an Github action which checks if the changelog file was modified and adds documentation about the changelog.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,15 @@ jobs:
           repository: 'wenzhixin/bootstrap-table-examples'
           path: './tools/bootstrap-table-examples'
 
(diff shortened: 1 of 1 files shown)


wenzhixin:
> - [x] Whether we need to update the CHANGELOG

I think we can add a checklist to the pull Request template, "Whether we need to update the CHANGELOG", the default value is `true`. 

And then, the GitHub actions get the value through the GitHub API. If it is `true`, check whether the new changelog format is correct, if it is `false`, ignore it and do not check.

UtechtDustin:
I close this PR, because @wenzhixin want to create a magic github action :-)

PR #6400, opened 2022-09-17
Fix #6399, HTML-formatted editable fields, issue #6399

### LINKED ISSUE #6399: Extension "editable", HTML-formatted values
### Description

Hello! Unfortunately "editable" extension doesn't allow displaying of HTML-formatted custom values.
Here is the quick fix...

Original code in "bootstrap-table-editable.js":

`              return "<a href=\"javascript:void(0)\"\n            data-name=\"".concat(column.field, "\"\n            data-pk=\"").concat(row[_this.options.idField], "\"\n            data-value=\"").concat(result, "\"\n            ").concat(editableDataMarkup.join(''), "></a>");
`
The fix is:

```
              return "<a href=\"javascript:void(0)\"\n            data-name=\"".concat(column.field, "\"\n            data-pk=\"").concat(row[_this.options.idField], "\"\n            data-value=\"")
                  .concat(undefined === value ? "" : value, "\"\n            ")
                  .concat(editableDataMarkup.join(''), ">"+result+"</a>");
```

So it displaying correct HTML-formatted `result` and still have correct value in the `data-value` attribute.

P.S. Also I wish I could set up all dataset programmatically, to be able to pass the `function` to `data-pk` value, but this is more complicated, will require fixes in the main "bootstrap-table.js"...

utilmind:
Fix described on https://github.com/wenzhixin/bootstrap-table/issues/6399

**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -48,7 +48,6 @@ $.BootstrapTable = class extends $.BootstrapTable {
       }
 
       const editableOptions = {}
-      const editableDataMarkup = []
       const editableDataPrefix = 'editable-'
       const processDataOptions = (key, value) => {
         // Replace camel case with dashes.
@@ -77,26 +76,26 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
         $.each(column, processDataOptions)
 
-        $.each(editableOptions, (key, value) => {
-          editableDataMarkup.push(` ${key}="${value}"`)
-        })
-
-        let noEditFormatter = false
         const editableOpts = Utils.calculateObjectValue(column,
           column.editable, [index, row], {})
+        let noEditFormatter = editableOpts.hasOwnProperty("noEditFormatter") &&
+                              editableOpts.noEditFormatter(value, row, index, field);
 
-        if (editableOpts.hasOwnProperty('noEditFormatter')) {
-          noEditFormatter = editableOpts.noEditFormatter(value, row, index, field)
+        if (noEditFormatter) {
+            return noEditFormatter
         }
 
-        if (noEditFormatter === false) {
-          return `<a href="javascript:void(0)"
-            data-name="${column.field}"
-            data-pk="${row[this.options.idField]}"
-            data-value="${result}"
-            ${editableDataMarkup.join('')}></a>`
-        }
-        return noEditFormatter
+        let editableDataMarkup = ""
+        $.each(editableOptions, (key, value) => {
+          editableDataMarkup += ` ${key}="${value}"`
+        })
+
+        return `<a href="javascript:void(0)"
+          data-name="${column.field}"
+          data-pk="${row[this.options.idField]}"
+          data-value="${value || ""}"
+          data-value="${result}"
+          ${editableDataMarkup}>${result}</a>` // expand all data-editable-XXX
       }
     })
   }
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -94,7 +94,6 @@ $.BootstrapTable = class extends $.BootstrapTable {
           data-name="${column.field}"
           data-pk="${row[this.options.idField]}"
           data-value="${value || ""}"
-          data-value="${result}"
           ${editableDataMarkup}>${result}</a>` // expand all data-editable-XXX
       }
     })
--- a/src/extensions/editable/bootstrap-table-editable.js
+++ b/src/extensions/editable/bootstrap-table-editable.js
@@ -78,14 +78,15 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
         const editableOpts = Utils.calculateObjectValue(column,
           column.editable, [index, row], {})
-        const noEditFormatter = editableOpts.hasOwnProperty("noEditFormatter") &&
-                                editableOpts.noEditFormatter(value, row, index, field);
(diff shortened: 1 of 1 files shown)


utilmind:
Please update changelog, if this merge request will be accepted. Thanks!

UtechtDustin:
The changes are made in the dist folder instead of the src folder, which means the PR is currently invalid.
Please reapply the changes to the source folder and also provide us examples using our [editor](https://live.bootstrap-table.com/).

utilmind:
Hello @UtechtDustin !

Example provided at https://live.bootstrap-table.com/code/utilmind/12883 (see `window.tablePrice()` method and Price column in the table, with broken HTML formatting due to unescaped "-characters within attributes of the HTML tag).

The source code also updated with minor optimization to avoid conversion of template literals to slow concat()'s, during es6 -> es5 translation.  (MDN recommends  to use the string concatenation operators (+, +=) instead of `concat()` method for perfomance reasons. Discussion at https://stackoverflow.com/questions/16124032/js-strings-vs-concat-method).

UtechtDustin:
Sorry for the late feedback it seems we all are quite busy right now.
The PR should only update the source code, the dist code will be updated while realeasing a new version.

utilmind:
@UtechtDustin now suggested updates are only in source code file. Other changes (in dist folder) reverted to original.

src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -48,11 +48,10 @@ $.BootstrapTable = class extends $.BootstrapTable {
       }
 
       const editableOptions = {}
-      const editableDataMarkup = []
       const editableDataPrefix = 'editable-'
       const processDataOptions = (key, value) => {
         // Replace camel case with dashes.
-        const dashKey = key.replace(/([A-Z])/g, $1 => `-${$1.toLowerCase()}`)
+        const dashKey = key.replace(/([A-Z])/g, $1 => "-" + $1.toLowerCase())
```
UtechtDustin, line 54: I don't think we need that change, as we optimize/transpile the source code on a new release (see dist folder).
Also we use templated strings everywhere in the project, so if it makes sense to refactor it we should do it in the whole project.

src/extensions/editable/bootstrap-table-editable.js
```diff
@@ -114,7 +110,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
       }
 
       const data = this.getData({ escape: true })
-      const $field = this.$body.find(`a[data-name="${column.field}"]`)
+      const $field = this.$body.find('a[data-name="' + column.field + '"]')
```
UtechtDustin, line 113: Same as in line 54

src/extensions/editable/bootstrap-table-editable.js
utilmind: reverted to original in this line. (Don't remember why I downgraded syntax to ES5, this was half year ago.)

src/extensions/editable/bootstrap-table-editable.js
utilmind: reverted to original in this line. (Don't remember why I downgraded syntax to ES5, this was half year ago.)

src/extensions/editable/bootstrap-table-editable.js
utilmind: Oh, I remembered! This because I modified file in "dist" folder which has ES5 syntax! Anyway this line reverted now.

UtechtDustin:
I tried to test you're changes with you're fork (https://live.bootstrap-table.com/code/UtechtDustin/14259).
But the editor is broken, because you're fork is not up to date with our develop branch.
Please fix that, so we can test it with our editor.

utilmind:
@UtechtDustin I have pulled latest updates and did the fix again.
I hope this fix will not be ignored during next 6 months :)

It really fixes the described issue #6399 of HTML formatted editable values.

Thanks!

UtechtDustin:
Please provide us an working example, i'm was not able to create one with you're PR.

UtechtDustin:
I saw that example, but it uses the version 1.21.0.   
If i run the example with you`re PR (utilmind:develop) the whole table crashes.

utilmind:
@UtechtDustin ok, here is an example which supposed to get sources from utilmind:develop: https://live.bootstrap-table.com/code/utilmind/15558
However, it does not. I opened the source code and see that it still use "editable" from CDN.

image

However, in any case, I guarantee that it fixes the mentioned issue since September 2022.

UtechtDustin:
You did it correctly and the editor works fine!
What you marked in red is just the x-editable plugin not the edtiable bootstrap-table plugin!
Check the second script tag on the body, it's the develop branch of your fork.

utilmind:
@UtechtDustin now it's really fixed: https://live.bootstrap-table.com/code/utilmind/15567
Removed 1 odd line of code and purged jsDelivr cache to reflect update.

UtechtDustin:
Seems working now, thanks!  
We will merge it, as soon the linting issues are fixed.

UtechtDustin (approved):

UtechtDustin:
Thank you!

PR #6419, opened 2022-09-22
Introduces new searchClearSelector

### LINKED ISSUE #6406: Add a `searchClearButtonSelector`
### Description

Add a `searchClearButtonSelector` option.

Together with the existing `searchSelector` this would make it very easy to create a different looking search bar **without having to write any custom javascript**.

E.g.:

image

Example html:

```
<form class="form-group form-group-light">
  <input id="customSearch" class="form-control" type="text" placeholder="Text input">
  <button id="customSearchClearButton" type="button" class="btn btn-primary">X</button>
</form>

<table
  id="table"
  data-toggle="table"
  data-search="true"
  data-search-selector="#customSearch"
  data-search-clear-button-selector="#customSearchClearButton"
  data-url="json/data1.json">
  <thead>
    <tr>
      <th data-field="id">ID</th>
      <th data-field="name">Item Name</th>
      <th data-field="price">Item Price</th>
    </tr>
  </thead>
</table>
```

marceloverdijk:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fixes #6406

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Introduces a new searchClearSelector table option.
Together with the existing `searchSelector` this would make it very easy to create a custom search bar without having to write any custom javascript as it automatically connects the BT events/methods to the selector.

**💡Example(s)?**

https://live.bootstrap-table.com/code/marceloverdijk/12945

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -919,6 +919,12 @@ class BootstrapTable {
       const $searchInput = Utils.getSearchInput(this)
 
       handleInputEvent($searchInput)
+
+      if (typeof opts.searchClearButtonSelector === 'string') {
+        $(opts.searchClearButtonSelector).click(() => {
+          this.resetSearch()
+        })
+      }
     }
   }
 
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -920,8 +920,8 @@ class BootstrapTable {
 
       handleInputEvent($searchInput)
 
-      if (typeof opts.searchClearButtonSelector === 'string') {
-        $(opts.searchClearButtonSelector).click(() => {
+      if (typeof opts.searchClearSelector === 'string') {
+        $(opts.searchClearSelector).click(() => {
           this.resetSearch()
         })
       }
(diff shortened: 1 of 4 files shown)


src/bootstrap-table.js
```diff
@@ -919,6 +919,12 @@ class BootstrapTable {
       const $searchInput = Utils.getSearchInput(this)
 
       handleInputEvent($searchInput)
+
+      if (typeof opts.searchClearSelector === 'string') {
```
UtechtDustin, line 923: We already handle the clear of the search somewhere, do we really need an additional event binding ?
Can't we replace the existing selector if the `searchClearSelector` option is set ?

Also does it make sense to restrict it to work only in combination with the `searchSelector`?

UtechtDustin (changes requested):

src/bootstrap-table.js
marceloverdijk: Oke thx for feedback. I will look into it later so if I can get your proposal working.

UtechtDustin: I guess you can move that part outside of that if, in a new if without that condition.

marceloverdijk: Hi @UtechtDustin I'm not sure what you mean unfortunately. Would you mind updating the PR ?

PR #6429, opened 2022-09-28
Indicate URL parameters for server side pagination

hepcat72:
I'd tried to use bootstrap table's server side pagination over a year ago and never figured out how to get it to work.  I was working in Django and tried to implement the view that the data-url was calling, but I was missing this key piece of information.  I happened upon the URL parameters in the last answer of a stack post and it was like an epiphany.  I went back to the documentation to see how I'd possibly missed the URL parameters that the view needed to be able to obtain, and was shocked to see that it wasn't a part of the documentation.  Hopefully, this addition to the documentation will make the server-side pagination attempts of others easier to implement...

And if there are other parameters that could be passed, such as for `serverSort`, it would be good to add those as well.

**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [x] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**

- [ ] **Core**
- [ ] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/api/table-options.md
+++ b/site/docs/api/table-options.md
@@ -1529,12 +1529,12 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
 
 - **URL parameters:**
 
-  When `data-side-pagination` is set to `server`, bootstrap table will make calls to the `data-url` with the following URL parameters:
+  When `sidePagination` is set to `server`, bootstrap table will make calls to the `data-url` with the following URL parameters:
   
-  - `offset` with a value between 0 and `total` - 1, indicating the first record to include
-  - `limit` with a value indicating the number of rows per page
+  - `offset` with a value between 0 and `total` - 1, indicating the first record to include.
+  - `limit` with a value indicating the number of rows per page.
 
-  When implementing server-side pagination, you must implement a JSON view in a format like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data).  That view must take the URL parameter values indicated above and return records starting at the `offset` index and returns the number of records indicated by `limit`.  For example, if you want records 11-20, your view must obtain the `offset=10` and `limit=10` from the URL string and return records like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data?offset=10&limit=10).
+  When implementing server-side pagination, you must implement a JSON view in a format like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data). That view must take the URL parameter values indicated above and return records starting at the `offset` index and returns the number of records indicated by `limit`.  For example, if you want records 11-20, your view must obtain the `offset=10` and `limit=10` from the URL string and return records like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data?offset=10&limit=10).
 
 - **Default:** `'client'`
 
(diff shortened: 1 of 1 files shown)


site/docs/api/table-options.md
```diff
@@ -1521,12 +1521,21 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
 
   Defines the side pagination of the table, can only be `'client'` or `'server'`.
   Using the `'server'` side requires setting the `'url'` or `'ajax'` option.
-
+
```
wenzhixin, line 1524: This change is not necessary.

site/docs/api/table-options.md
```diff
@@ -1521,12 +1521,21 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
...
   * [With server-side pagination](https://github.com/wenzhixin/bootstrap-table-examples/blob/master/json/data2.json)
 
+- **URL parameters:**
+
+  When `data-side-pagination` is set to `server`, bootstrap table will make calls to the `data-url` with the following URL parameters:
+  
+  - `offset` with a value between 0 and `total` - 1, indicating the first record to include
+  - `limit` with a value indicating the number of rows per page
+
+  When implementing server-side pagination, you must implement a JSON view in a format like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data).  That view must take the URL parameter values indicated above and return records starting at the `offset` index and returns the number of records indicated by `limit`.  For example, if you want records 11-20, your view must obtain the `offset=10` and `limit=10` from the URL string and return records like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data?offset=10&limit=10).
```
wenzhixin, line 1537: When `sidePagination` is set to `server`, the bootstrap table will make calls to the `data-url` with the following URL parameters:

  - `offset` with a value between 0 and `total` - 1, indicating the first record to include.
  - `limit` with a value indicating the number of rows per page.

  When implementing server-side pagination, you must implement a JSON view in a format like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data).  That view must take the URL parameter values indicated above and return records starting at the `offset` index and returns the number of records indicated by `limit`.  For example, if you want records 11-20, your view must obtain the `offset=10` and `limit=10` from the URL string and return records like [this example](https://examples.wenzhixin.net.cn/examples/bootstrap_table/data?offset=10&limit=10).

site/docs/api/table-options.md
wenzhixin, line 1524:
```suggestion
 
```

site/docs/api/table-options.md
```diff
@@ -1521,12 +1521,21 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
 
   Defines the side pagination of the table, can only be `'client'` or `'server'`.
   Using the `'server'` side requires setting the `'url'` or `'ajax'` option.
-
+
```
wenzhixin, line 1524:
```suggestion

```

wenzhixin (approved):

bababash:
Should this include searching? It seems that when `data-side-pagination="server"` the request URL also contains `search=""` thus disabling client side search

[Example](https://live.bootstrap-table.com/code/bababash/13433)

PR #6581, opened 2022-12-26
Added new table options "copyUnique" and "copyToLowerCase".

mmaravillo:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

Added a new table option `copyUnique` to copy only unique values to clipboard.
Added a new table option `copyToLowerCase` to convert values to lowercase; when used with `copyUnique`, guarantees uniqueness if there is any character case mismatch.

**💡Example(s)?**

https://live.bootstrap-table.com/code/mmaravillo/13892

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/extensions/copy-rows.md
+++ b/site/docs/extensions/copy-rows.md
@@ -56,6 +56,33 @@ This extension adds functionality for copying selected rows to the clipboard. Cu
 
 - **Default:** `'\n'`
 
+### copyToLowerCase
+
+- **Attribute:** `data-copy-to-lower-case`
+
+- **type:** `Boolean`
+
+- **Detail:**
+
+   Set `true` to convert values to lowercase when copying.
+
+- **Default:** `false`
+
+### copyUnique
+
+- **Attribute:** `data-copy-unique`
+
+- **type:** `Boolean`
+
+- **Detail:**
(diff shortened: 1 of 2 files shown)


UtechtDustin (changes requested):
Without Unique you're example would be
```
john@mycompany.com, alice@newcompany.com, bob@somecompany.com;alice@newcompany.com, bob@somecompany.com, john@mycompany.com;bob@somecompany.com, john@mycompany.com, alice@newcompany.com
``` 

With `copyUnique` it would be
```
john@mycompany.com;alice@newcompany.com;bob@somecompany.com
```

And i don't think this would be a good feature, as in your example in each row `john`, `alice` and `bob` is once the owner.
But With `copyRows` you only get one Row.
Sure if you check the values without context and column order every row have the same values and are not unique, but i don't think we should ignore the context and the column order.

I'm also unsure if we need a `copyToLowerCase` option, this looks more like a usecase you need. It would make more sense to have a `copyFormatter`.

PR #6633, opened 2023-02-07
Updated checkboxes and radios to match bootstrap 5

Nickno95:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
N/A

**📝Changelog**

Updated checkboxes and radios to match Bootstrap 5 styling (see https://getbootstrap.com/docs/5.3/forms/checks-radios/).
This allows for other features like the color of the checkboxes/radios to match the bootstrap primary color when customized

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/Nickno95/14139
After: https://live.bootstrap-table.com/code/Nickno95/14140

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -379,7 +379,7 @@ class BootstrapTable {
         if (column.checkbox) {
           text = ''
           if (!this.options.singleSelect && this.options.checkboxHeader) {
-            text = '<label><input name="btSelectAll" type="checkbox" /><span></span></label>'
+            text = '<label><input name="btSelectAll" type="checkbox" class="form-check-input" /><span></span></label>'
           }
           this.header.stateField = column.field
         }
@@ -682,7 +682,7 @@ class BootstrapTable {
 
             html.push(
               Utils.sprintf(this.constants.html.toolbarDropdownItem,
-                Utils.sprintf('<input type="checkbox" class="toggle-all" %s> <span>%s</span>',
+                Utils.sprintf('<div class="form-check"><input type="checkbox" class="toggle-all form-check-input" %s> <span>%s</span></div>',
                   allFieldsVisible ? 'checked="checked"' : '', opts.formatColumnsToggleAll())
               )
             )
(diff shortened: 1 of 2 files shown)


UtechtDustin (changes requested):
This PR would add the class `form-check-input` for every theme e.g. semanticui.
Please adjust it, so the class will only be used for BT5.

Nickno95:
> This PR would add the class `form-check-input` for every theme e.g. semanticui.
> 
> Please adjust it, so the class will only be used for BT5.

Any suggestion on best way for checking what the theme is set to in the javascript, maybe an example in the code that's already doing this? 
Another part of the issue in Bootstrap 5 is with needing an extra wrapping element that was not needed in Bootstrap 4, and I assume other frameworks don't need that either.

UtechtDustin:
We already do that quite often, a few hints:
https://github.com/wenzhixin/bootstrap-table/blob/ff06f54e48f8ffb4a7de2bb86e682997e5648ac1/src/constants/index.js#L8-L109
https://github.com/wenzhixin/bootstrap-table/blob/develop/src/themes/foundation/bootstrap-table-foundation.js
https://github.com/wenzhixin/bootstrap-table/blob/1a1c5cda6748997daf49d11d375788feea9f2747/src/bootstrap-table.js#L1248

marceloverdijk:
Actually I noticed this my self today using the Columns toolbar dropdown.
The checkboxes generated in the dropdown to toggle the column visibility are generated like:

```
<label class="dropdown-item dropdown-item-marker">
    <input type="checkbox" data-field="name" value="0" checked="checked">
    <span>Name</span>
</label>
```

and indeed does not use the BS theme (if using BS).

It would be great to have a way to have it generated as: 

```
<label class="dropdown-item dropdown-item-marker">
    <input class="form-check-input" type="checkbox" data-field="name" value="0" checked="checked">
    <span>Name</span>
</label>
```

@UtechtDustin are you working on this yourself?

marceloverdijk:
Just for reference if someone wants a quick but hack'ish solution.

```
$.BootstrapTable = class extends $.BootstrapTable {

  initToolbar() {
    super.initToolbar()
    $(`${this.options.buttonsToolbar} input[type='checkbox']`).addClass('form-check-input me-2')
  }
}
```

UtechtDustin:
@marceloverdijk No, i'm not working on this issue right now.

wenzhixin:
Updated in https://github.com/wenzhixin/bootstrap-table/pull/8145

PR #6711, opened 2023-03-28
Added searchable table option

fish-fly:
After the change in #6573, the latest version now sends searchable params to server-side pagination by default.
image

This will cause unknown errors such as 400 (Bad Request). 

In previous versions <= `v1.23.2`, there was no such behavior.
image

According to the instructions in the documentation and to maintain the original behavior, we need to disable this feature by default, and then set it to `true` to turn it on if necessary. 

So the `searchable` option (consistent with the column option) was added to control it.

---

**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #6573 

Added `searchable` table option to disable sending searchable params by default.

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**
* 1.23.3: https://live.bootstrap-table.com/example/welcome.html
* <= 1.23.2: https://live.bootstrap-table.com/code/fish-fly/14835
* Fixed: https://live.bootstrap-table.com/code/fish-fly/14836 

**☑️Self Check before Merge**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ ChangeLog
 
 #### Core
 
+- **New:** Added searchable table option to enable sending searchable (columns) parameters.
 - **Update:** Fixed Maximum call stack size exceeded error.
 - **Update:** Fixed getData bug with hidden rows.
 - **Update:** Added support for `select` form to the `searchSelector` option.
--- a/site/docs/api/table-options.md
+++ b/site/docs/api/table-options.md
@@ -1196,10 +1196,25 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
   Notes:
   - If you want to use a custom search input, use the [searchSelector](https://bootstrap-table.com/docs/api/table-options/#searchSelector).
   - You can also search via regex using the [regexSearch](https://bootstrap-table.com/docs/api/table-options/#regexSearch) option.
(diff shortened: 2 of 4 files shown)


wenzhixin:
Update the docs and add an example to https://github.com/wenzhixin/bootstrap-table-examples.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ ChangeLog
 
 #### Core
 
+- **New:** Added searchable table option to disable sending searchable params.
```
UtechtDustin, line 8:
```suggestion
- **New:** Added searchable table option to enable sending searchable (columns) parameters.
```
Default is `false` so the option allows enabling the feature, also added the hint that this options is a table option as we already have it as column option.

site/docs/api/table-options.md
```diff
@@ -1196,11 +1196,26 @@ The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
...
 
+## searchable
+
+- **Attribute:** `data-searchable`
+
+- **Type:** `Boolean`
+
+- **Detail:**
+
+  Set `true` to send searchable params to server-side pagination.
```
UtechtDustin, line 1213:
```suggestion
  Set `true` to send [searchable params](https://bootstrap-table.com/docs/api/column-options/#searchable) to the server while usign server-side pagination.
```

UtechtDustin:
Btw. you also can add @djhvscf and me as reviewer :)

Issue #6798, opened 2023-05-18
Switchable improvement ideas


marceloverdijk:
### Description

For the switchable functionality I have 2 requests/ideas for an improvement.

1) When e.g. having a table with the columns `name`, `a`, `b`, `c`, `d`, I always want the column `name` to be visible (so effectively `visible=true, switchable-false`). 
But with the other fields switchable one of them needs to be selected.
What I would like is that the `name` field is displayed **and counted** as switchable field, but not actually switchable; so e.g. a checkbox checked but disabled.
I mean the user could decide to hide `a`, `b`, `c`, `d`, but `name` is still being displayed.
I can make a PR for this but I'm looking for the best name for such an option... maybe `column.switchableCannotSwitchOff` that kicks in if column is `visible = true, switchable=true`.
That way it wil be counted as switchable.

2) To have the option to have a different label for in the switchable dropdown?
In a table (horizontal) sometimes we want to use an abbreviation for the column label, bun the switchable dropdown (vertical) we have more space to display the full label.
Again I could create a PR for this; e.g. to add a `column.switchableLabel` option, that if specified, overwrites the label?

@wenzhixin @UtechtDustin @DustinUtecht let me know what you think, then I can provide a PR.

marceloverdijk:
@UtechtDustin For 1. my use case is as follows: 
- to have the `name` column visible on the switchable dropdown; so user knows it is displayed but not hideable
- next to that, it would mean (like in my example) that `a`, `b`, `c`, `d` can all be hidden by the user as `name` is always visible; this because now always one column needs to be checked; so one of  `a`, `b`, `c`, `d` still needs to be checked.

UtechtDustin:
I guess the users can see that the column is visible and not hideable if the column is visible in the table and not part of the switchable dropdown, right ? :D   
   
So the the goal of the option should be to always show the `name` column and at least one of the other columns (`a`, `b`, `c` or `d`) ?   
I'm not sure if it makes sense to add more complexity for that feature, if the user hide's something that he needs to work it's his fault/problem or not ?    
I also can't complain that i can't work, because i don't turn on the display...or did i miss something ?

UtechtDustin:
Isn't that already possible with the [minimumCountColumns option](https://bootstrap-table.com/docs/api/table-options/#minimumcountcolumns) (That Option should be renamed i guess...) ?
Example: https://live.bootstrap-table.com/code/UtechtDustin/15552

[...]

PR #6832, opened 2023-06-11
Added switchableLabel column option

marceloverdijk:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

https://github.com/wenzhixin/bootstrap-table/issues/6798

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Added switchableLabel column option as discussed here: https://github.com/wenzhixin/bootstrap-table/issues/6798

**💡Example(s)?**

https://live.bootstrap-table.com/code/marceloverdijk/15617

**☑️Self Check before Merge**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -63,6 +63,7 @@ export interface BootstrapTableColumn {
   checkbox?: boolean;
   halign?: any;
   switchable?: boolean;
+  switchableLabel?: any;
   class?: any;
   escape?: boolean;
   events?: BootstrapTableEvents;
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -63,6 +63,7 @@ export interface BootstrapTableColumn {
   checkbox?: boolean;
   halign?: any;
   switchable?: boolean;
+  switchableLabel?: any;
```
UtechtDustin, line 66: `any` or `string` ?
Does this method only accept a string or also a function ?

marceloverdijk: Only String; PR updated.

UtechtDustin (approved):

PR #6973, opened 2023-09-13
Fix rows reorder when pagination is applied.

sammanan4:
When the pagination is activated and the rows are reordered on pages other than the first one, the order was not taking into account the page number and the page size. This commit changes the index at which the row needs to be inserted by accounting for the page size and page number.

**🤔Type of Request**
- [x] **Bug fix**

**📝Changelog**

- [x] **Extensions**

### DIFF
--- a/src/extensions/reorder-rows/bootstrap-table-reorder-rows.js
+++ b/src/extensions/reorder-rows/bootstrap-table-reorder-rows.js
@@ -97,6 +97,8 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
   onDrop (table) {
     this.$draggingTd.css('cursor', '')
+    const pageNum = $(table).bootstrapTable('getOptions').pageNumber;
+    const pageSize = $(table).bootstrapTable('getOptions').pageSize;
     const newData = []
 
     for (let i = 0; i < table.tBodies[0].rows.length; i++) {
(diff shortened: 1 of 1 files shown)


src/extensions/reorder-rows/bootstrap-table-reorder-rows.js
```diff
@@ -97,6 +97,8 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
   onDrop (table) {
     this.$draggingTd.css('cursor', '')
+    const pageNum = $(table).bootstrapTable('getOptions').pageNumber;
+    const pageSize = $(table).bootstrapTable('getOptions').pageSize;
```
wenzhixin, line 101:
```suggestion
    const pageNum = this.options.pageNumber
    const pageSize = this.options.pageSize
```

src/extensions/reorder-rows/bootstrap-table-reorder-rows.js
```diff
@@ -109,7 +111,7 @@ $.BootstrapTable = class extends $.BootstrapTable {
     const draggingRow = this.data[this.draggingIndex]
     const droppedIndex = newData.indexOf(this.data[this.draggingIndex])
     const droppedRow = this.data[droppedIndex]
-    const index = this.options.data.indexOf(this.data[droppedIndex])
+    const index = ((pageNum - 1)*pageSize) + this.options.data.indexOf(this.data[droppedIndex])
```
wenzhixin, line 114:
```suggestion
    const index = (pageNum - 1) * pageSize + this.options.data.indexOf(this.data[droppedIndex])
```

wenzhixin:

sammanan4:
Please see https://examples.bootstrap-table.com/#extensions/reorder-rows.html
When trying to reorder the rows on any page other than the first one, and then change the page, you can see the ordering is not correct. Thanks.

UtechtDustin:
This is an example using you're commit: https://live.bootstrap-table.com/code/UtechtDustin/16248
Please check the console, the reording is broken.
```
Uncaught SyntaxError: redeclaration of const pageNum
note: Previously declared at line 100, column 10
```

sammanan4:
@UtechtDustin I have fixed the redeclaration issue in the previous commit, please try to run again. I am not able to load the updated code in the live editor though.

UtechtDustin (approved):

UtechtDustin:
Thank you for the contibution, the latest commit is in the cache of the used cdn.   
I tested it locally, works fine.

PR #7008, opened 2023-10-04
Column-specific footerStyle added

Blue3957:
**🤔Type of Request**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
No

**📝Changelog**
Makes the footerStyle property available on columns, as an override of the table footerStyle property

- [x] **Core**
- [ ] **Extensions**

Our team uses bootstrap-table on a variety of pages, so we've taken to making global formatters to use wherever. This works great for cell formatters (date, currency, etc...), footer formatters (sums mostly), and cell styles (colors, backgrounds).
These three properties are defined at the column level. The footerStyle property is an outlier, being defined for the entire table. 

The column being passed to the callback as an argument feels like a workaround to apply different styles to different columns. It makes it difficult to create global formatters that include column-specific styles, because the column-specific logic has to be part of the function, and not the definition of the table. I hope I'm managing to convey the idea correctly.

This small ajustment allows defining the footerStyle on the column.

Naturally, the property can still be defined on the whole table, and this change will not break existing tables. This pull requests simply makes the property available on columns, in which case it'll override the global style formatter.

**💡Example(s)?**
Using both table-level and column-level footer styles: https://live.bootstrap-table.com/code/Blue3957/16322

**☑️Self Check before Merge**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -75,6 +75,7 @@ export interface BootstrapTableColumn {
   cellStyle?: any;
   searchable?: boolean;
   footerFormatter?: any;
+  footerStyle?: any;
   formatter?: any;
   checkboxEnabled?: boolean;
   field?: any;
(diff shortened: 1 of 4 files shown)


wenzhixin (approved):

wenzhixin:
Thanks for your contribution!

PR #7084, opened 2023-12-03
Fix #5713 set caret position to the end of input value

### LINKED ISSUE #5713: FilterControl input bug with cursor, if table has height + FIX
1.18.3

**Description**  
If table has no height, so it works fine. But if you set height to table > 0, it will work with incorrect header.

**Example**  
https://live.bootstrap-table.com/code/Federico-Luna/7412

**Possible  (optional)**  
You can fix this bug by adding code to bootstrap-table-filter-control.js in function setValues():
```
if (that.options.height) {
    var elementClass = getElementClass($this);
    $this = $__default['default'](".fixed-table-header .".concat(elementClass));
}
```

Full example:

```
function setValues(that) {
    var field = null;
    var result = [];
    var searchControls = getSearchControls(that);
    if (that.options.valuesFilterControl.length > 0) {
      //  Callback to apply after settings fields values
      var fieldToFocusCallback = null;
      searchControls.each(function (i, el) {
        var $this = $__default['default'](el);

        if (that.options.height) {
           var elementClass = getElementClass($this);
           $this = $__default['default'](".fixed-table-header .".concat(elementClass));
        }

        field = $this.closest('[data-field]').data('field');
        result = that.options.valuesFilterControl.filter(function (valueObj) {
          return valueObj.field === field;
        });

        if (result.length > 0) {
          if ($this.is('[type=radio]')) {
            return;
          }

          $this.val(result[0].value);

          if (result[0].hasFocus && result[0].value !== '') {
            // set callback if the field had the focus.
            fieldToFocusCallback = function (fieldToFocus, carretPosition) {
              // Closure here to capture the field and cursor position
              var closedCallback = function closedCallback() {
                fieldToFocus.focus();
                setCursorPosition(fieldToFocus);
              };

              return closedCallback;
            }($this.get(0), result[0].position);
          }
        }
      }); // Callback call.

      if (fieldToFocusCallback !== null) {
        fieldToFocusCallback();
      }
    }
 } 
```
  
Also I have changed function setCursorPosition() (because my table sets cursor at start of string, not end):

```
function setCursorPosition(el) {
     var v = el.value;
     $__default['default'](el).val("");
     $__default['default'](el).val(v);
}
```

vuvtdhh:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5713 

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/dist/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/dist/extensions/filter-control/bootstrap-table-filter-control.js
@@ -4394,6 +4394,10 @@
             callbacks.push(fieldToFocusCallback);
           }
         }
+        // fix #5713
+        if ($this[0] === document.activeElement) {
+          setCaretPosition($this[0], $this[0].value.length);
+        }
       });
 
       // Callback call.


wenzhixin:
We don't need to modify the files under dist.

PR #7184, opened 2024-02-14
When searching do not split header field by '.' if there is such property in the item.

veselin-kutsarov:
When searching do not split header field by '.' if there is such property in the item.


wenzhixin:

wenzhixin (approved):

wenzhixin:
Thanks for your contribution!

PR #7251, opened 2024-03-27
Allow uniqueId to be a mix of numbers and strings.

mampcs:
If some rows have the uniqueId as a number and some rows have it as a string, the code would try to convert the id being searched to a number, which would result in a NaN if it was a string. This would then fail to match any rows.
The proper fix is to always use the original _id value to convert to the appropriate type for the comparison.

**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2662,12 +2662,12 @@ class BootstrapTable {
       }
 
       if (typeof rowUniqueId === 'string') {
-        id = id.toString()
+        id = _id.toString()
       } else if (typeof rowUniqueId === 'number') {
         if (Number(rowUniqueId) === rowUniqueId && rowUniqueId % 1 === 0) {
-          id = parseInt(id, 10)
+          id = parseInt(_id, 10)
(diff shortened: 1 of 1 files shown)


wenzhixin:
Please provide an [Online Example](https://live.bootstrap-table.com) to show the problem, just like other PRs. Thanks!

Before: https://live.bootstrap-table.com...
After: https://live.bootstrap-table.com...

wenzhixin:
@mampcs

mampcs:
Link to example that shows the bug.

https://live.bootstrap-table.com/code/mampcs/17361

wenzhixin:
After example: https://live.bootstrap-table.com/code/wenzhixin/17381

wenzhixin:
Thanks for your contribution!

PR #7304, opened 2024-04-28
Possibility to modify/process loaded data before rendering into the t…

utilmind:
The update allows to modify the data before it will be rendered into the table.

Previously I did all data processing on the backend on the stage of producing the JSON. However it significantly cheaper for overall performance to process data on the front-end, when it needed. Do all required clean up, type/format conversions, sorting etc, after the receiving the JSON but before it will be actually loaded into the table.

I'm sure that it will not make any issues for developers who already using `onLoadSuccess` event handler. There is no asynchronous operations between the event and actual loading, rendering occurs after completion of entire 'success' handler on HTTP request. However, if there are any other reasons to call 'load-success' _before_ the actual `load()`, then it worth to add something like `onBeforeLoadSuccess` event, which will allow to modify data before its actual load.

**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [X] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**

- [X] **Core**
- [ ] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/api/events.md
+++ b/site/docs/api/events.md
@@ -215,9 +215,9 @@ $('#table').on('event-name.bs.table', function (e, arg1, arg2, ...) {
 
 - **Detail:**
 
-  It fires when remote data is loaded successfully. The parameters contain:
+  It fires when remote data is loaded successfully, before the data rendered in the table. The parameters contain:
 
-  * `data`: the remote data.
+  * `data`: the remote data. It can be modified (processed) in this event handler before actual loading into the table.
   * `status`: the status code of `jqXHR`.
   * `jqXHR`: jqXHR object, which is a super set of the XMLHTTPRequest object. For more information, see the [jqXHR Type](http://api.jquery.com/Types/#jqXHR).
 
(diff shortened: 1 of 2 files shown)


wenzhixin:
I believe that an event's primary role is to receive data rather than to modify the data it references. This is not a good solution. 

There is already an option for this use case: `responseHandler`. You can use this method to handle the response data format before loading remote data.

utilmind:
@wenzhixin yes, thanks! I checked out the code more thoroughly and see that `responseHandler` even better because it also allows to change the number of items in the list during processing (append or delete items before initialization of pagination).

If you don't mind I will update documentation, will mention `responseHandler` in description of the `onLoadSuccess` event, that it can be used to modify data. It was not obvious to me, I looked this functionality within events, not options.

(Although now I think that these 2 functions, `responseHandler` and `onLoadSuccess` could be combined for optimization, they probably don't need to be split :)

wenzhixin:
OK, merged.

PR #7312, opened 2024-05-01
Increment totalRows and pageSize on row insertion in server mode

OnlyPassionCode:
**🤔Type of Request**
- [x] **Bug**

- [x] **Core**

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2536,6 +2536,10 @@ class BootstrapTable {
 
     // #431: support pagination
     if (this.options.pagination && this.options.sidePagination === 'server') {
+      if (this.options.hasOwnProperty("backUpPageSizeServer")) {
+        this.options.pageSize = this.options.backUpPageSizeServer // reset pageSize
+        delete this.options.backUpPageSizeServer
+      }
       this.options.totalRows = data[this.options.totalField]
       this.options.totalNotFiltered = data[this.options.totalNotFilteredField]
       this.footerData = data[this.options.footerField] ? [data[this.options.footerField]] : undefined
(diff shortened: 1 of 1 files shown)


wenzhixin:

OnlyPassionCode:
> Please provide an [Online Example](https://live.bootstrap-table.com) to show your problem. Thanks!

Here is a example : https://live.bootstrap-table.com/code/OnlyPassionCode/17570

wenzhixin:
After Example: https://live.bootstrap-table.com/code/wenzhixin/17572 with your branch.

OnlyPassionCode:
> After Example: https://live.bootstrap-table.com/code/wenzhixin/17572 with your branch.

I just made a correction.

In fact when inserting data, I increased the pageSize variable by 1 but the problem is that this variable is not reset during the next initServer.
So I created a backUp system in order to reset this pageSize variable.
Additionally, now in the insertRow the this.data variable contains the options data, why is this important? This is due to another bug which consists of when we add a row at index 0 and we want to modify the row at index 1 by clicking on it for example and the event gives us the 'index 1 but actually gives us the row with index 2 because he did not put our new row in this.data because we are in server mode.

OnlyPassionCode:
Here is an example of a use case where we would like to retrieve the row clicked following the insertion of a row : 
https://live.bootstrap-table.com/code/OnlyPassionCode/17573

OnlyPassionCode:
Here is an example of the final usage, but I don't understand.
The libraries section does not take my latest commits and therefore it does not work.

https://live.bootstrap-table.com/code/OnlyPassionCode/17574

OnlyPassionCode:
I'm going to work on a cleaner system closer to that of client mode. And of course the debugs of the bugs that I cited too

PR #7353, opened 2024-05-27
fix/7318

### LINKED ISSUE #7318: 'refresh' method doesn't reuse parameters provided as "query", it only takes "url".

1.22.5 (latest)

### Description

Unfortunately "refresh" button on toolbar is useless, if I originally provide the request parameters as "query" property, instead of straightforward GET-query, as part of "url" property.

So if I originally provide data as follows, this will not work on manual "refresh" (click on refresh button on toolbar), it will just ignore "query" on repeating request:

```
$table.bootstrapTable('refresh', {
                            url: thisToolUrl,
                            query: prepareHttpQuery(),
                        });
```
However, the following will work:
```
$table.bootstrapTable('refresh', {
                            // httpBuildQuery converts array to &-separated and urlencoded 'key=value' pairs.
                            url: thisToolUrl + '?' + httpBuildQuery(prepareHttpQuery()), 
                        });
```

I think it should not ignore data provided in "query" on repeating "refreshes".
The workaround provided above is works for me, so I don't submitting the pool request, but would switch to "query" if it will be fixed in further Bootstrap Table versions.

### Example(s)

_No response_

### Possible Solutions

Fix initServer() method, reuse saved "query" parameter from previous "refresh" request.

### Additional Context

_No response_

UtechtDustin:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #7318

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**

Click the first button and after that the refresh button.

Before: https://live.bootstrap-table.com/code/utilmind/17722
After: https://live.bootstrap-table.com/code/UtechtDustin/17731

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1971,7 +1971,7 @@ class BootstrapTable {
     })
   }
 
-  initServer (silent, query, url) {
+  initServer (silent) {
     let data = {}
     const index = this.header.fields.indexOf(this.options.sortName)
 
@@ -1991,7 +1991,7 @@ class BootstrapTable {
       params.pageNumber = this.options.pageNumber
     }
 
-    if (!(url || this.options.url) && !this.options.ajax) {
+    if (!this.options.url && !this.options.ajax) {
       return
     }
 
@@ -2039,8 +2039,6 @@ class BootstrapTable {
       params.filter = JSON.stringify(this.filterColumnsPartial, null)
     }
 
-    Utils.extend(params, query || {})
-
     data = Utils.calculateObjectValue(this.options, this.options.queryParams, [params], data)
 
     // false to stop request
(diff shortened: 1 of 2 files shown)


wenzhixin:
The issue still requires verification and may impact the extension.

UtechtDustin:
Any news ?

wenzhixin:
Updated: https://github.com/wenzhixin/bootstrap-table/pull/7353/commits/d11d5fa24e99c4def7ab35f510d7e23bf6bcc669

We need to preserve the `query` parameter of the `initServer` method, as they are primarily intended for use by other extention. However, the `url` parameter here can indeed be removed.

For the `refresh` method, use your notation without passing the `query`.

PR #7365, opened 2024-06-17
fix: Loop Stops Processing Rows When Column is Missing

sokolovskiy0103:
This pull request fixes a bug  where the loop would exit early when a column is not found, instead of continuing to the next iteration. The current implementation uses return, which stops processing further columns for the current row. This fix changes return to continue to ensure that all columns are processed correctly.

**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

This bug fix ensures that all columns are processed for each row, even if a specific column is not found in the columns array. This is critical for cases where some columns may not be present but should not disrupt the processing of the remaining columns.

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

This could lead to incomplete data formatting and potential issues in the table display. With this fix, the loop continues to process all remaining columns, providing a more robust and reliable data handling experience.

**💡Example(s)?**
Before:
https://live.bootstrap-table.com/code/sokolovskiy0103/17796
After:
https://live.bootstrap-table.com/code/sokolovskiy0103/17798

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2514,7 +2514,7 @@ class BootstrapTable {
           const column = this.columns[this.fieldsColumnsIndex[key]]
 
           if (!column) {
-            return
+            continue
           }
 
           row[key] = Utils.calculateObjectValue(column, this.header.formatters[column.fieldIndex], [value, row, row.index, column.field], value)
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2509,7 +2509,7 @@ class BootstrapTable {
     }
 
     if (params && params.formatted) {
-      data.forEach(row => {
+      return data.map(row => {
         for (const [key, value] of Object.entries(row)) {
           const column = this.columns[this.fieldsColumnsIndex[key]]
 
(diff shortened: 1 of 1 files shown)


wenzhixin (approved):

src/bootstrap-table.js
```diff
@@ -2514,7 +2514,7 @@ class BootstrapTable {
           const column = this.columns[this.fieldsColumnsIndex[key]]
 
           if (!column) {
-            return
+            continue
           }
 
           row[key] = Utils.calculateObjectValue(column, this.header.formatters[column.fieldIndex], [value, row, row.index, column.field], value)
```
wenzhixin, line 2520: Found a problem: since `row[key] = …` will change the original value, if getData is clicked multiple times, it will be executed multiple times, leading to incorrect results. 
The better way:
```js
      return data.map(row => {
          const column = this.columns[this.fieldsColumnsIndex[key]]

          if (!column) {
            continue
          }

          return Utils.calculateObjectValue(column, this.header.formatters[column.fieldIndex],
            [value, row, row.index, column.field], value)
        }
      })
```

sokolovskiy0103: Agreed, fixed.

wenzhixin (approved):

PR #7413, opened 2024-07-17
Handle undefined values when searching fields

nwalters512:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

When using dotted fields for columns, a search will now correctly handle undefined values instead of throwing an error if they're encountered while navigating through objects with the dotted fields.

**💡Example(s)?**

https://live.bootstrap-table.com/code/nwalters512/17939

Try searching for USD; observe that the following error is logged to the console:

```
bootstrap-table.min.js:10 Uncaught TypeError: Cannot read properties of undefined (reading 'value')
    at bootstrap-table.min.js:10:53552
    at Array.filter (<anonymous>)
    at t.value (bootstrap-table.min.js:10:53149)
    at t.value (bootstrap-table.min.js:10:51731)
    at bootstrap-table.min.js:10:49468
```

This error should not appear after this change.

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1080,7 +1080,7 @@ class BootstrapTable {
             const props = key.split('.')
 
             for (let i = 0; i < props.length; i++) {
-              if (value[props[i]] !== null) {
+              if (value[props[i]] != null) {
                 value = value[props[i]]
               } else {
                 value = null
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1080,6 +1080,7 @@ class BootstrapTable {
             const props = key.split('.')
 
             for (let i = 0; i < props.length; i++) {
+              // eslint-disable-next-line eqeqeq
               if (value[props[i]] != null) {
                 value = value[props[i]]
               } else {
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1080,17 +1080,11 @@ class BootstrapTable {
             const props = key.split('.')
 
             for (let i = 0; i < props.length; i++) {
-              // eslint-disable-next-line eqeqeq
               if (value[props[i]] === null || value[props[i]] === undefined) {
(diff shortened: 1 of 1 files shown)


src/bootstrap-table.js
```diff
@@ -1080,7 +1080,8 @@ class BootstrapTable {
             const props = key.split('.')
 
             for (let i = 0; i < props.length; i++) {
-              if (value[props[i]] !== null) {
+              // eslint-disable-next-line eqeqeq
```
nwalters512, line 1083: An alternative to this is to update the ESLint config to use `"eqeqeq": ["error", "smart"]` instead of just `"eqeqeq": "error"`.

jonatanschroeder: This could also be replaced with the functionality of `getItemField` and consider any falsy value:

https://github.com/wenzhixin/bootstrap-table/blob/eab8d6f1bad7b2371c3560fd696592e7429a513f/src/utils/index.js#L492-L494

src/bootstrap-table.js
```diff
@@ -1080,7 +1080,8 @@ class BootstrapTable {
...
+              if (value[props[i]] != null) {
```
wenzhixin, line 1084: It's better to try not to modify the ESLint configuration.

```suggestion
              if (value[props[i]] === null || value[props[i]] === undefined) {
                value = null
                break
              } else {
                value = value[props[i]]
              }
```

wenzhixin:
@nwalters512, can you help to update the changes?

nwalters512:
@wenzhixin done! Sorry, I didn't realize your code suggestion had introduced syntax issues.

wenzhixin (approved):

PR #7445, opened 2024-08-01
[filter-control] Always return the same control container target

### LINKED ISSUE #7171: setting height option breaks the Filter Control extension

1.22.1

### Description

When enabling the height option on my bootstrap-table, the extension for column filters, Filter Control, no longer works, specifically for the 'select' filterControl using filterData like "var:SOMEVAR". The dropdowns are no longer populated. Removing the height option makes everything happy again.

### Example(s)

_No response_

### Possible Solutions

_No response_

### Additional Context

_No response_

jayaddison:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #7171.

**📝Changelog**
In the `filter-control` `getControlContainer` function, always return the same target control container, regardless of whether the `height` (fixed-header) option is enabled, and regardless of the filter control `_initialized` state when it is.

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**

  * **Before** - https://live.bootstrap-table.com/code/jayaddison/18021
  * **After** - https://live.bootstrap-table.com/code/jayaddison/18022

**☑️Self Check before Merge**

Edit: untick the reviewer checkboxes that I'd filled by mistake.

### DIFF
--- a/src/extensions/filter-control/utils.js
+++ b/src/extensions/filter-control/utils.js
@@ -17,10 +17,6 @@ export function getControlContainer (that) {
     return $(`${that.options.filterControlContainer}`)
   }
 
-  if (that.options.height && that._initialized) {
-    return that.$tableContainer.find('.fixed-table-header table thead')
-  }
-
   return that.$header
 }
 


jayaddison:
@wenzhixin I haven't yet heard from any affected bugreporters, but I believe that this pull request solves the problem described in #7171 when the `filter-control` extension is used in combination with the `mobile` extension.  I'd be glad for any code review and suggestions.

jayaddison:
Due to my unfamiliarity with this part of the `bootstrap-table` codebase, combined with an apparent absence of demand (and therefore assistance testing the fix), and also the fact that this feature isn't currently required by the project I'm working on (it would be nice, because it might allow removal of other dependencies) -- on balance I'm going to close this to reduce my personal tasklist.

To anyone who might find this could useful: please feel free to retrieve the commits from this branch and continue work from here.  Using `git fetch <remote> pull/7445/head:localbranch` should allow retrieving up to commit dda21c397752e7b10b77f5f18026ab55799811a9 into a local branch.

PR #7464, opened 2024-08-19
fix(reorder-columns): resolve incorrect column values order

Vinidamiaop:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**
fix(reorder-columns): resolve incorrect column values order with detail-view

Fixed an issue in the "reorder-columns" extension where columns had incorrect values when `data-detail-view` was set to true. The problem occurred because the code attempted to access a non-existent `field` property in the detail-view column. Added a check to ensure the `field` property exists before assigning the value.

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**

Here is an example of the problem:
- Before: https://live.bootstrap-table.com/code/Vinidamiaop/18108
- After: https://live.bootstrap-table.com/code/Vinidamiaop/18112

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
+++ b/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
@@ -136,7 +136,9 @@ $.BootstrapTable = class extends $.BootstrapTable {
         const sortOrder = {}
 
         table.el.find('th').each((i, el) => {
-          if ('field' in el.dataset) {
+          if (el.dataset.field !== undefined) {
+            sortOrder[el.dataset.field] = i
+          }
             sortOrder[$(el).data('field')] = i
           }
         })
--- a/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
+++ b/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
@@ -139,7 +139,6 @@ $.BootstrapTable = class extends $.BootstrapTable {
           if (el.dataset.field !== undefined) {
             sortOrder[el.dataset.field] = i
           }
-            sortOrder[$(el).data('field')] = i
           }
         })
 
--- a/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
+++ b/src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
@@ -139,7 +139,6 @@ $.BootstrapTable = class extends $.BootstrapTable {
           if (el.dataset.field !== undefined) {
             sortOrder[el.dataset.field] = i
           }
-          }
         })
 
         this.columnsSortOrder = sortOrder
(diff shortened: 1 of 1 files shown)


src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
```diff
@@ -136,7 +136,9 @@ $.BootstrapTable = class extends $.BootstrapTable {
         const sortOrder = {}
 
         table.el.find('th').each((i, el) => {
-          sortOrder[$(el).data('field')] = i
+          if ('field' in el.dataset) {
```
wenzhixin, line 139: > The [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) can check if a given attribute exists: `'keyname' in element.dataset`. Note that this will walk the [prototype chain](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain) of dataset and may be unsafe if you have external code that may pollute the prototype chain. Several alternatives exist, such as [`Object.hasOwn(element.dataset, 'keyname')`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn), or just checking if `element.dataset.keyname !== undefined`.

URL: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#accessing_values

```suggestion
          if (el.dataset.field !== undefined) {
            sortOrder[el.dataset.field] = i
          }
```

src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
```diff
@@ -136,7 +136,11 @@ $.BootstrapTable = class extends $.BootstrapTable {
         const sortOrder = {}
 
         table.el.find('th').each((i, el) => {
-          sortOrder[$(el).data('field')] = i
+          if (el.dataset.field !== undefined) {
+            sortOrder[el.dataset.field] = i
+          }
+            sortOrder[$(el).data('field')] = i
```
wenzhixin, line 142:
```suggestion
          if (el.dataset.field !== undefined) {
            sortOrder[el.dataset.field] = i
          }
```

src/extensions/reorder-columns/bootstrap-table-reorder-columns.js
```diff
@@ -136,7 +136,10 @@ $.BootstrapTable = class extends $.BootstrapTable {
         const sortOrder = {}
 
         table.el.find('th').each((i, el) => {
-          sortOrder[$(el).data('field')] = i
+          if (el.dataset.field !== undefined) {
+            sortOrder[el.dataset.field] = i
+          }
+          }
```
wenzhixin, line 142:
```suggestion
          if (el.dataset.field !== undefined) {
            sortOrder[el.dataset.field] = i
          }
```

wenzhixin (approved):

PR #7468, opened 2024-08-20
fix(column-reordering): fixed issue with reordered columns and cookies

### LINKED ISSUE #5827: Reorder column
Is there a way to reorders columns by calling a function manually.

I am saving the column order to the cookies using the library but the data underneath is not loading in the correct header.

The moment i change a column all the data will be in ordered in their correct header.

Vinidamiaop:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #5827

**📝Changelog**
Addressed an issue where column values were incorrectly aligned after reloading the page with reordered columns saved in cookies. Updated the data retrieval logic to respect the new column order by adjusting the fields to match the correct columns based on the saved order, ensuring accurate data display.

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**

https://live.bootstrap-table.com/code/bschootcovadis/8602

Steps to reproduce:

1. Move headers
2. Check row values
3. Refresh page

Headers are set correct, rows with the data are not correct

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1562,7 +1562,7 @@ class BootstrapTable {
     }
 
     const tds = this.header.fields.map((field, j) => {
-      const column = this.columns[j]
+      const column = this.options.columns[0][j]
       const value_ = Utils.getItemField(item, field, this.options.escape, column.escape)
       let value = ''
       const attrs = {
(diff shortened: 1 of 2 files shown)


wenzhixin:
Closing this PR in favor of a new fix that resolves the issue entirely within the reorder-columns extension without modifying core files. See branch `feature/fix-reorder-columns-cookies` for the replacement.

PR #7482, opened 2024-09-01
Fix the wrong title of toggle option

ajiho:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [x] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

**📝Changelog**

- [ ] **Core**
- [ ] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/site/docs/api/table-options.md
+++ b/site/docs/api/table-options.md
@@ -13,7 +13,7 @@ toc: true
 
 The table options are defined in `jQuery.fn.bootstrapTable.defaults`.
 
-## -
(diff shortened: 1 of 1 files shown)


wenzhixin:
This `toggle` option only has HTML attributes, not JavaScript attributes, because there is no need to write in `$table.bootstrapTable()`, so it is not defined here.

PR #7523, opened 2024-09-23
Return formatted data for all columns in all rows

### LINKED ISSUE #7500: getData method invocation with the `formatted` detail parameter set to true returns formatted data for only the first column of every row

1.23.0+

### Description

When getData is called with the formatted parameter set to true, not only does it return the formatted data for only the first column of every row, but the `{"column1key":"column1data","column2key":"column2data"}` structure of each row item in the array was lost.

### Example(s)

Before Bug v1.22.6 
https://live.bootstrap-table.com/code/Darc-Charlie/18173

With Bug v1.23.0
https://live.bootstrap-table.com/code/Darc-Charlie/18172

### Possible Solutions

The issue seems to stem from this commit:
https://github.com/wenzhixin/bootstrap-table/commit/5c1da3bdf56d7fcb762eb4f9dfc20a9876fcf55c under this PR #7365.

My suggestion for a fix: accumulate the key and formatted values for each found column into an object. Then return that object as the mapped value of the current row.

```diff
    if (params && params.formatted) {
      return data.map(row => {
+       const formattedColumnData = {};
        for (const [key, value] of Object.entries(row)) {
          const column = this.columns[this.fieldsColumnsIndex[key]]

          if (!column) {
            continue
          }

-         return Utils.calculateObjectValue(column, this.header.formatters[column.fieldIndex],
            [value, row, row.index, column.field], value)
+        formattedColumnData[key] = Utils.calculateObjectValue(column, this.header.formatters[column.fieldIndex],
            [value, row, row.index, column.field], value)
        }
+       return formattedColumnData;
      })
    }
``` 

### Additional Context

_No response_

Darc-Charlie:
- Ensures the returned array is an array of objects where each object has a property for every valid column

**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

Fix #7500 

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**

Before:
https://live.bootstrap-table.com/code/Darc-Charlie/18210
After:
https://live.bootstrap-table.com/code/Darc-Charlie/18212

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2465,6 +2465,7 @@ class BootstrapTable {
     if (params && params.formatted) {
       return data.map(row => {
         const formattedColumns = {}
+
         for (const [key, value] of Object.entries(row)) {
           const column = this.columns[this.fieldsColumnsIndex[key]]
 
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -2464,6 +2464,7 @@ class BootstrapTable {
 
     if (params && params.formatted) {
       return data.map(row => {
+        const formattedColumns = {}
         for (const [key, value] of Object.entries(row)) {
           const column = this.columns[this.fieldsColumnsIndex[key]]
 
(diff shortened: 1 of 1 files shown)


src/bootstrap-table.js
```diff
@@ -2464,16 +2464,18 @@ class BootstrapTable {
 
     if (params && params.formatted) {
       return data.map(row => {
+        const formattedColumns = {}
```
wenzhixin, line 2467:
```suggestion
        const formattedColumns = {}

```

wenzhixin (approved):

PR #7584, opened 2024-10-29
Fixed data field attr not working bug

### LINKED ISSUE #7581: Rowspan (and colspan ?) no more working

`>=1.23.3`

### Description

Hello,

adding `_<field>_rowspan` in the data is no more taking in account since the version 1.23.3 incl. It works with 1.23.2

### Example(s)

Here is the example with the version 1.23.3 ; same result with 1.23.4 and 1.23.5, but ok with 1.23.2: https://live.bootstrap-table.com/code/jeje-ch/18296

### Possible Solutions

_No response_

### Additional Context

_No response_

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #7581

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Fixed data field attr not working bug

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/jeje-ch/18296
After: https://live.bootstrap-table.com/code/wenzhixin/18307

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -1585,16 +1585,16 @@ class BootstrapTable {
       }
 
       // handle class, style, id, rowspan, colspan and title of td
-      for (const item of ['class', 'style', 'id', 'rowspan', 'colspan', 'title']) {
-        const value = item[`_${field}_${item}`]
+      for (const attr of ['class', 'style', 'id', 'rowspan', 'colspan', 'title']) {
+        const value = item[`_${field}_${attr}`]
 
         if (!value) {
           continue
         }
-        if (attrs[item]) {
-          attrs[item].push(value)
+        if (attrs[attr]) {
+          attrs[attr].push(value)
         } else {
-          attrs[item] = value
+          attrs[attr] = value
         }
       }
 


src/bootstrap-table.js
```diff
@@ -1585,16 +1585,16 @@ class BootstrapTable {
       }
 
       // handle class, style, id, rowspan, colspan and title of td
-      for (const item of ['class', 'style', 'id', 'rowspan', 'colspan', 'title']) {
-        const value = item[`_${field}_${item}`]
+      for (const attr of ['class', 'style', 'id', 'rowspan', 'colspan', 'title']) {
+        const value = item[`_${field}_${attr}`]
```
jayaddison, line 1589: It took me a few moments to understand what the bugfix is here -- because initially, this diff seems to simply rename two variables.  The fix relates to an outer, existing `item` variable name that was being shadowed/clobbered by the local loop variable (`const <x> of ...`).  The fix is to use a distinct variable name so that the outer object remains accessible and allows the `item[...]` syntax to retrieve a value from it.

PR #7632, opened 2024-11-28
fix issue 7625

### LINKED ISSUE #7625: Wrong behavior combining Filter control and Sticky header extensions + data-sort-empty-last=true + data-sort-select-options=true

1.23.5

### Description

Hi,
We observe an issue when using `Filter control extension` + `Sticky header extension` + `data-sort-empty-last=true` + `data-sort-select-options=true`.
When scrolling the table we see that the select inputs are not well initialized, displaying the first not null value instead of empty by default.
See example attached.

### Example(s)

https://live.bootstrap-table.com/code/adriencoa/18404

### Possible Solutions

My guess: when using the `data-sort-empty-last=true`, it does also on the select input options because of the `data-sort-select-options=true` other option.
But the select option sort should always keep the blank value as default.

### Additional Context

_No response_

EgorTrutnev:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #7625

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/adriencoa/18404
After: https://live.bootstrap-table.com/code/EgorTrutnev/18420

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/sticky-header/bootstrap-table-sticky-header.js
+++ b/src/extensions/sticky-header/bootstrap-table-sticky-header.js
@@ -91,6 +91,14 @@ $.BootstrapTable = class extends $.BootstrapTable {
 
     this.$stickyHeader = this.$header.clone(true, true)
 
+		// copy header inputs value in $stickyHeader clone
+		var $headerInputs = this.$header.find("select, input")
+		var _$stickyHeaderInputs = this.$stickyHeader.find("select, input")
+		$headerInputs.each(function(i) {
+		  var headerInputValue = $(this).val()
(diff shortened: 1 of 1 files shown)


EgorTrutnev:
I see another bug that can be obtained like this:
1. Select Item 5 in the filter
2. Reduce the window size and scroll down the table
3. Select Item 0. It won't be selected and will return Item 5
If you don't scroll down the table (item 2) and stay at the top of the list, then everything works fine

PR #7819, opened 2025-04-26
Refactoring - Extract buttonRendering from main initToolbar()

CloCkWeRX:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**

While trying to simply reverse the display order of when `columns` vs `search` was appended to the DOM in initToolbar, I found it difficult as everything was appending to `html`.

Minor refactor to split out a specific sub rendering behaviour to its own thing - more could be done, just doing one chunk at a time.

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/bootstrap-table.js
+++ b/src/bootstrap-table.js
@@ -608,6 +608,51 @@ class BootstrapTable {
     this.initBody()
   }
 
+  renderButton (buttonName, buttonConfig) {
+    let buttonHtml
+
+    if (buttonConfig.hasOwnProperty('html')) {
+      if (typeof buttonConfig.html === 'function') {
+        buttonHtml = buttonConfig.html()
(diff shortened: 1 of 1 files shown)


wenzhixin:
Hi @CloCkWeRX,

Thanks for this refactoring! I've created a similar PR (#8290) that includes the same extraction of `renderButton` from `initToolbar`, with one fix: the method call uses `this.renderButton()` instead of the bare `renderButton()` which would cause a `ReferenceError`.

That PR has been merged into `develop`, so I'll go ahead and close this one.

Thanks again for the initiative!

PR #7937, opened 2025-08-18
Inverse Search in Filter-Control

ghost:
**🤔Type of Request**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
It does not resolve an issue in the Filter Control extension, but it adds the ability to do an inverse search.

**📝Changelog**

By prepending a keyword with `[!]`, this keyword gets excluded from the list. It uses AND logic and can be combined with the regular filtering provided by Filter Control.

- [ ] **Core**
- [x] **Extensions**

For example, let's take the latest CVEs published by NVD:

image
 

The user is interested in vulnerabilities related to memory:

image
 

However, the user is not interested in Linux, and excludes these results:

image
 

**💡Example(s)?**
A demo is available at https://live.bootstrap-table.com/code/0x164C9DFC/18990, albeit with a reduced dataset.
The demo also features a customSearch function offering the same functionality to the regular search provided by Bootstrap-Table.

**☑️Self Check before Merge**

I have no updated the documentation. No new customizable parameters were introduced. `[!]` is hardcoded and the algorithm modified accordingly to fit my use-case. If you think this is a useful improvement, please modify as you seem fit for the project.

### DIFF
--- a/src/extensions/filter-control/bootstrap-table-filter-control.js
+++ b/src/extensions/filter-control/bootstrap-table-filter-control.js
@@ -235,9 +235,14 @@ $.BootstrapTable = class extends $.BootstrapTable {
             // Separate include and exclude values
             excludeValues = filterValues.filter(fv => fv.startsWith('[!]')).map(fv => fv.slice(3).trim())
             filterValues = filterValues.filter(fv => !fv.startsWith('[!]')).map(fv => fv.trim())
+            // Filter out empty values after trimming
+            excludeValues = excludeValues.filter(ev => ev !== '')
+            filterValues = filterValues.filter(fv => fv !== '')
           } else if (filterValue.startsWith('[!]')) {
             excludeValues = [filterValue.slice(3).trim()]
             filterValues = []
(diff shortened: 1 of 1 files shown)


wenzhixin:
Sorry for the delayed response! This extension was not originally maintained by me, so I haven't been able to review it in a timely manner.

I just took a look — this is a nice feature improvement. However, there are a couple of concerns:

1. **Breaking change (OR → AND):** The original multi-keyword logic uses OR (any match passes), but this PR changes it to AND (all keywords must match). This would break existing behavior for users relying on the OR logic.
2. **Hardcoded magic string (`[!]`):** The `[!]` prefix is hardcoded, which could conflict with actual data containing that pattern. It would be better to make it a configurable option.

If you'd like to pursue this, I'd suggest making the exclude prefix configurable and keeping the original OR behavior as the default to avoid breaking changes.

Thank you for your contribution!

Issue #8070, opened 2025-11-12
tableExport.jquery.plugin deprecated


Orscheler:
### Bootstraptable version(s) affected

1.25.0

### Description

The [export-function](https://bootstrap-table.com/docs/extensions/export/) is using the tableExport.jquery.plugin plugin, which is being deprecated. The [repo](https://github.com/hhurz/tableExport.jquery.plugin) was closed last month. 

Furthermore the code contains multiple vulnabilities with high (10) and medium (2) rating.

The plugin needs to be replaced by another lib.

marktech0813:
Can I handle this issue now?

marktech0813:
Please check PR.
Thanks.

PR #8071, opened 2025-11-13
Fix issues table export.jquery.plugin deprecated #8070

marktech0813:
What I changed
- Implemented a built-in exporter in src/extensions/export/bootstrap-table-export.js that:
  Supports types: json, xml, csv, txt, sql, and maps excel to CSV (.xls).
  Preserves existing behaviors: fileName function support, footer export, column visibility (forceExport/forceHide), selected/all/basic export, and detail-view column exclusion via ignoreColumn.
  Uses safe Blob download, avoiding deprecated/vulnerable plugin code.
- Updated docs in site/src/pages/docs/extensions/export.mdx:
  Removed plugin requirement.
  Documented built-in options (fileName, csvDelimiter, tableName) and supported types.
- Updated CONTRIBUTING.md to reflect that Table Export is now built-in.

Notes
- Users no longer need to include the deprecated plugin; exports work out of the box.
- If you publish new artifacts, run the build to regenerate dist/ so consumers loading from dist/ get the changes.
Status: I replaced the plugin with a built-in exporter, updated docs, and removed references.
I removed the deprecated tableExport.jquery.plugin usage and implemented a built‑in exporter.

Contribution by Gittensor, learn more at https://gittensor.io/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ restrictions:
 
 * Please **do not** open issues or pull requests regarding the code in [`bootstrap-table-examples`](https://github.com/wenzhixin/bootstrap-table-examples) and [`extensions plugin dependence`](https://github.com/wenzhixin/bootstrap-table/tree/develop/src/extensions) (open them in their respective repositories), the dependence list:
     * Table Editable: [x-editable](https://github.com/vitalets/x-editable)
-    * Table Export: [tableExport.jquery.plugin](https://github.com/hhurz/tableExport.jquery.plugin)
+    * Table Export: built-in exporter (no external plugin)
     * Table Reorder-Columns: [jquery-ui](https://code.jquery.com/ui/) and [dragTable](https://github.com/akottr/dragtable/)
     * Table Reorder-Rows: [tablednd](https://github.com/isocra/TableDnD)
     * Table Resizable: [jquery-resizable-columns](https://github.com/dobtco/jquery-resizable-columns)
--- a/site/src/pages/docs/extensions/export.mdx
+++ b/site/src/pages/docs/extensions/export.mdx
@@ -6,9 +6,7 @@ group: extensions
 toc: true
 ---
 
-Use Plugin: [tableExport.jquery.plugin](https://github.com/hhurz/tableExport.jquery.plugin)
-
-This is an important link to check out as some file types may require extra steps.
+This extension now uses a built-in exporter (no external plugin required).
 
 ## Usage
 
@@ -66,7 +64,10 @@ This is an important link to check out as some file types may require extra step
 
 - **Detail:**
 
-  Export [options](https://github.com/hhurz/tableExport.jquery.plugin#options) of `tableExport.jquery.plugin`
+  Options for the built-in exporter:
+  - `fileName`: String or function returning the base filename (without extension).
+  - `csvDelimiter`: CSV delimiter (default `,`).
+  - `tableName`: Table name used for SQL/XML exports (default `table`).
 
   `exportOptions.fileName` can be a string or a function, for example:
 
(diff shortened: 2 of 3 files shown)


marktech0813:
Please check PR.
thanks.

ajiho:
That's awesome!

marktech0813:
I fixed, Please check PR again,
thanks.

fsay2604:
Nice! Any idea when/if it will be merge?

MartinPerrie:
Same question as @fsay2604 ! Any idea when this might be merged? I hit an issue with jQuery 4, where the the "trim" function has been removed. This function is used by tableexport.jquery.plugin - so the Export extension won't work with jQuery 4.

simonsolutions:
+1 for review, this is a great change to not rely on external (and now deprecated) repos. Well done <3

wenzhixin:
Okay, I will review and conduct test verification as soon as possible.

wenzhixin:
Hi @marktech0813, thanks for working on this! Replacing the deprecated plugin with a built-in exporter is a much-needed change. I've reviewed the PR in detail and have some feedback:

## Key Issues

### 1. Data collection via DOM parsing instead of data API
`collectTableMatrix()` reads table content from the DOM using `$(td).text().trim()`. This approach has several problems:
- It only captures the **display text**, not the original data values (e.g., formatted numbers, dates get exported as display strings, not raw values)
- It conflicts with the existing `forceExport`/`forceHide` logic — the code still calls `showColumn`/`hideColumn` to manipulate which columns are visible, but the actual data collection walks the DOM independently
- It may produce inconsistent results depending on current sort/filter state

I'd recommend collecting export data from bootstrap-table's internal data model (`getData()`, `getVisibleFields()`, column definitions with `field`, `formatter` etc.) rather than scraping the rendered DOM.

### 2. Export type reduction is a breaking change
The plugin previously supported 10 types (`json`, `xml`, `png`, `csv`, `txt`, `sql`, `doc`, `excel`, `xlsx`, `pdf`). This PR reduces it to 6, dropping `png`, `xlsx`, `pdf`, `doc`, and `powerpoint`. For users who rely on PDF or Excel exports, this is a breaking change.

Would you consider keeping optional support for these types via lightweight libraries (e.g., SheetJS for xlsx, jsPDF for pdf)?

### 3. `excel` type is misleading
The `excel` type generates CSV content but uses a `.xls` extension. This causes Excel to show a format warning when opening the file. It's not a real Excel file — please consider being upfront about this in the docs, or use `.csv` extension instead.

## Copilot Review Items (agreed)
The Copilot review raised valid points that should be addressed:

1. **Trailing whitespace** — ESLint will fail with `no-trailing-spaces: 'error'`
2. **XML tag names from headers** — Headers with spaces (e.g., "First Name") produce invalid XML tags like `<First Name>`. Header-derived tag names should be sanitized to valid XML element names, falling back to `colN`
3. **`tableName` in XML root** — Should be sanitized for valid XML element names
4. **SQL export edge cases** — `tableName` not escaped for backticks; empty rows produce invalid `INSERT` statement (`VALUES\n;`). Should return empty content when no rows exist
5. **Unsupported types silently no-op** — When an unsupported type is passed, `export-saved` event still fires even though no file was produced. Should return a flag or skip the event

## Missing Updates

- **Chinese docs not updated** — Only the English `export.mdx` was updated. `site/src/pages/zh-cn/docs/extensions/export.mdx` still references the old plugin
- **Cypress tests need rewriting** — `cypress/extensions/export/options.js` (line 184-211) mocks `$.fn.tableExport` and asserts the plugin was called. These tests need to be updated to verify the built-in exporter behavior instead

## Summary
The direction is right and this PR is much needed (especially for jQuery 4 compatibility). The main concern is the DOM-based data collection approach — switching to a data-API-driven approach would make the exporter more robust and consistent with the rest of bootstrap-table's architecture.

PR #8221, opened 2026-02-28
Fixed SyntaxError when processing `data-*` attributes on inline HTML elements

### LINKED ISSUE #8215: SyntaxError with attributes data-xxx-yyy on inline html

1.27.0

### Description

Image

Hello, I encounter an error that occurs when a data-tableexport-xlsxformati attribute is written in an HTML td tag

### Example(s)

_No response_

### Possible Solutions

_No response_

### Additional Context

_No response_

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #8215

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Fixed SyntaxError when processing `data-*` attributes on inline HTML elements

**💡Example(s)?**
Before: https://live.bootstrap-table.com/code/duongvinhad/19119
After: https://live.bootstrap-table.com/code/wenzhixin/19120

**☑️Self Check before Merge**

### DIFF
--- a/src/utils/table-data.js
+++ b/src/utils/table-data.js
@@ -213,7 +213,7 @@ export function trToData (columns, els) {
     // save tr's id, class and data-* attributes
     row._id = DOMHelper.attr(el, 'id')
     row._class = DOMHelper.attr(el, 'class')
-    row._data = getRealDataAttr(el.dataset || {})
+    row._data = getRealDataAttr({ ...el.dataset })
     row._style = DOMHelper.attr(el, 'style')
 
     const cells = DOMHelper.children(el, 'td,th')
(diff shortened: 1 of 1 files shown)


jayaddison:
> Fixes a runtime `SyntaxError` triggered while converting HTML `data-*` attributes to row/cell metadata during `tr`/`td` parsing, by avoiding mutating the live `dataset` (`DOMStringMap`) when normalizing attribute names.

I'm not sure that it's the mutation of the `dataset` that is the main problem; it is possible to assign many string key-value pairs into it at runtime (even if that's not good practice).  However: it does not accept keys that have a hyphen (`-`) in the name.

For example, in a Firefox developer console session:

```javascript
>> var x = document.createElement("a");
<- undefined
>> x.dataset["foo"] = 1;
<- 1
>> x.dataset
<- DOMStringMap { foo → "1" }
>> x.dataset["foo-bar"] = 1;
!! Uncaught DOMException: An invalid or illegal string was specified
!!     <anonymous> debugger eval code:1
!! debugger eval code:1:1
```

The fix uses the spread operator (`...`) to transform the `DOMStringMap` contents into a regular object before calling `getRealDataAttr` (that, prior to `v1.27.0`, operated on the output of jQuery `$(el).data()` -- so this fixes a regression that occurred when jQuery was removed in #8147, I think).

wenzhixin:
Thanks for the clarification, @jayaddison. You're right — the root cause is that `DOMStringMap` doesn't accept keys containing hyphens. Appreciate the detailed analysis\!

PR #8271, opened 2026-04-02
fix: prevent division by zero when pageSize is 0 or negative

appfast-codev:
In initPagination, if pageSize is 0 or negative, the calculation ~~((totalRows - 1) / pageSize) + 1 causes a division by zero. Added a guard to treat pageSize <= 0 the same as 'show all rows', which sets pageSize to totalRows and allSelected to true. All 339 existing tests pass.

### DIFF
--- a/src/modules/pagination.js
+++ b/src/modules/pagination.js
@@ -50,6 +50,11 @@ export default {
         allSelected = true
       }
 
+      if (opts.pageSize <= 0) {
+        opts.pageSize = opts.totalRows
+        allSelected = true
+      }
+
       this.totalPages = ~~((opts.totalRows - 1) / opts.pageSize) + 1
 
       opts.totalPages = this.totalPages


wenzhixin:

wenzhixin:
Thanks for the review! Here are some suggestions:

**1. Move the guard before `if (opts.totalRows)`**

The current guard only normalizes `pageSize` when `totalRows` is truthy, but `pageFrom`/`pageTo` are calculated unconditionally afterwards. When `totalRows` is 0 and `pageSize` is 0 or negative, this leads to an invalid page range (`pageFrom=1, pageTo=0`). Consider moving the guard before the `if (opts.totalRows)` check:

```js
if (opts.pageSize <= 0) {
  opts.pageSize = opts.totalRows || 1
}
```

Note the `|| 1` fallback — when `totalRows` is 0, setting `pageSize` to 0 would still cause issues in `pageFrom`/`pageTo` calculations.

**2. Add `console.warn` for invalid input**

A `pageSize` of 0 or negative is likely a caller configuration error. Silently correcting it could mask the real problem. Consider adding a warning:

```js
if (opts.pageSize <= 0) {
  console.warn('pageSize must be a positive number, falling back to show all rows.')
  opts.pageSize = opts.totalRows || 1
}
```

wenzhixin:
Thanks for the PR! I've addressed the review feedback and created #8294 with the following improvements based on your work:

1. Moved the `pageSize <= 0` guard **before** the `if (opts.totalRows)` check to also cover the `pageFrom`/`pageTo` calculation.
2. Added `console.warn` to help developers detect invalid `pageSize` configuration.

I'll close this PR in favor of #8294. Thanks again for the contribution!

wenzhixin:
Closing in favor of #8294 which includes the fix with additional improvements.

PR #6950, opened 2023-08-31
Update utils.js

manukieli:
The jQuery line are not working for me, the no filter, after debug test, when I click to one select option, the event "on change" are executed 3 time and in the second time the selected option are completly lost. After replace with pure JS, all work properly and working.

**🤔Type of Request**
- [X] **Bug fix**

PR #8292, opened 2026-04-15
Fix filter-control select option loss and sticky-header undefined crash

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**

**🔗Resolves an issue?**
Fix #6950

**📝Changelog**

- [ ] **Core**
- [x] **Extensions**
- Fixed filter-control select `on change` event firing multiple times and selected option being lost, by replacing manual option toggle with `.val()` for reliable value syncing
- Fixed sticky-header `Cannot read properties of undefined (reading 'offset')` error when `$stickyBegin` / `$stickyEnd` / `$stickyContainer` are undefined

**💡Example(s)?**

**☑️Self Check before Merge**

### DIFF
--- a/src/extensions/filter-control/utils.js
+++ b/src/extensions/filter-control/utils.js
@@ -474,18 +474,11 @@ export function createControls (that, header) {
       const $selectControl = $(currentTarget)
       const value = $selectControl.val()
 
-      if (Array.isArray(value)) {
-        for (let i = 0; i < value.length; i++) {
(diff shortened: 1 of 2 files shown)


wenzhixin:
Good suggestion. We plan to add Cypress tests for select filter controls (single-select, multi-select, and clearing scenarios) in a separate PR to keep this bug fix focused.

PR #8298, opened 2026-04-16
Fix table layout breaking when sorting rows with rowspan

### LINKED ISSUE #3462: Sorting on rowspan tbody breaks table
When sorting a table with data having a defined rowspan causes the "child" rows within the group of rows to disassociated and break rendered improperly. 
[jsfiddle](http://jsfiddle.net/e3nk137y/17419/). Please see following animation:

ezgif-5-caeb0ca3a9

wenzhixin:
**🤔Type of Request**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Improvement**
- [ ] **Documentation**
- [ ] **Other**

**🔗Resolves an issue?**
Fix #3462

**📝Changelog**

- [x] **Core**
- [ ] **Extensions**

Flatten rowspan cells before sorting to prevent table layout from breaking.

**💡Example(s)?**
- Before (broken): https://live.bootstrap-table.com/code/wenzhixin/19155
- After (fixed): https://live.bootstrap-table.com/code/wenzhixin/19156

**☑️Self Check before Merge**

---

### Why this approach?

When sorting a table with rowspan cells, the sort operation only reorders the data array without considering the visual grouping created by rowspan. This causes spanned rows to break away from their group, resulting in a broken table layout.

The fix flattens rowspan cells before sorting by copying the parent cell value into all spanned child rows, making each row independent. After sorting, the table renders correctly with each row containing its own data.

This approach is consistent with how **Excel** handles merged cells during sorting — Excel's recommended practice is to unmerge cells and fill values before sorting, as merged cells are fundamentally incompatible with sort operations. Excel even blocks sorting entirely on ranges with merged cells, showing the error: *"All the merged cells need to be the same size"*.

### DIFF
--- a/src/modules/data.js
+++ b/src/modules/data.js
@@ -186,6 +186,10 @@ export default {
         })
       }
 
+      if (Utils.hasRowspanCells(this.data)) {
+        Utils.flattenRowspanCells(this.data)
+      }
+
       if (this.options.customSort) {
         Utils.calculateObjectValue(this.options, this.options.customSort, [
           this.options.sortName,
--- a/src/utils/table-data.js
+++ b/src/utils/table-data.js
@@ -272,3 +272,69 @@ export function checkAutoMergeCells (data) {
   }
   return false
 }
+
+/**
+ * Checks if any row in the data has rowspan cells with a span greater than 1.
+ *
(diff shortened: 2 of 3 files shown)


wenzhixin (commented):
copilot:review

